PR #8, opened 2023-09-29
Add codeql, scorecard, and verifying license

Hayden-IO:

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -36,7 +36,7 @@ jobs:
       security-events: write
       # Needed to publish results and get a badge (see publish_results below).
       id-token: write
-    uses: sigstore/community/.github/workflows/reusable-scorecard.yml@d0c95c8803672313d0bf72e1a44021be5b583c24 # main
+    uses: sigstore/community/.github/workflows/reusable-scorecard.yml@main
     # (Optional) Disable publish results:
     # with:
     #   publish_results: false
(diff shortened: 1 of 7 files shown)


.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,46 @@
...
+
+jobs:
+  analysis:
+    name: Scorecards analysis
+    permissions:
+      # Needed to upload the results to code-scanning dashboard.
+      security-events: write
+      # Needed to publish results and get a badge (see publish_results below).
+      id-token: write
+    uses: sigstore/community/.github/workflows/reusable-scorecard.yml@d0c95c8803672313d0bf72e1a44021be5b583c24 # main
```
bobcallaway: i'd just pin to main since we're managing this within the community.

Hayden-IO: Done

phillmv (approved):
lgtm!

codysoyland (approved):

codysoyland:
The linter is only scheduled to run [when go code changes](https://github.com/sigstore/sigstore-go/blob/main/.github/workflows/golangci-lint.yml#L21-L24), but it's a required check. Do you want to just remove that constraint so it runs every time? CodeQL will always take longer anyway.

Hayden-IO:
Ah good catch! I thought it was an issue with the import. I’ll get that removed.

codysoyland (approved):

PR #14, opened 2023-10-05
VerificationResult now has a proper mediaType field.

### LINKED ISSUE #13: Update `VerificationResult` to have a proper `mediaType`
**Description**

At the time of writing, `VerificationResult.Version` is [set to an `int`](https://github.com/sigstore/sigstore-go/blob/4bae881388746e5d80b0277301baa12d4373669c/pkg/verify/signed_entity.go#L177). Versioning is well and good, and should help us handle changes to the struct over time.

However, after thinking about it some more, it became apparent that:
1. should `VerificationResult` ever get stored to disk, a plain `int` Version is insufficient for figuring out how to parse the file
2. it is incongruous with other sigstore types & documents in the broader ecosystem

For that reason, I hereby propose we rename `Version` to `MediaType` and set it to `"application/vnd.dev.sigstore.verificationresult+json;version=0.1"` as the current default.

phillmv:
#### Summary

Closes https://github.com/sigstore/sigstore-go/issues/13

As described in the issue, this PR solves the problem of `VerificationResults.Version` being insufficiently descriptive and incongruous with the other json document types that are handled in the broader Sigstore ecosystem.

This PR:
- Changes it from `Version int` to `MediaType string`
- Updates the default value to `application/vnd.dev.sigstore.verificationresult+json;version=0.1`
- And for good measure, adds the option to use `WithoutArtifactUnsafe()` in the `sigstore-go` binary, which I used to test this feature.

#### Release Note

- Changed VerificationResult.Version to VerificationResult.MediaType

#### Documentation

We haven't created a release for this yet, so I don't think any additional docs changes are necessary beyond what will be generated when we do create a release for this library.

### DIFF
--- a/pkg/verify/signed_entity.go
+++ b/pkg/verify/signed_entity.go
@@ -25,6 +25,10 @@ import (
 	"github.com/sigstore/sigstore-go/pkg/root"
 )
 
+const (
+	VerificationResultMediaType01 = "application/vnd.dev.sigstore.verificationresult+json;version=0.1"
+)
+
 type SignedEntityVerifier struct {
 	trustedMaterial root.TrustedMaterial
 	config          VerifierConfig
@@ -174,7 +178,7 @@ type TimestampVerificationResult struct {
 
 func NewVerificationResult() *VerificationResult {
 	return &VerificationResult{
-		MediaType: "application/vnd.dev.sigstore.verificationresult+json;version=0.1",
+		MediaType: VerificationResultMediaType01,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/verify/signed_entity.go
```diff
@@ -174,7 +174,7 @@ type TimestampVerificationResult struct {
 
 func NewVerificationResult() *VerificationResult {
 	return &VerificationResult{
-		Version: 20230823,
+		MediaType: "application/vnd.dev.sigstore.verificationresult+json;version=0.1",
```
kommendorkapten: Nit but we should define the media type as a `const` so it's easier to work with.

steiza (approved):
Looks good to me!

kommendorkapten (approved):

PR #16, opened 2023-10-06
Update README with an example that works

steiza:
#### Summary

I was doing some testing with sigstore-go, and I noticed the README examples were out of date, so I think we should update them.

#### Release Note

NONE

#### Documentation

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Features:
 - Support for custom [trusted root](https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_trustroot.proto)
 - Basic CLI
 
-For an example of how to use this library, see [cmd/sigstore-go](./cmd/sigstore-go/main.go), or see the CLI examples below.
+For an example of how to use this library, see [cmd/sigstore-go](./cmd/sigstore-go/main.go), or see the CLI examples below. Note that the CLI is to demonstrate how to use the library, and not intended as a fully-featured Sigstore CLI like [cosign](https://github.com/sigstore/cosign).
 
 ## Background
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -27,26 +27,37 @@ Sigstore already has a canonical Go client implementation, [cosign](https://gith
...
+}
 ```
 
-```bash
-$ go run cmd/sigstore-go/main.go -tufRootURL tuf-repo-cdn.sigstore.dev examples/bundle-provenance.json
-Verification successful!
-```
+You can also specify a TUF root with something like `-tufRootURL tuf-repo-cdn.sigstore.dev`.
 
 Alternatively, you can install a binary of the CLI like so:
```
Hayden-IO, line 56: Is this binary meant for typical users or just for testing?

steiza: That's correct - the binary is to illustrate how to use the library, and is not intended to be fully functional like other Sigstore CLI tools (e.g. `cosign`).

Hayden-IO: Do we want to make a note of that in the README?

steiza: ... of course, after I push additional language, I notice right below it we also have:

> Sigstore already has a canonical Go client implementation, [cosign](https://github.com/sigstore/cosign), which was developed with a focus on container image signing/verification. It has a rich CLI and a long legacy of features and development. `sigstore-go` is a more minimal and friendly API for integrating Go code with Sigstore, with a focus on the newly specified data structures in [sigstore/protobuf-specs](https://github.com/sigstore/protobuf-specs).

Hayden-IO: I think your clarification helps - The quote does not explicitly say that sigstore-go as a CLI is discouraged.

Hayden-IO (approved):

PR #30, opened 2023-11-29
Example of verifying OCI referenced images using sigstore-go

rdimitrov:
#### Summary

The following PR provides an example of verifying OCI referenced images using sigstore-go. 

Details:
* Uses the sigstore-go CLI as an example and adds on top a `--ociImage` flag
* There's no need to pass any additional bundle alongside the image reference. 
* It builds the bundle by fetching the related image manifests and extracts and normalises the necessary data to construct it.
* Added a document providing an example use of this

I realise I might have missed some corner case so I'll appreciate your feedback on how to make this better 😃

Thanks in advance! 

#### Release Note

Example of verifying OCI referenced images using sigstore-go.


kommendorkapten:
Awesome!

docs/oci-image-verification.md
```diff
@@ -0,0 +1,129 @@
...
+
+The verifier allows you to use the Sigstore Public Good TUF root or your own custom [trusted root](https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_trustroot.proto) containing the root/intermediate certificates of the Fulcio/TSA/Rekor instances used to sign the bundle, in order to verify common open source bundles or bundles signed by your own private Sigstore instance.
+
+## Verification Process
+
+In this example, we'll use the `sigstore-go` CLI to verify an OCI image reference. The CLI is a thin wrapper around the Go API, so the process is the same, but the CLI provides a convenient way to verify an OCI image reference without writing any code.
+The image we are going to verify is `ghcr.io/stacklok/minder/server:latest`
+
+```shell
+sigstore-go --ociImage="ghcr.io/stacklok/minder/server:latest" --tufRootURL="tuf-repo-cdn.sigstore.dev" --expectedSANRegex="^https://github.com/stacklok/minder/" --expectedIssuer="https://token.actions.githubusercontent.com"
```
kommendorkapten:
```suggestion
$ sigstore-go --ociImage="ghcr.io/stacklok/minder/server:latest" \
    --tufRootURL="tuf-repo-cdn.sigstore.dev" \
    --expectedSANRegex="^https://github.com/stacklok/minder/" \
    --expectedIssuer="https://token.actions.githubusercontent.com"
```

kommendorkapten: Slightly more readable IMHO.

kommendorkapten (dismissed):
Thanks @rdimitrov , this is great! 🚀 

One nit comment, otherwise this looks really good.

ping: @haydentherapper @codysoyland

kommendorkapten:
To clarify, the purpose of sigstore-go is **not** to have functionality like this, it should be in cosign. But until cosign is ready to work with bundles, having an example in this repo is fine IMHO.

Hayden-IO:
Overall I’m fine adding this, but I would caution having the script be much more than an example. I don’t expect this will evolve into a more fleshed out CLI. 

We will be working on adding support for the bundle in Cosign very shortly, you can track progress in https://github.com/sigstore/cosign/issues/3139.

cmd/sigstore-go/main.go
```diff
@@ -214,3 +234,214 @@ func trustedPublicKeyMaterial(pk crypto.PublicKey) *root.TrustedPublicKeyMateria
...
+		Content:                   signingCert,
+		TlogEntries:               tlogEntries,
+		TimestampVerificationData: nil,
+	}, nil
+}
+
+// getVerificationMaterialTlogEntries returns the verification material transparency log entries from the simple signing layer
+func getVerificationMaterialTlogEntries(manifestLayer *v1.Descriptor) ([]*protorekor.TransparencyLogEntry, error) {
+	// 1. Get the bundle annotation
+	bun := manifestLayer.Annotations["dev.sigstore.cosign/bundle"]
```
Hayden-IO: This is a reuse of the existing annotation, this is going to be an issue for any existing OCI signatures. 

Given this is not cosign, if this is just meant as an example, I would use a different annotation specific to this library. .

Hayden-IO (changes requested):

cmd/sigstore-go/main.go
Hayden-IO: NVM, misunderstood this function, ignore me.

Hayden-IO:
@rdimitrov, how did you envision the purpose of this code? I don’t want to put words in your mouth if you see this as more than an example. If so, then I’m a little hesitant to put so much in without tests, but I also don’t want this to become cosign.

what do you think about putting some caveats in the documentation that this is meant as an example?

cmd/sigstore-go/main.go
```diff
@@ -214,3 +234,214 @@ func trustedPublicKeyMaterial(pk crypto.PublicKey) *root.TrustedPublicKeyMateria
...
+			},
+			KindVersion: &protorekor.KindVersion{
+				Kind:    kind,
+				Version: apiVersion,
+			},
+			IntegratedTime: int64(integratedTime),
+			InclusionPromise: &protorekor.InclusionPromise{
+				SignedEntryTimestamp: signedEntryTimestamp,
+			},
+			InclusionProof:    nil,
```
Hayden-IO: Inclusion proofs are required. If the OCI image doesn’t store it, then it can’t be used to build a valid bundle.

rdimitrov:
Hey, @haydentherapper 👋 Thanks for your feedback! 🙏

> @rdimitrov, how did you envision the purpose of this code? I don’t want to put words in your mouth if you see this as more than an example. If so, then I’m a little hesitant to put so much in without tests, but I also don’t want this to become cosign.

That makes a lot of sense 👍 We were talking with Fredrik and we agreed this is not intended to be something more than a proof-of-concept. My intention is to replace the use of the cosign library with something smaller like sigstore-go and so we talked if it'd be worth sharing what I've done so far as another example 👍 

> what do you think about putting some caveats in the documentation that this is meant as an example?
 
Of course, updated the doc to explicitly state this 👍

cmd/sigstore-go/main.go
rdimitrov: I was using the example of https://github.com/sigstore/sigstore-go/blob/main/examples/bundle-provenance.json as a reference, so thus why it's like that. Can you give me some pointers on how to construct the necessary data for this?

kommendorkapten: The test image (see the `oci-image-verification.md`) is signed with `cosign`, it appears `cosign` does not upload the inclusion proof, is that correct @haydentherapper ? If correct, thoughts of adding inclusion proofs when uploading to an OCI registry.

The code could be modified to verify it the inclusion proof is present, if not it could be fetched from Rekor prior to creating the bundle (it's present in the entries endpoint, how to view via curl: ` curl -s 'https://rekor.sigstore.dev/api/v1/log/entries?logIndex=54521426'`.

Edit: asked a question if cosign should upload inclusion proof.

cmd/sigstore-go/main.go
```diff
@@ -214,3 +234,214 @@ func trustedPublicKeyMaterial(pk crypto.PublicKey) *root.TrustedPublicKeyMateria
...
+		return nil, nil, fmt.Errorf("error getting verification material: %w", err)
+	}
+	// 3. Build the message signature for the bundle
+	msgSignature, err := getBundleMsgSignature(simpleSigning)
+	if err != nil {
+		return nil, nil, fmt.Errorf("error getting message signature: %w", err)
+	}
+	// 4. Construct and verify the bundle
+	pb := protobundle.Bundle{
+		MediaType:            bundle.SigstoreBundleMediaType01,
```
kommendorkapten: we should use `bundle.SigstoreBundleMediaType02`, but this will make the verification to fail (as the inclusion proof is missing as @haydentherapper pointed out). See https://github.com/sigstore/sigstore-go/pull/30/files#r1412940744 for more discussion on inclusion proof.

cmd/sigstore-go/main.go
kommendorkapten: For a transition period, maybe using the bundle format `v0.1` (i.e only require inclusion promise) is the best way forward, then when we modify cosign, we can start to produce `v0.2` bundle versions.

evankanderson: The goal here is to produce a valid signature, correct (not to specifically implement the current `cosign` behavior)?  It sounds like maybe there's also a bug report for `cosign` here?

cmd/sigstore-go/main.go
Hayden-IO: +1, we shouldn't proliferate v01 bundles.

cmd/sigstore-go/main.go
```diff
@@ -214,3 +234,214 @@ func trustedPublicKeyMaterial(pk crypto.PublicKey) *root.TrustedPublicKeyMateria
...
+	// 2. Get the transparency log entries
+	tlogEntries, err := getVerificationMaterialTlogEntries(manifestLayer)
+	if err != nil {
+		return nil, fmt.Errorf("error getting tlog entries: %w", err)
+	}
+	// 3. Construct the verification material
+	return &protobundle.VerificationMaterial{
+		Content:                   signingCert,
+		TlogEntries:               tlogEntries,
+		TimestampVerificationData: nil,
```
Hayden-IO: Timestamp data is available too, stored in `rfc3161timestamp`

cmd/sigstore-go/main.go
Hayden-IO: You're right that we don't store the inclusion proof - https://github.com/sigstore/cosign/blob/main/pkg/cosign/bundle/rekor.go#L23-L34

Ideally we would be storing the inclusion proof too. For some history, it's because we previously weren't doing anything with inclusion proofs, it's only been recently that we're improving the offline verification from rekor.

I'd rather not make the change in Cosign since the "rekor bundle" format has been stable for some time now. Instead we should implement protobuf bundle support which will include the inclusion proof, since it'll be stored after signing.

I'd personally prefer to not have more v01 bundles without these stronger proofs and instead make the change in Cosign rather than try to translate the existing OCI contents into an old bundle format.

codysoyland:
Thanks @rdimitrov for writing this up! It's interesting that the existing cosign signatures can be packaged into a compliant bundle, as that wasn't really the intent of this library. I think many of us see the future of using Sigstore Bundles in cosign would look something like uploading the bundle to an OCI registry as a referring artifact, and verifying the stored bundle directly. Of course this depends on the ability to sign/create bundles, which is as of yet unsupported in sigstore-go and cosign.

I'm a bit hesitant to merge this for a few reasons:
- As mentioned above, we would like to push bundles directly into OCI registries, which is not compatible with this implementation. See secondary goals in https://github.com/sigstore/cosign/issues/3139
- As mentioned by other reviewers, we would like discourage any more creation of v01 bundles.
- I don't want to add more dependencies to support OCI/docker here, as that is more the role of cosign.
- IMO, the CLI here in sigstore-go is not really something I want to support long-term, and it's intentionally small to demonstrate sigstore-go's core functionality, to use as prior art for people who are trying to use sigstore-go. I have considered moving the CLI into an `examples/` directory to emphasize this.
- No test coverage.

I'd be open to moving this code into an `examples` dir with its own go.mod, but otherwise I'd prefer to reject this PR on these bases.

kommendorkapten:
Just a general comment here:

> I'd personally prefer to not have more v01 bundles 

I don't believe the intent with this example code is to actually store the bundles anywhere? I was always under the impression that this example code is a way for a client to verify a bundle with sigstore-go instead cosign, by creating an in-memory representation, verify it, then throw it away. In that sense, verifying a v1 bundle is as safe as verifying it with cosign.

kommendorkapten:
> I'd be open to moving this code into an examples dir with its own go.mod, but otherwise I'd prefer to reject this PR on these bases.

yes, this is a great option, as this is just an example client using sigstore-go 👍

rdimitrov:
@kommendorkapten @haydentherapper @codysoyland - Thanks all for your feedback! 💯I appreciate it very much!  🙏 

I think @kommendorkapten stated all of this correctly 👍 The bundle is used strictly as an **intermediate format** so I can leverage sigstore-go to verify an OCI image in runtime. I don't intend to produce and output the constructed bundle, store it somewhere, etc. This is just an example of using sigstore-go to verify an image and not intended to a tool for others to use.

I'll try to summarise what I've understood so far (correct me if I'm wrong):
* The `inclusionProof` is necessary to allow for offline verification, but obsolete for online verification.
* Required in v2, but not in v1.
* Its presence will be addressed once the protobuf bundle support in cosign is implemented.
* Until then, signing with `cosign` **does not** produce enough information to construct it and there are no plans to include it in cosign's rekor bundle. 

Based on this:
* This approach is safe to be used in the use case of doing online verifications **only**.
* It will allow to act as an intermediate step to switch seamlessly to v2 once the protobuf bundle signing support is implemented and begins to roll out as part of the OCI image manifests.
* _An open question here is - once this happens, how will cosign handle verifying an image signed with the current (old) format?_

> I'd be open to moving this code into an examples dir with its own go.mod, but otherwise I'd prefer to reject this PR on these bases.

@codysoyland - I was debating whether I should put it there in the first place. I can rework this PR so it aligns with this 👍 

 Thanks again for your support! 🙏

Hayden-IO:
@rdimitrov Thanks for the clarification. My concerns are largely mitigated if this is just an intermediate format and not persisted. 

To Cody’s last point, do we want to add some tests even if this is an example?

docs/oci-image-verification.md
```diff
@@ -0,0 +1,141 @@
...
+
+- Unix-compatible OS
+- [Go 1.21](https://go.dev/doc/install)
+
+## Getting started
+
+Clone this repository and navigate to the `examples/oci-image-verification` directory:
+
+```shell
+
```
codysoyland:
```suggestion
```

codysoyland:
> @codysoyland - I was debating whether I should put it there in the first place. I can rework this PR so it aligns with this 👍

Thanks @rdimitrov for understanding and moving this to an example! I just ran through the doc and everything works on my local machine. 👍🏻

rdimitrov:
Thanks @codysoyland and @haydentherapper! 🙏  

Updated the PR title and description to match, let me know if there's anything else blocking this 👍

codysoyland (approved):

Hayden-IO (approved):

PR #40, opened 2023-12-11
Support Fulcio certificate "chains" that just have a root

steiza:
#### Summary

There were some fantastic conformance tests added with https://github.com/sigstore/sigstore-conformance/pull/112.

However, one of the tests includes a mock Sigstore where Fulcio only has a root certificate and no intermediates. These changes allow that conformance test to get further (note that the conformance test won't pass with these changes, but I think that's because of an issue with the test material provided, see https://github.com/sigstore/sigstore-conformance/pull/112#discussion_r1423082636).

#### Release Note

NONE

#### Documentation

NA

### DIFF
--- a/pkg/verify/sct.go
+++ b/pkg/verify/sct.go
@@ -52,16 +52,21 @@ func VerifySignedCertificateTimestamp(leafCert *x509.Certificate, threshold int,
 		}
 
 		for _, fulcioCa := range fulcioCerts {
+			fulcioChain := make([]*ctx509.Certificate, len(certChain))
+			copy(fulcioChain, certChain)
+
+			var parentCert []byte
+
 			if len(fulcioCa.Intermediates) == 0 {
-				continue
+				parentCert = fulcioCa.Root.Raw
+			} else {
+				parentCert = fulcioCa.Intermediates[0].Raw
 			}
-			fulcioIssuer, err := ctx509.ParseCertificates(fulcioCa.Intermediates[0].Raw)
+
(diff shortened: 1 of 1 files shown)


pkg/verify/sct.go
```diff
@@ -52,16 +52,21 @@ func VerifySignedCertificateTimestamp(leafCert *x509.Certificate, threshold int,
 		}
 
 		for _, fulcioCa := range fulcioCerts {
+			fulcioChain := make([]*ctx509.Certificate, len(certChain))
```
Hayden-IO: nit: Could we rename line 41 to be `leafCTCert` or similar rather than `certChain`, helps readability to see that it should only be a single certificate.
Also can assert that it parsed only a single certificate with a length check

Hayden-IO (dismissed):

pkg/verify/sct.go
steiza: ++ to clearer names!

Since `leafCert` is a `*x509.Certificate`, I don't think there's a case where we'd end up with more than one certificate? I'm going to leave that off for now, but happy to add if I misunderstand.

Hayden-IO: Yea, it shouldn't be possible, I was just saying you could assert that `ParseCertificates` returns a single cert since it returns an array, but it isn't necessary.

Hayden-IO (approved):

kommendorkapten (approved):

codysoyland (approved):

Issue #38, opened 2023-12-07
Implement Sigstore TUF Client Spec


codysoyland:
**Description**

The TUF client in `sigstore-go` is very much an MVP, and we would like to enhance it to accomplish a variety of goals:

- Add the TUF Options specified in the [Sigstore TUF Client](https://docs.google.com/document/d/1QWBvpwYxOy9njAmd8vpizNQpPti9rd5ugVhji0r3T4c/edit) specification. This includes controls for caching behavior and trusted repository mapping.
- Add support for fetching arbitrary targets, including delegate targets. Notably, this client should support the [npm TUF delegation](https://github.com/sigstore/root-signing/tree/main/repository/repository/targets/registry.npmjs.org).
- Add API/ability to initialize multiple TUF repos in their own namespaces, given a URL and `root.json`, which can be used by private Sigstore instances with their own fully-contained TUF repos.

codysoyland:
@kommendorkapten and I have been working on a new Sigstore TUF client internally at GitHub that accomplishes the goals stated above, and we hope to open-source it soon.

We are currently evaluating switching to https://github.com/rdimitrov/go-tuf-metadata in place of https://github.com/theupdateframework/go-tuf.

codysoyland:
This is done in #41.

Hayden-IO:
@codysoyland To confirm, "Add API/ability to initialize multiple TUF repos in their own namespaces, given a URL and root.json, which can be used by private Sigstore instances with their own fully-contained TUF repos." is supported? Slightly separate question, is concurrent multi-TUF-repo supported?

codysoyland:
> [@codysoyland](https://github.com/codysoyland) To confirm, "Add API/ability to initialize multiple TUF repos in their own namespaces, given a URL and root.json, which can be used by private Sigstore instances with their own fully-contained TUF repos." is supported? Slightly separate question, is concurrent multi-TUF-repo supported?

Missed your message until now! Yes, that bullet point should be supported here. If you supply a trust anchor (TUF root.json contents) to Options.Root, you can initialize your own TUF repo.

As for multi-TUF-repo, I think the answer is no -- we're not using the [multirepo package](https://github.com/theupdateframework/go-tuf/blob/master/metadata/multirepo/multirepo.go), but to be honest, I'm not sure how that works exactly.

rdimitrov:
[...]

PR #41, opened 2023-12-11
Update TUF client to support options and add LiveTrustedRoot

codysoyland:
#### Summary

This PR implements #38. This adds:

- TUF client options consistent with Sigstore TUF client spec
- Switch to go-tuf v2
- A live-updating TrustedRoot implementation which is compatible with long-running processes to ensure that TUF metadata is synchronized daily.

Credit to @kommendorkapten for some parts of the client code, which we developed in collaboration at GitHub.

To do:

- [x] Update examples and commands
- [x] Update go.mod
- [x] Add tests


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/sigstore/sigstore-go
 
-go 1.21
+go 1.21.5
```
Hayden-IO: Can this be kept at 1.21? Unless we're using a specific method from 1.21.5, it doesn't have to be precise

pkg/root/trusted_root.go
```diff
@@ -256,3 +259,101 @@ func NewTrustedRootProtobuf(rootJSON []byte) (*prototrustroot.TrustedRoot, error
 	}
 	return pbTrustedRoot, nil
 }
+
+// FetchTrustedRoot fetches the Sigstore trusted root from TUF and returns it.
+func FetchTrustedRoot() (*TrustedRoot, error) {
```
Hayden-IO, line 264: It might be useful to have another method `FetchTrustedRootWithOpts(opts Options)`, which could be used by the main.go example script.

Another option (no pun intended) would be to rework the Options struct to be functional like in https://www.sohamkamani.com/golang/options-pattern/. Then you'd just need one method `FetchTrustedRoot(opts ...Option)`. Also has the benefit of a cleaner API for initialization, rather than needing to initialize Options and then set each field: `opts := tuf.DefaultOptions().WithRepoBaseURL(url)` for example.

pkg/tuf/options.go
```diff
@@ -0,0 +1,81 @@
...
+// Note that currently, the cache control is not working. Upon initialization
+// the client will *ALWAYS* perform a TUF update.
+type Options struct {
+	// CacheValidity period in days (default 1)
+	CacheValidity int
+	// ForceCache controls if the cache should be used without update
+	// as long as the metadata is valid
+	ForceCache bool
+	// Root is the TUF trust anchor
+	Root []byte
```
Hayden-IO, line 48: What do you think about having this be a path instead of the bytes? Maybe a private field that signifies whether it's embedded or not to be used during client initialization too?

pkg/tuf/options.go
```diff
@@ -0,0 +1,81 @@
...
+	// RepositoryBaseURL is the TUF repository location URL
+	// (default https://tuf-repo-cdn.sigstore.dev)
+	RepositoryBaseURL string
+	// DisableLocalCache mode allows a client to work on a read-only
+	// files system if this is set, cache path is ignored.
+	DisableLocalCache bool
+}
+
+// DefaultOptions returns an options struct for the public good instance
+func DefaultOptions() (*Options, error) {
```
Hayden-IO: Can we rework this to not return an error? This would be necessary if we were to follow a functional options pattern, though I think it's a cleaner interface regardless. `DefaultRoot()` could return a path (`DefaultRootPath()`) and set an embedded flag on the struct so the client knows to read from embedded vs disk.

If `UserHomeDir()` fails, we could have some default path. If it were to fail because of an environmental issue (maybe disk failure or something?), it would fail on client initialization.

pkg/tuf/options.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+//go:embed repository
+var embeddedRepo embed.FS
+
+const DefaultMirror = "https://tuf-repo-cdn.sigstore.dev"
+
+// Options represent the various options for a Sigstore TUF Client
+//
+// Note that currently, the cache control is not working. Upon initialization
```
Hayden-IO: Can this be a TODO linking to an issue?

pkg/tuf/client.go
```diff
@@ -15,137 +15,184 @@
...
-
-	fileJSONStore, err := filejsonstore.NewFileJSONStore(tufPath)
-	if err != nil {
-		return nil, err
+// New returns a new client with custom options
+func New(opts *Options) (*Client, error) {
+	var c = Client{
+		opts: opts,
 	}
+	var dir = filepath.Join(opts.CachePath, URLToPath(opts.RepositoryBaseURL))
```
Hayden-IO:
```suggestion
	dir := filepath.Join(opts.CachePath, URLToPath(opts.RepositoryBaseURL))
```

pkg/tuf/client.go
```diff
@@ -15,137 +15,184 @@
...
-			return nil, err
-		}
+	c.cfg.LocalMetadataDir = dir
+	c.cfg.LocalTargetsDir = filepath.Join(dir, "targets")
+	c.cfg.RemoteTargetsURL, err = url.JoinPath(opts.RepositoryBaseURL, "targets")
+	if err != nil {
+		return nil, fmt.Errorf("malformed config mirror: %w", err)
 	}
+	c.cfg.DisableLocalCache = c.opts.DisableLocalCache
+	c.cfg.PrefixTargetsWithHash = true
```
Hayden-IO: Should we make this a top-level configuration, like "DisableConsistentSnapshot"?

pkg/tuf/client.go
```diff
@@ -15,137 +15,184 @@
...
+		c.opts.CacheValidity = 0
+		c.opts.ForceCache = false
 	}
 
-	tufRemoteStore, err := tufclient.HTTPRemoteStore(fmt.Sprintf("https://%s", tufRootURL), tufRemoteOptions, nil)
+	// Upon client creation, we may not perform a full TUF update,
+	// based on the cache control configuration. Start with a local
+	// client (only reads content on disk) and then decide if we
+	// must perform a full TUF update.
+	var tmpCfg = *c.cfg
```
Hayden-IO:
```suggestion
	tmpCfg := *c.cfg
```

pkg/tuf/client.go
```diff
@@ -15,137 +15,184 @@
...
+	tmpCfg.UnsafeLocalMode = true
```
Hayden-IO, line 72: What's this referring to? Should this be set only if the cache is disabled?

pkg/tuf/client.go
Hayden-IO, line 1: Can we add tests?

pkg/tuf/client.go
```diff
@@ -15,137 +15,184 @@
...
+		// this is most likely due to the lack of metadata files
+		// on disk. Perform a full update and return.
+		return c.Refresh()
 	}
 
-	trustedroot, ok := tufMetaMap[TrustedRootTUFPath]
-	if ok {
-		if ok, _ := validTarget(trustedrootMeta, trustedroot); ok {
-			return trustedroot, nil
+	var tm = c.up.GetTrustedMetadataSet()
```
Hayden-IO:
```suggestion
	tm := c.up.GetTrustedMetadataSet()
```

pkg/tuf/client.go
```diff
@@ -15,137 +15,184 @@
...
 	}
 
-	return writer.Bytes, nil
+	return c.Refresh()
 }
 
-func checkEmbedded(tufRootURL string, fileJSONStore *filejsonstore.FileJSONStore) (json.RawMessage, error) {
-	embeddedRootPath := path.Join("repository", tufRootURL, RootTUFPath)
+// Refresh forces a refresh of the underlying TUF client.
+// As the tuf client does not support multiple refreshes during its
```
Hayden-IO: Out of scope, is this something that's being tracked in go-tuf-metadata?

pkg/tuf/client.go
```diff
@@ -15,137 +15,184 @@
...
+			return nil
+		}
+		// A TUF client refresh will now happen (c.Refresh),
+		// update the mod time for the timestamp.
+		//
+		// Ignore the error here, there is no need to fail
+		// operation only because the file's metadata could
+		// not be updated
+		//nolint:errcheck
+		os.Chtimes(p, time.Now(), time.Now())
```
Hayden-IO: Does this need to happen? Messing with the time of on-disk files feels a bit unexpected

Does `c.up.Refresh` forcefully update the on-disk metadata, as in downloads a new timestamp and writes it to the metadata folder regardless if the timestamp file has changed?

pkg/tuf/client.go
```diff
@@ -15,137 +15,184 @@
...
+func (c *Client) GetTarget(target string) ([]byte, error) {
+	ti, err := c.up.GetTargetInfo(target)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("target %s not found: %w", target, err)
 	}
 
-	return root, nil
-}
+	path, tb, err := c.up.FindCachedTarget(ti, "")
```
Hayden-IO: nit: For this and line 180, can you add `/* name-of-var=*/` for the empty strings for readability?

pkg/tuf/client.go
```diff
@@ -15,137 +15,184 @@
...
+func (c *Client) Refresh() error {
+	var err error
 
-	root, err := embeddedRepos.ReadFile(embeddedRootPath)
+	c.up, err = updater.New(c.cfg)
 	if err != nil {
-		return nil, err
+		return err
 	}
+	return c.up.Refresh()
```
Hayden-IO: How is concurrency handled with calling `Refresh` in paralell? By the underlying TUF repo?

pkg/root/trusted_root.go
```diff
@@ -256,3 +259,101 @@ func NewTrustedRootProtobuf(rootJSON []byte) (*prototrustroot.TrustedRoot, error
...
+// refreshes the trusted root from TUF. This is needed for long-running
+// processes to ensure that the trusted root does not expire.
+type LiveTrustedRoot struct {
+	*TrustedRoot
+	mu sync.RWMutex
+}
+
+// NewLiveTrustedRoot returns a LiveTrustedRoot that will periodically
+// refresh the trusted root from TUF.
+func NewLiveTrustedRoot(opts *tuf.Options) (*LiveTrustedRoot, error) {
```
Hayden-IO, line 296: Maybe for later, do you think this should be built into the default TrustedRoot rather than maintain a separate one? It shouldn't impact performance beyond having a background thread.

Hayden-IO (changes requested):
Super excited to see the spec implemented!

pkg/tuf/options.go
kommendorkapten: Ooops, this comment is wrong. Cache control _is_ working.

pkg/tuf/options.go
```diff
@@ -0,0 +1,81 @@
...
+// the client will *ALWAYS* perform a TUF update.
```
kommendorkapten:
```suggestion
```

pkg/tuf/options.go
kommendorkapten: What would be the benefit of having it as a path, so it can't be modified during runtime? Or to simplify the UX for a client so they don't need to read it from disk themselves?

pkg/tuf/options.go
kommendorkapten: Yeah, this is actually a one. 

The embed operations **does not fail** as long as the correct path is provided, so that removes one case where the function could error.

> If `UserHomeDir()` fails, we could have some default path 

This is the tricky one I believe, as we would need to cover a few OS, which depends on different filesystem layouts. What about init the RUF root in the default temporary directory? This function does not fail. 
See my proposed change.

pkg/tuf/options.go
```diff
@@ -0,0 +1,81 @@
...
+	var opts Options
+	var err error
+
+	opts.Root, err = DefaultRoot()
+	if err != nil {
+		return nil, err
+	}
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return nil, err
```
kommendorkapten:
```suggestion
	home, err := os.UserHomeDir()
	if err != nil {
	        // Fall back to using a repository in the default temp directory
		home = os.TempDir()
```

pkg/tuf/options.go
kommendorkapten: This suggestion: https://github.com/sigstore/sigstore-go/pull/41/files#r1427737332

pkg/tuf/client.go
kommendorkapten: This is a deliberate choice to maintain consistency with the reference implementation (python-tuf) which does not support multiple Refresh invocations: https://github.com/theupdateframework/python-tuf/blob/f711997a08cbb558fc8ab91406a846bbe4883d1a/tuf/ngclient/updater.py#L115

It has been discussed in this issue: https://github.com/rdimitrov/go-tuf-metadata/issues/86

I don't think this is a blocker for go-tuf, as in the majority of cases I would reckon a TUF client is a short lived client, so multiple calls to `Refresh()` would not be needed.

pkg/tuf/client.go
kommendorkapten: No, go-tuf does not modify the disk if the downloaded timestamp is equal to the current one. So this is a hack to implement the non-standard metadata cache Sigstore want to have.

pkg/tuf/client.go
kommendorkapten: No, this should be documented as a method unsafe for concurrent use. For concurrency it appears the `LiveTrustedRoot` implements required serialization via a `Mutex`.

rdimitrov: I'll add an issue for that in go-tuf-metadata so we can work on it 👍

pkg/tuf/options.go
Hayden-IO: This was in the context of avoiding options erroring out on initialization, to have the client handling reading the default root.

pkg/tuf/client.go
Hayden-IO: I can chime in there, but I think the issue will come up if TUF is used by a server rather than client. Not a blocker for Sigstore since the services don't depend on TUF though.

pkg/tuf/client.go
Hayden-IO: What do you think about persisting this state in a separate file? We had something like this for the previous client, to persist the location of the remote repo. Might be helpful in the future if we add more configuration too.

kommendorkapten: Good idea, relying a separate file is a better IMHO!

go.mod
kommendorkapten: This is interesting.

```shell
~/git/sigstore-go % head -3 go.mod
module github.com/sigstore/sigstore-go

go 1.21
~/git/sigstore-go % go mod tidy
~/git/sigstore-go % head -3 go.mod
module github.com/sigstore/sigstore-go

go 1.21.5
````

So running `go mod tidy` updates it, can't remember ever seeing that before.

File: pkg/tuf/options.go (hunk 3 above)
Comment (kommendorkapten) [reply pkg/tuf/options.go:48]: I updated to that `DefaultOptions` is not returning an error, resolving.

File: pkg/tuf/client.go (hunk 12 above)
Comment (kommendorkapten) [reply pkg/tuf/client.go]: Fixed now.

File: pkg/tuf/client.go (hunk 9 above)
Comment (kommendorkapten) [reply pkg/tuf/client.go:72]: Clarified with a comment.

**Comment by ramonpetgrave64** [comment]
fyi, with slsa-verifier, we are planning to use the npmjs delegation, pending this sigstore-go PR and in
https://github.com/slsa-framework/slsa-verifier/pull/731

File: pkg/root/trusted_root.go [hunk 18]
```diff
@@ -256,3 +259,98 @@ func NewTrustedRootProtobuf(rootJSON []byte) (*prototrustroot.TrustedRoot, error
...
+	}
+	tr, err := GetTrustedRoot(client)
+	if err != nil {
+		return nil, err
+	}
+	ltr := &LiveTrustedRoot{
+		TrustedRoot: tr,
+		mu:          sync.RWMutex{},
+	}
+	ticker := time.NewTicker(time.Hour * 24)
```
Hayden-IO, line 309: This should be configurable if metadata expires more frequently than 24 hours.

kommendorkapten:
The `go-tuf-metadata` repo are expected to be the official `v2` package in `go-tuf` soon, so we should wait with this PR for that to happen, to avoid too many changes.

Hayden-IO:
Move  has begun: https://github.com/theupdateframework/go-tuf/pull/583

kommendorkapten:
go-tuf is now updated with the new `v2` package, so I updated this PR to use that.

go.mod
codysoyland: The reason `go mod tidy` is pinning to 1.21.5 is due to the deps tree. I noticed that even after upgrading to go 1.21.6, it will change it to 1.21.5.

If you run `go mod tidy -go=1.21` (in order to specify the intended go version), it will err:

```
go: github.com/theupdateframework/go-tuf/v2@v2.0.0-20240129093820-4e440e28cdf6 requires go@1.21.5, but 1.21 is requested
```

I would suggest we just roll with go 1.21.5 for the `go` directive for now, as we can't run `go mod tidy` otherwise.

pkg/root/trusted_root.go
codysoyland: @kommendorkapten and I have had a few discussions about using functional options, and I think generally I'm *for* using them, but they can have a downside of being less explicit about defaults and can cause indirection/confusion. I think your idea of using methods on the `Option` struct is clear though (ie. `opts := tuf.DefaultOptions().WithRepoBaseURL(url)`), so I'll try adding that.

pkg/root/trusted_root.go
codysoyland: That makes sense to me, however since `tuf.Options.CacheValidity` is using *days* as the unit, I wonder if we should also change that? If that config value is >= 1, then this ticker will have no effect if configured <= 24 hours, as it will just pull the cached data.

codysoyland:
I think this is mostly complete, except for tests. I would be okay with merging before tests are written, as we are pre-1.0 and still wanting to move fast on this. If other reviewers disagree, we can spend the time to add some tests based on the [basic repo example from go-tuf](https://github.com/theupdateframework/go-tuf/blob/master/examples/repository/basic_repository.go).

pkg/root/trusted_root.go
codysoyland: That's a fair assessment! The `LiveTrustedRoot` was kind of a shower thought I had as we were about to open-source it, but if others agree, we can definitely make this the default to reduce API surface and ensure that the default client is compatible with long-running processes. I agree that it doesn't affect performance too much. Just a mutex hit and background thread.

codysoyland: Just realized there's a bit of a foot-gun with this func... If a long-running process calls `NewLiveTrustedRoot` for every verification (if the author doesn't realize it auto-updates), it will spawn goroutines for each invocation, so the clients would not get garbage collected and would gradually fill up memory and threads.

go.mod
ramonpetgrave64: I'm a bit new to go, but If you do `go mod vendor` and search for `1.21.5`, it will generate a local `./vendor/modules.txt` that will show that it's `go-tuf-metadata` that explicitly requires `1.21.5`

* https://github.com/rdimitrov/go-tuf-metadata/blob/df718890900a95dccb025e4fb5e28641ee66f108/go.mod#L3

```
## explicit; go 1.21.5
github.com/rdimitrov/go-tuf-metadata/metadata
github.com/rdimitrov/go-tuf-metadata/metadata/config
github.com/rdimitrov/go-tuf-metadata/metadata/fetcher
github.com/rdimitrov/go-tuf-metadata/metadata/trustedmetadata
github.com/rdimitrov/go-tuf-metadata/metadata/updater
```

ramonpetgrave64:
@codysoyland No strong preference, but I would like to at least have some unit tests that are closer to end-to-end tests.

go.mod
rdimitrov: Opened a PR to resolve this (iirc it was motivated by a warning from `govulncheck`)- https://github.com/theupdateframework/go-tuf/pull/588 👍

kommendorkapten: @rdimitrov @codysoyland @ramonpetgrave64 once the PR in go-tuf is merged, I'll update this to use that (without pinning to minor version).

Hayden-IO:
I'll do another pass over the PR this week.

I would very much prefer tests, given the criticality of this client. I would take a look at https://github.com/sigstore/sigstore/blob/main/pkg/tuf/client_test.go, you might be able to adapt these unit tests for this.

go.mod
kommendorkapten: Updated now, go mod is `1.21`.

pkg/tuf/client.go
kommendorkapten: Yeah I think so, I'll work on getting some in.

codysoyland:
> I would very much prefer tests, given the criticality of this client. I would take a look at https://github.com/sigstore/sigstore/blob/main/pkg/tuf/client_test.go, you might be able to adapt these unit tests for this.

Totally fair, I'll try to carve out some time to write tests this week.

codysoyland:
I added a basic test to initialize a TUF repo and client in https://github.com/sigstore/sigstore-go/pull/41/commits/96326fa3f01fa28463ec97367d68d80e3c3dc2c1.

I was hoping to use the [`RepositorySimulator` from go-tuf](https://github.com/theupdateframework/go-tuf/blob/master/testutils/simulator/repository_simulator.go) to create the repo, but I think there are some bugs with the target support in that utility, and it looks like the authors [intend to move it to an internal package](https://github.com/theupdateframework/go-tuf/pull/601). So using the [basic_repository.go example](https://github.com/theupdateframework/go-tuf/blob/master/examples/repository/basic_repository.go) as a starting point, I roughly created a TUF repo that I can use in tests. I also partially implemented the Fetcher interface, so we don't need to open a port for tests.

Assuming I'm on the right path, I can continue to add more tests based on this to validate the caching behavior.

/cc @kommendorkapten

pkg/tuf/config.go
```diff
@@ -45,7 +45,7 @@ func (c *Config) Persist(p string) error {
 	if err != nil {
 		return fmt.Errorf("failed to JSON marshal config: %w", err)
 	}
-	err = os.WriteFile(p, b, 0400) // Read only by current user
+	err = os.WriteFile(p, b, 0600)
```
codysoyland, line 48: @kommendorkapten FYI, I changed this from 0400 to 0600 as future calls to Persist will fail otherwise.

codysoyland:
I added tests for the `tuf` package, so we're up to 88.0% statement coverage in that package.

It turns out that writing some tests has already paid off as I was able to identify a few bugs and dead code.

pkg/tuf/config.go
kommendorkapten: Ah, yes of course. Good find!

kommendorkapten (dismissed):
Nice cody!

@haydentherapper Have some time to give this a review?

pkg/tuf/client.go
```diff
@@ -15,137 +15,186 @@
...
+	// client (only reads content on disk) and then decide if we
+	// must perform a full TUF update.
+	tmpCfg := *c.cfg
+	// Create a temporary config for the first use where UnsafeLocalMode
+	// is true. This means that when we first initialize the client,
+	// we are guaranteed to only read the metadata on disk.
+	// Based on that metadata we take a decision if a full TUF
+	// refresh should be done or not. As so, the tmpCfg is only needed
+	// here and not in future invocations.
+	tmpCfg.UnsafeLocalMode = true
```
Hayden-IO, line 72: To confirm, this means that if the timestamp along with all other metadata is not expired, then we will not make any network connections, correct?

Hayden-IO:
Yep reviewing now!

examples/oci-image-verification/go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/sigstore/sigstore-go/examples/oci-image-verification
 
-go 1.21
+go 1.21.5
```
Hayden-IO: Can this be reverted to go 1.21?

pkg/root/trusted_root.go
Hayden-IO: LGTM, thanks for adding these options!

pkg/root/trusted_root.go
Hayden-IO: Good with leaving these separate!

pkg/root/trusted_root.go
Hayden-IO: To confirm, cache validity ignores whether the metadata is still valid?

Maybe it should be max(24 hours, CacheValidity)? If cache validity is set, just refresh once the cache expires, otherwise refresh every 24 hours?

pkg/tuf/client.go
```diff
@@ -15,137 +15,186 @@
...
+		c.opts.CachePath,
+		fmt.Sprintf("%s.json", URLToPath(c.opts.RepositoryBaseURL)),
+	)
+
+	return p
+}
+
+// Refresh forces a refresh of the underlying TUF client.
+// As the tuf client does not support multiple refreshes during its
+// life-time, this will replace the TUF client with a new one.
```
Hayden-IO: Confirming, when we say replacing the TUF client, this means "updater.New"?

pkg/tuf/options.go
```diff
@@ -0,0 +1,135 @@
...
+	var opts Options
+	var err error
+
+	opts.Root = DefaultRoot()
+	home, err := os.UserHomeDir()
+	if err != nil {
+		// Fall back to using a TUF repository in the temp location
+		home = os.TempDir()
+	}
+	opts.CacheValidity = 1
```
Hayden-IO: Should we have a cache by default? I think cache is expected to be used in the case of airgapped environments, so should we have this be 0 and just rely on timestamp expiration for when to fetch metadata by default?

Hayden-IO (commented):
Just a few questions! This looks really good, nice work on the tests.

pkg/tuf/client.go
kommendorkapten: Correct.

examples/oci-image-verification/go.mod
kommendorkapten: Oh, I think I missed that, probably only updated the top-level `go.mod`. I'll take a look.

pkg/tuf/client.go
kommendorkapten: Yes, I'll clarify that.

pkg/tuf/options.go
kommendorkapten: Back in the summer of 2022 when we where discussing this, 1 day was proposed https://docs.google.com/document/d/1QWBvpwYxOy9njAmd8vpizNQpPti9rd5ugVhji0r3T4c/edit#heading=h.s9pcbj5k0o1z

In retrospect, I think that the tuf package maybe should set the cache validity to 0 (i.e no cache) but each client should set the value to the best in their domain. This way we minimize any possible footguns (i.e secure by default). Even if a single is quick, I could still see possible issues being opened when the client (e.g. cosign) does not immediately get the new root. 

I can change this. cc @codysoyland

Hayden-IO: How does this interact with the target cache files? I don’t believe it does, correct? As in, the client will still cache targets on disk regardless of this configuration.

kommendorkapten: Reading this comment again:

> Should we have a cache by default? I think cache is expected to be used in the case of airgapped environments, so should we have this be 0 and just rely on timestamp expiration for when to fetch metadata by default?

This is not really how it works. If caching is not used, the client will always try to get a new timestamp and root, regardless if the local timestamp is expired or not (this is according to the TUF spec). 

The cache allows us to control if this (full TUF online refresh) should be performed on every invocation, or only once per day. But I'm still thinking that maybe having the cache off by default is a better option, as a full TUF update is not that expensive, if the client is up to date, it's two egress HTTP requests (one for root and one for timestamp). So the cache is not really with arigapped, it's more a way to speed up invocations where e.g. frequent cosign verify is performed in a short period of time.

kommendorkapten: > How does this interact with the target cache files? I don’t believe it does, correct? As in, the client will still cache targets on disk regardless of this configuration.

Exactly, it's only the metadata that is cached in non-standard way to avoid a full metadata refresh against the remote TUF repository.

kommendorkapten:
I broke some tests when I updated to latest go-tuf, I will fix that.

pkg/tuf/options.go
Hayden-IO: Since we're using the `UnsafeLocalMode` updater for the initial update, won't cache validity only matter if the timestamp is expired?

Hayden-IO: Also, should cache validity be checked in `GetTarget` too? Otherwise `GetTargetInfo` forces a refresh if metadata is expired.

Hayden-IO: I might be wrong about the above, so maybe I should ask that as a question: In what cases is the underlying tuf client implicitly refreshing data? (Edit: Ah I remember, the updater only refreshes once, so the answer should be never)

kommendorkapten: > Since we're using the `UnsafeLocalMode` updater for the initial update, won't cache validity only matter if the timestamp is expired?

No, timestamp expiration is always respected. As the timestamp may expire at any time, or a new root (or target) can be published, so the client always looks for that on every invocation. This is the behaviour that's controlled via `CacheValidity`

> Also, should cache validity be checked in GetTarget too? Otherwise GetTargetInfo forces a refresh if metadata is expired.

Yes, if the metadata is expired, this will happen. If this is in an airgapped environment this would fail, but for that, I would think a better approach is to just provide the `trusted_root.json` instead of using TUF, as we won't get any real security benefit from it. And also, I would think that the normal scenario is that a client does:
1. Create a TUF client
2. Get `trusted_root.json` target
and the two calls happens during initialization, pretty close to each other in time. And once the `trusted_root.json` is retrieved, it's used for the rest of the program's execution?

Hayden-IO: Ah, I took another pass over the code and now see what I was missing. In the current Sigstore TUF client, we have it set up where if the timestamp is not expired, no additional metadata is fetched. With the current implementation, if you wanted to replicate that behavior, you would set `CacheValidity` to be greater than the timestamp expiration - Then it would not refresh until the timestamp expires, which will then force a refresh.

Have you considered having this behavior controlled via a boolean rather than a time period? The current implementation does give finer control though. One idea to improve UX, maybe a NO_CACHE and MAX_CACHE constant, set to 0 and MAX_INT respectively? Then when setting up the config, you can choose to a) follow the TUF spec with NO_CACHE, b) only refresh when absolutely necessary with MAX_CACHE, or c) fine-tune based on needs.

I might also add a comment making this explicit that `CacheValidity` is not for airgapped environments, since it still does respect timestamp expiration.

Hayden-IO (dismissed):
Left one comment about how to clarify the purpose of cache validity and use consts, but it's not a blocker. Thanks for all of the work on this!

pkg/tuf/options.go
codysoyland: > Have you considered having this behavior controlled via a boolean rather than a time period? 

I can't speak for @kommendorkapten, but in my opinion, the `CacheValidity` flag works well for systems that perform a lot of verifications, without having to manually bump the cache daily, and avoids the risk of missing new metadata if the repo's timestamp interval is long. However, I could see the argument of changing it to hours or time.Duration, or adding a few constants as you suggested.

> I might also add a comment making this explicit that CacheValidity is not for airgapped environments, since it still does respect timestamp expiration.

What is your opinion on the correct behavior for airgapped environments? Would you suggest having an option to disregard an expired timestamp? That is certainly possible, but I would probably want to slap the `Insecure` prefix in front of that option. 😆

Hayden-IO: > CacheValidity flag works well for systems that perform a lot of verifications, without having to manually bump the cache daily, and avoids the risk of missing new metadata if the repo's timestamp interval is long

This sounds good. I think the constants are a nice compromise for those who are configuring the client and want a simple "boolean" approach of "no cache" vs "cache up to timestamp expiration".

> What is your opinion on the correct behavior for airgapped environments? Would you suggest having an option to disregard an expired timestamp?

I think we should omit airgapped support for now, there's much more nuance with getting this right, given it doesn't follow the TUF spec. I agree with @kommendorkapten that for airgapped clients, a reasonable approach is to fetch the trust root out of band. The python client has been thinking about what it means to be [truly offline](https://github.com/sigstore/sigstore-python/issues/483) too.

pkg/root/trusted_root.go
Hayden-IO: Thinking more on this, I think it's simpler to leave this as-is and let the client initialization handle the check for cache validity. Worst case, it just means an unnecessary reinitialization of the TUF client from disk.

pkg/tuf/options.go
codysoyland: Got it, I missed @kommendorkapten's comment on that, but I agree with you both!

Hayden-IO (approved):
🎉

PR #42, opened 2023-12-12
Have conformance test require signed timestamps for bundles v02

steiza:
#### Summary

This is a bit of a philosophical question, although either `sigstore-go` needs to behave this way or we need to change `test_verify_rejects_bad_tsa_timestamp` that was added in https://github.com/sigstore/sigstore-conformance/pull/112.

In that conformance test, we have a bundle with media type v0.2 that includes `timestampVerificationData` (so far so good). But it also includes an `inclusionPromise.signedEntryTimestamp`, and it expects verification to fail when `timestampVerificationData` is incorrect but `inclusionPromise.signedEntryTimestamp` is correct.

This change makes it so v0.2 bundles require valid `timestampVerificationData` and for those bundles ignores `inclusionPromise.signedEntryTimestamp`. It's possible that instead we want to require any information provided to be valid.

#### Release Note

NONE

#### Documentation

N/A

### DIFF
--- a/cmd/conformance/main.go
+++ b/cmd/conformance/main.go
@@ -222,7 +222,19 @@ func main() {
 		tr := getTrustedRoot()
 
 		// Verify bundle
-		sev, err := verify.NewSignedEntityVerifier(tr, verify.WithTransparencyLog(1), verify.WithSignedCertificateTimestamps(1))
+		verifierConfig := []verify.VerifierOption{}
+		verifierConfig = append(verifierConfig, verify.WithSignedCertificateTimestamps(1))
+
+		switch b.Bundle.MediaType {
+		case bundle.SigstoreBundleMediaType01:
+			verifierConfig = append(verifierConfig, verify.WithTransparencyLog(1))
+		case bundle.SigstoreBundleMediaType02:
+			verifierConfig = append(verifierConfig, verify.WithSignedTimestamps(1))
+		default:
+			log.Fatalf("Unknown bundle media type: %s", b.Bundle.MediaType)
+		}
+
+		sev, err := verify.NewSignedEntityVerifier(tr, verifierConfig...)
 		if err != nil {
 			log.Fatal(err)
 		}
--- a/cmd/conformance/main.go
+++ b/cmd/conformance/main.go
@@ -221,17 +221,23 @@ func main() {
 		// Load trust root
 		tr := getTrustedRoot()
 
-		// Verify bundle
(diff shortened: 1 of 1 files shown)


Hayden-IO:
Slack thread: https://sigstore.slack.com/archives/C03SZ6SHU90/p1702418056454379

It should be possible to have both a signed TSA timestamp and a SignedEntryTimestamp. To me, this is the same as having multiple signed TSA timestamps. A verifier will use its roots of trust to determine which timestamps to trust.

I'm thinking about treating the SET the same as the signed timestamp, removing the ability to configure this in policy, and instead relying on the configured roots of trust. This has some open questions: Do we simply ignore any verification failures and just search for one valid timestamp? Also, should verifiers be able to configure trusting a transparency log for timestamps vs inclusion proofs?

kommendorkapten:
To re-iterate what I wrote in the slack thread:

@haydentherapper wrote:
> A verifier will use its roots of trust to determine which timestamps to trust.

What I believe is that all signatures not under the closure of the trust root **should be ignored**.

steiza:
The existing conformance tests supply the trusted root and the bundle (not an explicit policy), but with these changes the conformance driver can infer what the policy should be based on what's in the trusted root and the bundle.

cmd/conformance/main.go
```diff
@@ -221,12 +221,31 @@ func main() {
...
+		verifierConfig = append(verifierConfig, verify.WithSignedCertificateTimestamps(1))
+
+		// Check bundle and trusted root for signed timestamp information
+		bundleTimestamps, err := b.Timestamps()
+		if err != nil {
+			fmt.Println(err)
+			os.Exit(1)
+		}
+
+		if len(tr.TSACertificateAuthorities()) > 0 && len(bundleTimestamps) > 0 {
```
Hayden-IO, line 234: Mentioned this on slack, but what if the trust root specifies a TSA certificate and the bundle contains a signed timestamp issued by a different TSA cert? This conformance check expects that there is at least one 1-1 mapping between trusted TSA certs and signed timestamps.

As is I think this is fine for now, because the "fix" is another policy flag to say that the verifier just needs a timestamp.

steiza: Yes, I think the scope of this PR is "allow bundles relying on valid signed timestamps from a TSA to pass conformance testing", which previously was not the case.

Hayden-IO (approved):

Issue #43, opened 2023-12-13
Timestamp verification should not fail unless the threshold is not met


Hayden-IO:
**Description**

Looking at https://github.com/sigstore/sigstore-go/blob/main/pkg/verify/tsa.go#L64, if any timestamp verification fails, then the verifier errors out. Instead, the verifier should succeed as long as the threshold for successful timestamp signatures is met. If there's N timestamps, and the threshold is 1, then the verifier only needs 1 timestamp to verify to succeed.

phillmv:
When writing out that function, my interpretation was:

- suppose you have N timestamps
- you may stop verifying timestamps once you hit threshold M < N

But the question of how to handle invalid timestamps was curious.

Adopting the perspective of, well, writing an attestation/sigstore bundle store where we prize data integrity above all else, I found myself asking:

How exactly would you end up with an invalid timestamp embedded in the bundle? Either you received a timestamp that is outside of the validity period, which the signer should have rejected, or you used an "Observer" whose key is not part of the verifier's trusted materials.

---

In every other part of the spec we're careful to fail as early as possible and to stop processing the data as soon as we encounter an error. Initially, we were tempted to do the opposite, and try to muddle along the verification process so we could generate more meaningful errors: what if you have not just 1 but 2 or 3 different bugs in your signer?

But after meditating on this behaviour, I decided that it was wise to bail early: continuing to process the data after we entered invalid state would invite subtle bugs or exploits further down the road that could be used to compromise or fake verification. 

So, on a similar note, I ask: is this a bug?

Under what circumstances are you legitimately creating a bundle with either a timestamp in an invalid range, or asking the verifier to verify a signature whose trusted materials we do not possess? On a balance of probabilities, it feels like it's more likely that _shenanigans_ are happening.

I'm not wedded to this behaviour!, but the idea of aggressively rejecting shenanigans (aka weirdly shaped bundles) is appealing.

Could we narrow down the undesirable behaviour? Like, is it permissible to have a (signature-verified) timestamp that is outside of the validity period [which we do not count towards the threshold], but we fail early if we have a signature we can't verify?

I forget if the spec explains what the use case is for having M of N timestamp thresholds. Are we hoping to have multiple transparency logs whose keys are not always available?

steiza:
[...]

PR #45, opened 2023-12-14
Don't fail on TSA signatures we can't verify

steiza:
#### Summary

See https://github.com/sigstore/sigstore-go/issues/43

As long as there are enough TSA signatures (that we can verify) to meet the threshold specified by the verification policy, we say that the bundle is valid.

#### Release Note

NONE

#### Documentation

N/A

### DIFF
--- a/pkg/verify/tsa_test.go
+++ b/pkg/verify/tsa_test.go
@@ -40,33 +40,33 @@ func TestTimestampAuthorityVerifier(t *testing.T) {
 	_, err = verify.VerifyTimestampAuthority(entity, virtualSigstore2, 1)
 	assert.Error(t, err) // different sigstore instance should fail to verify
 
-	otherEntity, err := virtualSigstore2.Attest("foo@fighters.com", "issuer", []byte("statement"))
+	untrustedEntity, err := virtualSigstore2.Attest("foo@fighters.com", "issuer", []byte("statement"))
 	assert.NoError(t, err)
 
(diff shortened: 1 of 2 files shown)


Hayden-IO (dismissed):
Looks good!

We have the same fail-fast vs fail-after-threshold for tlogs - https://github.com/sigstore/sigstore-go/blob/main/pkg/verify/tlog.go#L85-L161
I can make this update, or would you like to?

pkg/verify/tsa_test.go
```diff
@@ -39,6 +39,34 @@ func TestTimestampAuthorityVerifier(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	_, err = verify.VerifyTimestampAuthority(&oneGoodOneBadTimestampEntity{entity, otherEntity}, virtualSigstore, 1)
+	assert.NoError(t, err)
+
+	_, err = verify.VerifyTimestampAuthority(&oneGoodOneBadTimestampEntity{entity, otherEntity}, virtualSigstore, 2)
+	assert.Error(t, err) // only 1 good should not meet threshold of 2
+}
+
+type oneGoodOneBadTimestampEntity struct {
```
Hayden-IO: tiny nit would be calling this "goodAndUntrustedTimestampEntity" rather than "bad"

Hayden-IO:
Rekor is done in https://github.com/sigstore/sigstore-go/pull/47 (wanted to code some today :) )

Hayden-IO (approved):

phillmv (approved):

PR #47, opened 2023-12-15
Don't fail on Rekor entry verification for untrusted entries

### LINKED ISSUE #43: Timestamp verification should not fail unless the threshold is not met

Hayden-IO:
Like #45, as long as the threshold of expected timestamps is met, then verification should succeed. Otherwise, entries without trust root material should be skipped.

One benefit of having the log key ID be used to look up the correct trust root material is that we can still error out if the signature is invalid.

Fixes #43

### DIFF
--- a/pkg/verify/sct.go
+++ b/pkg/verify/sct.go
@@ -29,6 +29,7 @@ import (
 // leaf certificate, will extract SCTs from the leaf certificate and verify the
 // timestamps using the TrustedMaterial's FulcioCertificateAuthorities() and
 // CTlogAuthorities()
+// TODO(issue#46): Add unit tests
 func VerifySignedCertificateTimestamp(leafCert *x509.Certificate, threshold int, trustedMaterial root.TrustedMaterial) error { // nolint: revive
 	ctlogs := trustedMaterial.CTlogAuthorities()
 	fulcioCerts := trustedMaterial.FulcioCertificateAuthorities()
@@ -48,7 +49,8 @@ func VerifySignedCertificateTimestamp(leafCert *x509.Certificate, threshold int,
 		encodedKeyID := hex.EncodeToString(sct.LogID.KeyID[:])
 		key, ok := ctlogs[encodedKeyID]
 		if !ok {
-			return fmt.Errorf("unable to find ctlogs key for %s", encodedKeyID)
+			// skip entries the trust root cannot verify
+			continue
 		}
 
 		for _, fulcioCa := range fulcioCerts {
--- a/pkg/verify/tlog.go
+++ b/pkg/verify/tlog.go
@@ -77,7 +77,9 @@ func VerifyArtifactTransparencyLog(entity SignedEntity, trustedMaterial root.Tru
 		}
 
 		if !online {
-			// TODO: do we validate that an entry has EITHER a promise OR a proof?
+			if !entry.HasInclusionPromise() && !entry.HasInclusionProof() {
+				return nil, fmt.Errorf("entry must contain and inclusion proof and/or promise")
+			}
 			if entry.HasInclusionPromise() {
 				err = tlog.VerifySET(entry, trustedMaterial.TlogAuthorities())
 				if err != nil {
--- a/pkg/verify/tlog.go
+++ b/pkg/verify/tlog.go
@@ -55,9 +55,6 @@ func VerifyArtifactTransparencyLog(entity SignedEntity, trustedMaterial root.Tru
 			}
 		}
 	}
-	if len(entries) < threshold {
-		return nil, fmt.Errorf("not enough transparency log entries: %d < %d", len(entries), threshold)
-	}
 
 	sigContent, err := entity.SignatureContent()
 	if err != nil {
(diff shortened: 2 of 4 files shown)


pkg/verify/sct.go
```diff
@@ -48,7 +49,8 @@ func VerifySignedCertificateTimestamp(leafCert *x509.Certificate, threshold int,
 		encodedKeyID := hex.EncodeToString(sct.LogID.KeyID[:])
 		key, ok := ctlogs[encodedKeyID]
 		if !ok {
-			return fmt.Errorf("unable to find ctlogs key for %s", encodedKeyID)
+			// skip entries the trust root cannot verify
```
steiza, line 52: This is definitely the behavior we want, but we are losing a lot of valuable feedback from the library as it verifies the bundle, which has been very helpful in debugging!

I wonder if we should add a debug mode to sigstore-go, where it still prints these messages out, but continues / doesn't error out.

pkg/verify/tlog.go
```diff
@@ -78,14 +75,14 @@ func VerifyArtifactTransparencyLog(entity SignedEntity, trustedMaterial root.Tru
 		if err != nil {
 			return nil, err
 		}
+		inclusionVerified := false
```
steiza: Minor: the reason this was scoped to `!online` before was because we want to ensure there was at least one of an inclusion promise or an inclusion proof (e.g. you couldn't cheat by having neither, and falling through to have your timestamp be listed as verified).

I think with the `continue` statements, and with the placement of `if inclusionVerified` inside the `for` loop, we don't actually need it for the `online` case? But also, it's not hurting anything.

pkg/verify/tlog_test.go
```diff
@@ -52,6 +52,47 @@ func TestTlogVerifier(t *testing.T) {
 	assert.Error(t, err)
 }
 
+type goodAndUntrustedLogEntry struct {
```
steiza: Minor: I updated the wording on https://github.com/sigstore/sigstore-go/pull/45/files that you could consider adopting here: `oneTrustedOneUntrustedTimestampEntity`, `UntrustedTestEntity`, etc.

steiza (commented):
Generally looks good to me! Looking for other folk's opinion about debug statements / a debug mode.

pkg/verify/sct.go
kommendorkapten: Something like an addition to the `VerificationResult`, were ignored entries are stored? Could of course be a separate struct too. Then a client may opt to offer a debug or verbose mode where such information is shown?

kommendorkapten:
Looks good, and I agree with @steiza that we may want to have some option to get details on ignored signatures. I'm fine to merge this then track that as a separate issue as it would be a larger refactor (to cover TSAs too).

kommendorkapten (dismissed):

pkg/verify/tlog_test.go
Hayden-IO: done

pkg/verify/sct.go
Hayden-IO: Good suggestion! Tracking in https://github.com/sigstore/sigstore-go/issues/48 if that's alright.

pkg/verify/tlog.go
Hayden-IO: Yep, I think this looks cleaner without. Lemme know if I misunderstood something and removed too much.

Hayden-IO: I miss `goto` statements. I genuinely think they're cleaner than `continue`....

pkg/verify/tlog.go
```diff
@@ -106,19 +103,15 @@ func VerifyArtifactTransparencyLog(entity SignedEntity, trustedMaterial root.Tru
...
 					return nil, err
 				}
-
-				inclusionVerified = true
-			}
-
-			if inclusionVerified {
-				verifiedTimestamps = append(verifiedTimestamps, entry.IntegratedTime())
 			}
+			verifiedTimestamps = append(verifiedTimestamps, entry.IntegratedTime())
```
steiza, line 109: Ah, so we don't want to append to `verifiedTimestamps` in the `if !entry.HasInclusionPromise() && !entity.HasInclusionProof()` case; we want to make sure at least one of these is true, so I think we still need `inclusionVerified` around this part.

Hayden-IO: @steiza Good catch! Instead of adding `inclusionVerified` back, I fixed the TODO to check that one of these is true, since it would be an invalid entry if neither was.

I added two tests to catch this, one that creates an invalid entry without a proof or promise and one without any entries.

Hayden-IO:
@steiza Just a post-holiday bump to review! The change is also in https://github.com/sigstore/sigstore-go/pull/51

steiza (approved):
Nice work! The test to ensure we don't "fall through" and verify if there's no inclusion proof and no inclusion promise is great.

PR #51, opened 2023-12-18
Add option to unify signed and log timestamps

### LINKED ISSUE #44: Unifying timestamp verification across SETs and TSA signed timestamps
**Description**

With the current spec, a client specifies how many SETs and how many timestamps are expected. The current implementation expects that every timestamp that is present is verifiable by material in the trusted root. We propose changing this in two ways: First, ignoring any timestamps that cannot be verified, as long as the threshold is met (noted in the bug below). Two, treating SETs and TSA signed timestamps as equals for fetching verified timestamp.

One approach for this is to expose a policy configuration that states a timestamp can be obtained from either an SET or TSA signed timestamp. This could be via a policy flag that states a threshold must be met for either SETs or signed timestamps. Personally, I like this approach of unifying the two, as SETs no longer are used as a proof of inclusion and instead are just a signed timestamp from the log. This also opens us up to adding timestamp providers, like Roughtime.

Related bug: https://github.com/sigstore/sigstore-go/issues/43 - Once fixed, as long as some number of timestamps are found that meet the threshold, verification will continue, and any failures will be ignored.

Hayden-IO:
#### Summary

This adds the WithObserverTimestamps option to verify that a threshold is
met using both RFC3161 signed timestamps and log integrated timestamps.

This also adds WithIntegratedTimestamps to verify a threshold for log integrated
timestamps.

This updates the verification API for the log to only verify log
inclusion proofs or SETs, so that log and timestamp verification are not
conflated.

This also fixes a bug where the integrated time is used with an
inclusion proof. This is not safe to do, since the integrated time is
not authenticated metadata without a SignedEntryTimestamp signature.

Fixes https://github.com/sigstore/sigstore-go/issues/44

### DIFF
--- a/cmd/conformance/main.go
+++ b/cmd/conformance/main.go
@@ -237,7 +237,7 @@ func main() {
 
 		// Check bundle and trusted root for Tlog information
 		if len(tr.TlogAuthorities()) > 0 && b.HasInclusionPromise() {
-			verifierConfig = append(verifierConfig, verify.WithTransparencyLog(1))
+			verifierConfig = append(verifierConfig, verify.WithTransparencyLog(1), verify.WithObserverTimestamps(1))
 		}
 
 		sev, err := verify.NewSignedEntityVerifier(tr, verifierConfig...)
--- a/pkg/verify/signed_entity.go
+++ b/pkg/verify/signed_entity.go
@@ -44,10 +44,12 @@ type VerifierConfig struct { // nolint: revive
 	// signedTimestampThreshold is the minimum number of verified
 	// RFC3161 timestamps in a bundle
 	signedTimestampThreshold int
-	// trustIntegratedTime, when used along with requireObserverTimestamps,
-	// lets log integrated timestamps count towards the observer timestamp
-	// threshold
-	// trustIntegratedTime bool
+	// requireIntegratedTimestamps requires log entry integrated timestamps to
+	// verify short-lived certificates
+	requireIntegratedTimestamps bool
+	// integratedTimeThreshold is the minimum number of log entry
+	// integrated timestamps in a bundle
+	integratedTimeThreshold int
 	// requireObserverTimestamps requires RFC3161 timestamps and/or log
 	// integrated timestamps to verify short-lived certificates
 	requireObserverTimestamps bool
@@ -159,6 +161,17 @@ func WithTransparencyLog(threshold int) VerifierOption {
 	}
 }
 
+// WithIntegratedTimestamps configures the SignedEntityVerifier to
+// expect log entry integrated timestamps from either SignedEntryTimestamps
+// or live log lookups.
+func WithIntegratedTimestamps(threshold int) VerifierOption {
+	return func(c *VerifierConfig) error {
+		c.requireIntegratedTimestamps = true
+		c.integratedTimeThreshold = threshold
+		return nil
+	}
+}
+
 // WithSignedCertificateTimestamps configures the SignedEntityVerifier to
 // expect the Fulcio certificate to have a SignedCertificateTimestamp, and
 // verify it using the TrustedMaterial's CTLogAuthorities().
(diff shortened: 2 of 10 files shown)


Hayden-IO:
@steiza @kommendorkapten @phillmv PTAL (at commit [1a0fa2b](https://github.com/sigstore/sigstore-go/pull/51/commits/1a0fa2ba0c8e3fecfc955d6a1de0163adaa76257), PR will be rebased once the other PR is merged)

The summary of the various options is in https://github.com/sigstore/sigstore-go/issues/44#issuecomment-1861337720

This is a WIP, looking for early comments. Remaining TODOs:

- [x] Fix validation logic for what combo of options are allowed
- [x] Add tests for WithObserverTimestamp
- [x] Clean up TODOs I left in code (documentation)
- [ ] Separate PR: ~Add configuration to not count SETs towards log threshold (Default: false for v0.2 bundles, true for v0.1)~

Also, this fixes a bug during log entry verification. When an inclusion proof is verified, the integrated timestamp was marked as verified. This is unsafe, because the integrated timestamp is unauthenticated metadata. It can only be trusted when signed over in the case of an SET. Given this library is not widely used, I don't think this needs a CVE (unless you think otherwise), but we should at least ping those who we know use this to upgrade.

pkg/verify/signed_entity.go
```diff
@@ -35,13 +35,40 @@ type SignedEntityVerifier struct {
...
 
 type VerifierConfig struct { // nolint: revive
-	performOnlineVerification          bool
-	weExpectSignedTimestamps           bool
-	signedTimestampThreshold           int
-	weExpectTlogEntries                bool
-	tlogEntriesThreshold               int
-	weExpectSCTs                       bool
-	ctlogEntriesThreshold              int
+	// performOnlineVerification queries logs during verification.
```
kommendorkapten, line 38: LOVE THE ADDED COMMENTS ❤️

pkg/verify/tlog_test.go
```diff
@@ -48,8 +59,114 @@ func TestTlogVerifier(t *testing.T) {
...
+
+	return append(entries, otherEntries...), nil
+}
+
+func TestIgnoredTLogEntries(t *testing.T) {
+	statement := []byte(`{"_type":"https://in-toto.io/Statement/v0.1","predicateType":"customFoo","subject":[{"name":"subject","digest":{"sha256":"deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef"}}],"predicate":{}}`)
+
+	virtualSigstore, err := ca.NewVirtualSigstore()
+	assert.NoError(t, err)
+	entity, err := virtualSigstore.Attest("foo@fighters.com", "issuer", statement)
```
kommendorkapten: That is a good email address 🎸

kommendorkapten (dismissed):
Looks good 🚀

Hayden-IO:
Thank you @kommendorkapten for the discussions and reviews!

@steiza any comments?

pkg/verify/tlog_test.go
Hayden-IO: Borrowed that from other tests, I laughed when I first saw it

Hayden-IO:
Rebasing against HEAD, no changes besides go.mod updates (https://github.com/sigstore/sigstore-go/compare/0c43326f2ecf7d6d6a294c9391ff7831cb173bcd..8684030bcc34417d4d61e94c804e8e3a6ca21f12)

cmd/conformance/main.go
```diff
@@ -237,7 +237,7 @@ func main() {
 
 		// Check bundle and trusted root for Tlog information
 		if len(tr.TlogAuthorities()) > 0 && b.HasInclusionPromise() {
-			verifierConfig = append(verifierConfig, verify.WithTransparencyLog(1))
+			verifierConfig = append(verifierConfig, verify.WithTransparencyLog(1), verify.WithObserverTimestamps(1))
```
steiza, line 240: Minor: In trying to make sure I understand this pull request, I'm wondering if this should be `verify.WithIntegratedTimestamps(1)` instead.

Generally, I think we want people to use `WithObserverTimestamps()` and not have to reason about the specifics of `WithIntegratedTimestamps()` or `WithSignedTimestamps()`.

But I thought what we were saying on https://github.com/sigstore/sigstore-go/pull/42 is that the conformance test will infer the verification policy from the trust root, and if the trust root has `tlogs` entries, we should rely on them both for inclusion and for establishing a timestamp?

I guess when inferring the policy from the trust root, there isn't a way to differentiate today. This isn't a problem really, but it does highlight the complexity of the problem space generally!

steiza (approved):
I think this change gives us the flexibility to cover almost any use case.

I do worry about the complexity of the interface, but I think for many (most?) users, they should just use `WithObserverTimestamps()` and not need to dive into `WithIntegratedTimestamps()` vs `WithSignedTimestamps()`.

One minor comment, but I don't feel strongly that the current implementation need to change.

Hayden-IO:
Thanks for the review @steiza! Agreed that the goal is that most if not all users would leverage `WithObserverTimestamps`. The other options would be for those who have specific requirements, eg only trusting a timestamp authority. At some point in the future, we might be able to deprecate `WithIntegratedTimestamps` if the log requires timestamps on entry inclusion.

cmd/conformance/main.go
Hayden-IO: You're right, this should be `WithIntegratedTimestamps` since this is inferring a policy based on if a promise is present. I'll update in a follow up.

In order to infer `WithObserverTimestamps`, which would be more permissive, it would probably look something like a check for either bundled timestamps or SETs. We could do that here, but we can be more precise with `WithSignedTimestamps` and `WithIntegratedTimestamps`.

PR #67, opened 2024-01-03
Update policy inference, add flags for OCI example

### LINKED ISSUE #62: Add flag for observer timestamps to OCI example
**Description**

Tracking so it's not forgotten. Blocked by https://github.com/sigstore/sigstore-go/pull/51

Hayden-IO:
Follow up from #51

Fixes #62


examples/oci-image-verification/main.go
```diff
@@ -68,6 +70,8 @@ func init() {
 	expectedSAN = flag.String("expectedSAN", "", "The expected identity in the signing certificate's SAN extension")
 	expectedSANRegex = flag.String("expectedSANRegex", "", "The expected identity in the signing certificate's SAN extension")
 	requireTSA = flag.Bool("requireTSA", false, "Require RFC 3161 signed timestamp")
+	requireIntegratedTs = flag.Bool("requireIntegratedTs", false, "Require log entry integrated timestamp")
+	requireTimestamp = flag.Bool("requireTimestamp", false, "Require either an RFC3161 signed timestamp or log entry integrated timestamp")
```
steiza: Warning, I'm about to get nit-picky. : )

So we definitely want the library to cover as many reasonable use-cases as possible. But I don't think our examples (like `oci-image-verification` or `sigstore-go`) need to cover all those cases - that's `cosign`'s job.

So what do you think about the following changes (for both this and the `sigstore-go` example binary):
- Change `requireTimestamp` to default to `true` (since `requireTlog` defaults to `true`; also I'd expect most bundles will have at least one timestamp?)
- Remove the `requireIntegratedTs` and `requireTSA` options from these binaries (but of course leave them in the library; we can always point people at `conformance` to see how these options are used if they're looking for an example)

Hayden-IO: Good with me! I didn't know how we wanted to maintain the examples going forward, given they looked like mini-cosigns but we didn't want to ship binaries for them. Made the suggested changes.

steiza (approved):
LGTM!

codysoyland (approved):

PR #76, opened 2024-01-16
Allow configurable algorithms in verification

tetsuo-cpp:
#### Summary

Support configurable signing algorithms in `sigstore-go`'s verification flow. We already have the signing key/certificate so we can use the artifact digest algorithm information to figure out what hash function is being used.


tetsuo-cpp:
CC: @ret2libc

tetsuo-cpp:
I think it's best to get this in first and follow up with the change to restrict algorithms via a `--allowed-signing-algorithms` flag since it's not straightforward with the current CLI flags library.

pkg/verify/signature.go
```diff
@@ -54,14 +53,21 @@ func VerifySignatureWithArtifact(sigContent SignatureContent, verificationConten
 	var verifier signature.Verifier
 	var err error
 
-	verifier, err = getSignatureVerifier(verificationContent, trustedMaterial)
-	if err != nil {
-		return fmt.Errorf("could not load signature verifier: %w", err)
-	}
-
 	if envelope := sigContent.EnvelopeContent(); envelope != nil {
+		verifier, err = getSignatureVerifier(verificationContent, trustedMaterial)
```
tetsuo-cpp, line 56: As discussed, the envelope flow just uses the defaults that we had before.

cmd/sigstore-go/main.go
ret2libc, line 1: I think all these changes are about the rekor server and it's assumed to always be ecdsa/sha256. Moreover, I'm not sure it depends on the artifact digest algorithm.

tetsuo-cpp: I might be misreading the code, but it looks to me that the `TrustedMaterial.PublicKeyVerifier` gets used to verify the artifact signature, not anything about the Rekor entry.

pkg/verify/signature.go
```diff
@@ -73,25 +79,35 @@ func VerifySignatureWithArtifactDigest(sigContent SignatureContent, verification
...
+	return getSignatureVerifierWithHash(content, tm, crypto.SHA256)
+}
+
+func getSignatureVerifierWithHash(verificationContent VerificationContent, tm root.TrustedMaterial, hash crypto.Hash) (signature.Verifier, error) {
 	if leafCert, ok := verificationContent.HasCertificate(); ok {
-		// TODO: Inspect certificate's SignatureAlgorithm to determine hash function
-		return signature.LoadVerifier(leafCert.PublicKey, crypto.SHA256)
+		return signature.LoadVerifier(leafCert.PublicKey, hash)
 	} else if pk, ok := verificationContent.HasPublicKey(); ok {
 		return tm.PublicKeyVerifier(pk.Hint())
```
tetsuo-cpp, line 109: @ret2libc This is the spot here.

cmd/sigstore-go/main.go
ret2libc: Yeah you are right! Sorry about that!

tetsuo-cpp: No worries!

go.mod
```diff
@@ -5,15 +5,15 @@ go 1.21
...
 	github.com/go-openapi/strfmt v0.22.0
-	github.com/go-openapi/swag v0.22.7
+	github.com/go-openapi/swag v0.22.9
 	github.com/google/certificate-transparency-go v1.1.7
 	github.com/in-toto/in-toto-golang v0.9.0
 	github.com/secure-systems-lab/go-securesystemslib v0.8.0
 	github.com/sigstore/protobuf-specs v0.2.1
-	github.com/sigstore/rekor v1.3.4
-	github.com/sigstore/sigstore v1.8.0
+	github.com/sigstore/rekor v1.3.5-0.20240129120212-63aa08fd13eb
```
tetsuo-cpp, line 15: - [ ] Pin to a version once a new Rekor release is cut.

pkg/bundle/verification_content.go
```diff
@@ -28,11 +28,11 @@ type CertificateChain struct {
 }
 
 type PublicKey struct {
-	hint string
+	HintString string
```
tetsuo-cpp, line 31: I had to make this public so I could create one in the test module.

pkg/verify/signature_test.go
```diff
@@ -113,3 +120,50 @@ func TestSignatureVerifierMessageSignature(t *testing.T) {
 	assert.Error(t, err)
 	assert.Nil(t, result)
 }
+
+type nonExpiringVerifier struct {
```
tetsuo-cpp, line 125: This struct and the `trustedPublicKeyMaterial` helper is just copied from `sigstore-go/main.go` which isn't great. Not 100% sure what to do here because we need this to test the relevant functionality, however it probably doesn't belong in the public API.

pkg/verify/signature.go
```diff
@@ -234,3 +242,29 @@ func verifyMessageSignatureWithArtifactDigest(verifier signature.Verifier, msg M
...
+		hashFunc = crypto.SHA384
+	case "sha512":
+		hashFunc = crypto.SHA512
+	default:
+		return crypto.Hash(0), fmt.Errorf("unrecognized digest algorithm: %s", digestAlgorithm)
+	}
+	return hashFunc, nil
+}
+
+func getHashForHashAlgorithmString(hashAlgorithm string) (crypto.Hash, error) {
```
tetsuo-cpp, line 261: This is the string representation of the `HashAlgorithm` enum which gets used in the case where we're providing a trusted root (rather than public key material). Currently, different hash algorithms only work in the case where we provide trusted public key material since this enum only has a `SHA2_256` value. But if we add more values to this enum, we should be able to "unlock" different hash algorithms.

pkg/verify/signature.go
```diff
@@ -234,3 +242,29 @@ func verifyMessageSignatureWithArtifactDigest(verifier signature.Verifier, msg M
 
 	return nil
 }
+
+func GetHashForDigestAlgorithm(digestAlgorithm string) (crypto.Hash, error) {
```
tetsuo-cpp, line 246: - [ ] Figure out what this value should look like for Ed25519.

tetsuo-cpp (commented):
Unfortunately, unit testing this was a bit of a pain. I've left a few comments about things we probably need to discuss.

pkg/verify/signature_test.go
```diff
@@ -113,3 +121,58 @@ func TestSignatureVerifierMessageSignature(t *testing.T) {
...
+
+	virtualSigstore.TrustedMaterial = trustedPublicKeyMaterial(rsaKey.Public(), crypto.SHA256)
+
+	entity, err = virtualSigstore.SignWithPrivateKey("foofighters@example.com", "issuer", []byte(artifact), rsaKey, crypto.SHA256)
+	assert.NoError(t, err)
+
+	result, err = verifier.Verify(entity, verify.NewPolicy(verify.WithArtifact(bytes.NewBufferString(artifact)), verify.WithoutIdentitiesUnsafe()))
+	assert.NoError(t, err)
+
+	assert.Equal(t, result.VerifiedTimestamps[0].Type, "Tlog")
```
tetsuo-cpp, line 177: We should add a test for Ed25519 here. We'll need https://github.com/sigstore/rekor/pull/1945 for that.

Hayden-IO:
Now implemented.

PR #82, opened 2024-01-25
Bundle verification should require specifying expected certificate issuer and SAN

steiza:
#### Summary

Since many OIDC issuers are shared platforms, you need to ensure the Fulcio certificate SAN identity matches the identity on that platform you are expecting.

#### Release Note

* Bundle verification now requires not just specifying expected certificate issuer, but also expected identity via the certificate SAN 

#### Documentation

N/A

### DIFF
--- a/pkg/verify/certificate_identity.go
+++ b/pkg/verify/certificate_identity.go
@@ -91,6 +91,10 @@ func (s SubjectAlternativeNameMatcher) Verify(actualCert certificate.Summary) bo
 }
 
 func NewCertificateIdentity(sanMatcher SubjectAlternativeNameMatcher, extensions certificate.Extensions) (CertificateIdentity, error) {
+	if sanMatcher.SubjectAlternativeName.Type == "" && sanMatcher.SubjectAlternativeName.Value == "" && sanMatcher.Regexp.String() == "" {
+		return CertificateIdentity{}, errors.New("when verifying a certificate identity, there must be subject alternative name criteria")
+	}
+
 	certID := CertificateIdentity{SubjectAlternativeName: sanMatcher, Extensions: extensions}
 
 	if certID.Issuer == "" {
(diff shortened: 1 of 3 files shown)


pkg/verify/certificate_identity.go
```diff
@@ -91,6 +91,10 @@ func (s SubjectAlternativeNameMatcher) Verify(actualCert certificate.Summary) bo
 }
 
 func NewCertificateIdentity(sanMatcher SubjectAlternativeNameMatcher, extensions certificate.Extensions) (CertificateIdentity, error) {
+	if sanMatcher.SubjectAlternativeName.Type == "" && sanMatcher.SubjectAlternativeName.Value == "" && sanMatcher.Regexp.String() == "" {
```
Hayden-IO: Should we exclude `Type` from this check, and just require `Value` or `Regexp`?

`Type` seems more like a detail to cover specific cases where various SANs might look similar, rather than an identity string.

pkg/verify/certificate_identity_test.go
```diff
@@ -90,11 +90,17 @@ func TestCertificateIdentityVerify(t *testing.T) {
 	assert.Nil(t, ci)
 }
 
-func TestThatCertIDsHaveToHaveAnIssuer(t *testing.T) {
+func TestThatCertIDsAreFullySpecified(t *testing.T) {
 	_, err := NewShortCertificateIdentity("", "", "", "")
 	assert.NotNil(t, err)
 
 	_, err = NewShortCertificateIdentity("foobar", "", "", "")
+	assert.NotNil(t, err)
```
kommendorkapten: Here I think we should use `assert.Error(t, err)` instead (and line 95 above).

kommendorkapten: And the line below of course 😄

kommendorkapten (approved):
🚀

Hayden-IO (approved):

PR #92, opened 2024-02-05
Update linter, codeql

Hayden-IO:
CodeQL was failing because this version did not support Go 1.21. Also bumped the linter and fixed lint failures.

### DIFF
--- a/.github/workflows/verify_license.yml
+++ b/.github/workflows/verify_license.yml
@@ -30,7 +30,7 @@ jobs:
           go-version-file: './go.mod'
           check-latest: true
       - name: Install addlicense
-        run: go install github.com/google/addlicense@latest
+        run: go install github.com/google/addlicense@v1.1.1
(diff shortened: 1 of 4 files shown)


.github/workflows/verify_license.yml
```diff
@@ -30,8 +30,8 @@ jobs:
           go-version-file: './go.mod'
           check-latest: true
       - name: Install addlicense
-        run: go install github.com/google/addlicense@v1.0.0
+        run: go install github.com/google/addlicense@latest
```
kommendorkapten: Any specific reason for binding this to `latest` and not a git OID?

Hayden-IO: There's rarely updates, and dependabot also won't pick up updates for this since it's part of the command. I pinned it to v1.1.1

kommendorkapten (approved):

PR #114, opened 2024-03-05
Improve method naming in TrustedMaterial interface

### LINKED ISSUE #71: TrustedMaterial method naming
**Description**

Before a major release, I'd like to change the names of the interface methods so that each is not called an Authority. Overuse of Authority is going to lead to confusion since transparency logs are never referred to as authorities in any literature. I'd like to make the following changes:

* `TSACertificateAuthorities()` -> `TimestampingAuthorities()` (or `TimestampAuthorities()`, but the former is the RFC3161 name)
* `TlogAuthorities()` -> `RekorLogs()` (Specifying Rekor because this refers only to the Rekor personality and wouldn't work with some generic representation of a log)
* `CTlogAuthorities` -> `CTLogs()`
* The `TlogAuthority` struct -> `TransparencyLog`

A last change I am debating is changing `FulcioCertificateAuthorities` to `CodesigningCertificateAuthorities`. I don't think it's the right thing to do, but wanted to note my thoughts. By using Fulcio, we're stating a requirement for a specific certificate profile. "Codesigning" would allow BYO PKI scenarios. However, there are many assumptions made in the codebase around the structure of the certificate - short lived certificates, certain extensions, identities, etc. I think the better approach is to keep `FulcioCertificateAuthorities` and later add a separate type for a generic code-signing CA which relaxes many of the constraints. We were thinking something along these lines for Cosign, having a dedicated CLI namespace for BYO PKI.

codysoyland:
This implements @haydentherapper's proposal for improved naming for the methods in the TrustedMaterial interface.

Fixes https://github.com/sigstore/sigstore-go/issues/71

### DIFF
--- a/pkg/root/trusted_root.go
+++ b/pkg/root/trusted_root.go
@@ -35,11 +35,11 @@ const TrustedRootMediaType01 = "application/vnd.dev.sigstore.trustedroot+json;ve
 
 type TrustedRoot struct {
 	BaseTrustedMaterial
-	trustedRoot           *prototrustroot.TrustedRoot
-	tlogAuthorities       map[string]*TlogAuthority
-	fulcioCertAuthorities []CertificateAuthority
-	ctLogAuthorities      map[string]*TlogAuthority
-	tsaCertAuthorities    []CertificateAuthority
+	trustedRoot             *prototrustroot.TrustedRoot
+	rekorLogs               map[string]*TransparencyLog
+	fulcioCertAuthorities   []CertificateAuthority
+	ctLogs                  map[string]*TransparencyLog
+	timestampingAuthorities []CertificateAuthority
 }
 
 type CertificateAuthority struct {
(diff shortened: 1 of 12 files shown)


pkg/root/trusted_root.go
```diff
@@ -94,21 +94,21 @@ func NewTrustedRootFromProtobuf(protobufTrustedRoot *prototrustroot.TrustedRoot)
...
 	if err != nil {
 		return nil, err
 	}
 
 	return trustedRoot, nil
 }
 
-func ParseTlogAuthorities(tlogs []*prototrustroot.TransparencyLogInstance) (tlogAuthorities map[string]*TlogAuthority, err error) {
-	tlogAuthorities = make(map[string]*TlogAuthority)
+func ParseRekorLogs(tlogs []*prototrustroot.TransparencyLogInstance) (rekorLogs map[string]*TransparencyLog, err error) {
```
Hayden-IO: Should this function be `ParseTransparencyLogs` instead? It would help readability for something like line 102, where you assign what looks like "rekor" to "ctLogs"

codysoyland: Yeah, good point, I'll change that!

Hayden-IO (approved):
Looks good!

PR #144, opened 2024-04-08
Update embedded root to latest

Hayden-IO:
This bumps the embedded root to 9.root.json to avoid the extra network call for clients that have not cached the latest root yet.

All of the key changes are because of the change in keytype.


kommendorkapten (approved):
Thanks

Issue #64, opened 2024-01-02
Add unit tests for bundle package


Hayden-IO:
**Description**

Code: https://github.com/sigstore/sigstore-go/tree/f7518a68b37b7887b69e1607b9ddb668ddec117d/pkg/bundle

PR #146, opened 2024-04-10
Require inclusion proof for all bundle versions newer than v0.1

### LINKED ISSUE #145: Bundle v0.3 doesn't require tlog entries to have inclusion proofs
**Description**

The function that [validates the bundle](https://github.com/sigstore/sigstore-go/blob/main/pkg/bundle/bundle.go#L76-L102) bypasses the check for inclusion proofs if the bundle version is v0.3.

**Version**

codysoyland:

#### Summary

This PR modifies the logic in `ProtobufBundle.validate` to require inclusion proof for all bundles newer than v0.1.

This includes new tested logic to extract the bundle version instead of continuing to add new consts for each bundle version.

The validate function is still untested. This should be addressed with #64.

Fixes #145

### DIFF
--- a/pkg/bundle/bundle.go
+++ b/pkg/bundle/bundle.go
@@ -33,14 +33,8 @@ import (
 	"github.com/sigstore/sigstore-go/pkg/verify"
 )
 
-const SigstoreBundleMediaType01 = "application/vnd.dev.sigstore.bundle+json;version=0.1"
-const SigstoreBundleMediaType02 = "application/vnd.dev.sigstore.bundle+json;version=0.2"
-const SigstoreBundleMediaType03Legacy = "application/vnd.dev.sigstore.bundle+json;version=0.3"
-const SigstoreBundleMediaType03 = "application/vnd.dev.sigstore.bundle.v0.3+json"
-const IntotoMediaType = "application/vnd.in-toto+json"
-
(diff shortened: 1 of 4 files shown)


pkg/bundle/bundle.go
```diff
@@ -74,33 +68,71 @@ func NewProtobufBundle(pbundle *protobundle.Bundle) (*ProtobufBundle, error) {
...
 
 		if certs != nil {
 			return errors.New("verification material cannot be X.509 certificate chain (for bundle v0.3)")
 		}
-	default:
-		return ErrIncorrectMediaType
+	}
+
+	// if bundle version is newer than v0.3, return error as this version is not supported
+	if semver.Compare(bundleVersion, "v0.3") > 0 {
```
Hayden-IO, line 109: Do we need to handle v0.3.x? I saw we include a test for `v0.3.1` but I don't believe we'll include the patch version. I think this is fine as-is, but if we do want to handle patch versions, does this need to be >= v0.4?

Hayden-IO (approved):

pkg/bundle/bundle.go
codysoyland: I turned on auto-merge so this didn't get addressed, but good point! I can file another quick PR.

Issue #136, opened 2024-04-05
Support Signing


steiza:
### Description

It would be great if sigstore-go could not just verify, but also sign bundles.

There aren't many libraries that support signing bundles today (just sigstore-js?) This would also allow sigstore-go to support the full range of tests in sigstore-conformance.

### Goals

- `sigstore-go/pkg/sign/` supports signing Sigstore bundles
- `cmd/conformance/main.go` is updated to support `sign-bundle` and `sign` (similar to `verify`, `sign` will "wrap" the bundle flow).
- Have a "signers API"
  - Have built-in support for signing with Fulcio and a keypair where you have direct access to the private key
    - I'm leaning towards a "batteries included" Fulcio support in the library itself, but if we decide against it we'll at least need support in the conformance driver
- Have a "witness API"
  - Built-in support for a timestamp authority
  - Built-in support for Rekor: entry types hashedrekord and DSSE
- Produces bundles with v0.3.1 (open to feedback here)
- Signing requires exactly 1 signer and at least 1 witness

### Anti-Goals

- If we update `cmd/sigstore-go/main.go`, it should not compete with cosign
- API will support providing custom signers, but we won't support any KMS / HSM-based signing in `sigstore-go` itself

### References

- [Sigstore client specification](https://docs.google.com/document/d/1kbhK2qyPPk8SLavHzYSDM8-Ueul9_oxIMVFuWMWKz0E/edit)
- https://github.com/sigstore/sigstore-js/blob/main/packages/sign/README.md
  - and the discussion on https://github.com/sigstore/sigstore-js/issues/499

steiza:
So here is a very rough example of what `pkg/sign/` might look like:

For `interface.go`:
```go
package sign                                                                    
                                                                                
import (                                                                        
    protobundle "github.com/sigstore/protobuf-specs/gen/pb-go/bundle/v1"        
    protocommon "github.com/sigstore/protobuf-specs/gen/pb-go/common/v1"        
)                                                                               
                                                                                
// this is just a convienient way to bundle the information from the signer to the witness(es)
type SignedEntity struct{                                                       
    dataDigest []byte                                                           
    signature []byte                                                            
    ephemeralPublicKey []byte                                                   
    envelope []byte                                                             
}                                                                               
                                                                                
func Sign(data []byte, signer *Signer, witnesses []Witness, bundleVersion string) (*protobundle.Bundle, error) {
    bundle := Bundle{}
    
[...]

PR #148, opened 2024-04-12
Add partial example of signing

steiza:
#### Summary

For #136.

This isn't a full implementation of signing - it supports an on-disk keypair and a timestamp authority. It doesn't yet support Fulcio or Rekor. But I wanted to post a work-in-progress to get feedback on signing API generally.

We could choose to land this functionality and have a follow-on PR that implements Fulcio and Rekor support (and enables conformance testing on signing) - smaller reviews are nice! Or I could continue to update this PR with more functionality.

#### Release Note

NONE - for now, until at least Fulcio is supported

#### Documentation

NONE

### DIFF
--- a/pkg/sign/signer.go
+++ b/pkg/sign/signer.go
@@ -18,7 +18,7 @@ import (
 	"crypto"
 	"crypto/rand"
 	"crypto/sha256"
-	_ "crypto/sha512" // if user chooses SHA2-384 or SHA2-256 for hash
+	_ "crypto/sha512" // if user chooses SHA2-384 or SHA2-512 for hash
 	"crypto/x509"
 	"encoding/base64"
 	"errors"
--- a/pkg/sign/signer.go
+++ b/pkg/sign/signer.go
@@ -1,0 +1,95 @@
+// Copyright 2024 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sign
+
+import (
+	"crypto"
+	"crypto/rand"
+	_ "crypto/sha256" // hash function will be used by Sign
+	_ "crypto/sha512" // hash function will be used by Sign
+	"errors"
+
+	protobundle "github.com/sigstore/protobuf-specs/gen/pb-go/bundle/v1"
+	protocommon "github.com/sigstore/protobuf-specs/gen/pb-go/common/v1"
+)
+
+type Signer interface {
+	Sign(data []byte) (*protobundle.Bundle, error)
+}
+
+// TODO: implement
+// type Fulcio struct {
+//	baseURL       string
+//	identityToken string
+// }
+
+// func (f *Fulcio) Sign(data []byte) (*protobundle.Bundle, error) {}
+
+type Keypair struct {
+	signer        crypto.Signer
+	hashAlgorithm protocommon.HashAlgorithm
+	publicKeyHint []byte
+}
+
+func SignerKeypair(s crypto.Signer, ha protocommon.HashAlgorithm, pkh []byte) Keypair {
+	return Keypair{signer: s, hashAlgorithm: ha, publicKeyHint: pkh}
+}
+
+func (k Keypair) Sign(data []byte) (*protobundle.Bundle, error) {
+	bundle := protobundle.Bundle{
+		MediaType: "application/vnd.dev.sigstore.bundle.v0.3+json",
+		VerificationMaterial: &protobundle.VerificationMaterial{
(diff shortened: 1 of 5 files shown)


pkg/sign/witness.go
Hayden-IO, line 1: We need a different name for this concept, this will conflict with transparency log witnessing.

Actually, per the comment below, I would just call this timestamping.

pkg/sign/witness.go
```diff
@@ -0,0 +1,113 @@
...
+		vm.TimestampVerificationData = tsVerificationData
+	}
+
+	tsVerificationData.Rfc3161Timestamps = append(tsVerificationData.Rfc3161Timestamps, signedTimestamp)
+
+	return nil
+}
+
+// TODO: implement
+// type Rekor struct {
```
Hayden-IO: I mentioned this just now on the issue, but I don't think we should conflate timestamping and logging. These are two different concepts. It just so happens that the log can provide a timestamp, and it can be trusted by the client policy if they want to trust the log.

I would rather have two separate steps, one for timestamping (which could fetch a timestamp from a TSA or a Roughtime server or something else in the future), and one for logging (calling out to N logs). It's up to the client to determine if it trusts the timestamp returned from the last step.

pkg/sign/witness.go
```diff
@@ -0,0 +1,113 @@
...
+		req.HashAlgorithm = crypto.SHA256
+		req.HashedMessage = signatureHash[:]
+	}
+
+	reqBytes, err := req.Marshal()
+	if err != nil {
+		return err
+	}
+
+	client, err := tsaclient.GetTimestampClient(ta.baseURL, tsaclient.WithUserAgent("sigstore-go"), tsaclient.WithContentType(tsaclient.TimestampQueryMediaType))
```
Hayden-IO: Can we include version in the user agent (if that's possible to dynamically fetch?)?

pkg/sign/witness.go
```diff
@@ -0,0 +1,113 @@
...
+	tsaclient "github.com/sigstore/timestamp-authority/pkg/client"
+	tsagenclient "github.com/sigstore/timestamp-authority/pkg/generated/client/timestamp"
+)
+
+type Witness interface {
+	Witness(*protobundle.Bundle) error
+}
+
+type TimestampAuthority struct {
+	baseURL string
```
Hayden-IO: We should include a context or common parameters like timeout to initialize the client.

pkg/sign/signer.go
```diff
@@ -0,0 +1,95 @@
...
+
+// func (f *Fulcio) Sign(data []byte) (*protobundle.Bundle, error) {}
+
+type Keypair struct {
+	signer        crypto.Signer
+	hashAlgorithm protocommon.HashAlgorithm
+	publicKeyHint []byte
+}
+
+func SignerKeypair(s crypto.Signer, ha protocommon.HashAlgorithm, pkh []byte) Keypair {
```
Hayden-IO: Do we want to make an options struct to initialize the signer? Then if we need to expand the options in the future, it won't require a breaking API change.

pkg/sign/signer.go
```diff
@@ -0,0 +1,95 @@
...
+	return Keypair{signer: s, hashAlgorithm: ha, publicKeyHint: pkh}
```
Hayden-IO: For `publicKeyHint`, can we make this an optional parameter, and if it's not given, derive it ourselves (`SHA-256(DER(SPKI(pub key)))`)? Most clients won't be opinionated on the key hint I suspect.

Hayden-IO (changes requested):
+100 to small PRs! This is much easier to review! Let's go with incremental PRs, and I suspect most of the discussion will be in the first couple PRs as we figure out the interface.

pkg/sign/witness.go
steiza: 👍 When we get to it, we'll put Rekor logging in a separate file in the `sign` directory.

pkg/sign/signer.go
steiza: I'm not sure I produced the hint in exactly the way you suggested, but we do now generate a hint string if one is not provided.

pkg/sign/witness.go
steiza: For now I added this as an optional parameter to `TimestampAuthorityOptions`, populated by `git describe --tags` in the `Makefile`. @codysoyland what do you think about how/if we should track the version of `sigstore-go` in the codebase?

pkg/sign/signer.go
```diff
@@ -0,0 +1,113 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sign
+
+import (
+	"crypto"
+	"crypto/rand"
+	"crypto/sha256"
+	_ "crypto/sha512" // if user chooses SHA2-384 or SHA2-256 for hash
```
Hayden-IO:
```suggestion
	_ "crypto/sha512" // if user chooses SHA2-384 or SHA2-512 for hash
```

pkg/sign/signer.go
Hayden-IO: Looks great! That's equivalent, DER(SPKI()) is the same as a marshaled PKIX key.

Hayden-IO (dismissed):
Looks great!

Hayden-IO (approved):

Sec32fun32 (commented):
Sizlerin incelemelerde bulunmasını talep ediyorum

PR #154, opened 2024-04-23
Add mediatype string builder to bundle package

### LINKED ISSUE #153: Avoid repeating media type strings
**Description**

With the release of v0.3 all apps relying on the media type constants in `pkg/bundle` broke as the constants were removed in https://github.com/sigstore/sigstore-go/commit/20c2ce9c55fc2d05bf01604c491edfe46e43e4dc

We should return the (existing) constants and/or add a method to create the bundle media type string to avoid duplicating the strings in all code using the library.

**Version**

v0.3

puerco:
#### Summary

This commit adds a media type string build function to handle the media type string in the module.

With the release of 0.3 all apps relying on the media type constants in `pkg/bundle` broke as the constants were removed in https://github.com/sigstore/sigstore-go/commit/20c2ce9c55fc2d05bf01604c491edfe46e43e4dc. This PR introduces a function to generate the media type string from the module and avoid spreading media type strings on applications using the module.

Fixes: https://github.com/sigstore/sigstore-go/issues/153

/cc @codysoyland 

#### Release Note

Added a `bundle.MediaTypeString()` function to handle the media type strings from the bundle module

### DIFF
--- a/pkg/bundle/bundle.go
+++ b/pkg/bundle/bundle.go
@@ -42,6 +42,8 @@ var ErrMissingEnvelope = fmt.Errorf("%w: missing envelope", ErrInvalidAttestatio
 var ErrDecodingJSON = fmt.Errorf("%w: decoding json", ErrInvalidAttestation)
 var ErrDecodingB64 = fmt.Errorf("%w: decoding base64", ErrInvalidAttestation)
 
+const mediaTypeBase = "application/vnd.dev.sigstore.bundle"
+
 func ErrValidationError(err error) error {
 	return fmt.Errorf("%w: %w", ErrValidation, err)
 }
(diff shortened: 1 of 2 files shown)


pkg/bundle/bundle.go
```diff
@@ -113,17 +115,32 @@ func (b *ProtobufBundle) validate() error {
...
 }
 
+// MediaTypeString returns a mediatype string for the specified bundle version.
+// The function returns an error if the resulting string does validate.
+func MediaTypeString(version string) (string, error) {
+	if version == "" {
+		return "", fmt.Errorf("unable to build media type string, no version defined")
+	}
+
+	mtString := fmt.Sprintf("%s+json;version=%s", mediaTypeBase, strings.TrimPrefix(version, "v"))
```
Hayden-IO: This is dependent on version now, since we made a change to be compliant with OCI as per https://github.com/sigstore/protobuf-specs/issues/165.

v0.1 and v0.2 should be `application/vnd.dev.sigstore.bundle+json;version=0.1` and `application/vnd.dev.sigstore.bundle+json;version=0.2` respectively, and for v0.3 onwards, `application/vnd.dev.sigstore.bundle.v0.%s+json` where `%s` is the trimmed version.

Hayden-IO (changes requested):

pkg/bundle/bundle.go
puerco: Ah thanks for the pointer @haydentherapper I was wondering if they were interchangeable. OK, now the function will return the old style for 0.1 - 0.2 and the new one for the rest. I've added cases to test them.

Hayden-IO (approved):
Thanks!

codysoyland (approved):
Nice!

PR #158, opened 2024-05-01
Add Fulcio support to signing

steiza:
#### Summary

Continuing to work towards #136 with small / incremental diffs.

This involved a bit of a refactor to support MessageSignature and DSSE bundle content.

Also, we should probably start adding signing tests soon.

#### Release Note

NONE - still no Rekor support or testing via conformance tests or integration tests with mocks

#### Documentation

NONE

### DIFF
--- a/examples/signing/main.go
+++ b/examples/signing/main.go
@@ -64,38 +64,26 @@ func main() {
 		}
 	}
 
-	var signer sign.Signer
-	var keypair sign.Keypair
+	keypair, err := sign.NewEphemeralKeypair(nil)
+	if err != nil {
+		log.Fatal(err)
+	}
+	publicKeyPem, err := keypair.GetPublicKeyPem()
+	if err != nil {
+		log.Fatal(err)
+	}
+	fmt.Printf("Using public key:\n\n%s\n\n", publicKeyPem)
+
+	var fulcio *sign.Fulcio
 
 	if *idToken != "" {
 		fulcioOpts := &sign.FulcioOptions{
 			BaseURL:        "https://fulcio.sigstage.dev",
-			IdentityToken:  *idToken,
 			Timeout:        time.Duration(30 * time.Second),
 			LibraryVersion: Version,
 		}
 
-		signer = sign.NewFulcio(fulcioOpts)
-	} else {
-		// Create a keypair to sign with
-		var err error
-		keypair, err = sign.NewEphemeralKeypair(nil)
-		if err != nil {
-			log.Fatal(err)
-		}
-
-		publicKeyPem, err := keypair.GetPublicKeyPem()
-		if err != nil {
-			log.Fatal(err)
-		}
-		fmt.Printf("Using public key:\n\n%s\n\n", publicKeyPem)
-
-		signer = &sign.KeySigner{}
-	}
-
-	bundle, err := signer.Sign(content, keypair)
-	if err != nil {
-		log.Fatal(err)
+		fulcio = sign.NewFulcio(fulcioOpts)
 	}
 
 	tsaOpts := &sign.TimestampAuthorityOptions{
@@ -105,7 +93,8 @@ func main() {
 	}
 
 	tsa := sign.NewTimestampAuthority(tsaOpts)
-	err = tsa.GetTimestamp(bundle)
+
+	bundle, err := sign.Bundle(content, keypair, fulcio, *idToken, &tsa)
 	if err != nil {
 		log.Fatal(err)
 	}
--- a/examples/signing/main.go
+++ b/examples/signing/main.go
@@ -20,6 +20,7 @@ import (
 	"crypto/rand"
 	"crypto/x509"
 	"encoding/pem"
+	"flag"
 	"fmt"
 	"log"
 	"time"
@@ -33,34 +34,60 @@ import (
 var Version string
 
 func main() {
-	privateKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
-	if err != nil {
-		log.Fatal(err)
+	flag.Parse()
+
+	var signer sign.Signer
+
+	if flag.NArg() == 0 {
+		// Assume we're signing with a keypair
+		privateKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		// Print out privateKey.PublicKey() in PEM format
+		pubKeyBytes, err := x509.MarshalPKIXPublicKey(&privateKey.PublicKey)
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		pemBlock := pem.Block{
+			Type:  "PUBLIC KEY",
+			Bytes: pubKeyBytes,
+		}
+		fmt.Println(string(pem.EncodeToMemory(&pemBlock)))
+
+		keypairOpts := &sign.KeypairOptions{
+			Signer:        privateKey,
+			HashAlgorithm: protocommon.HashAlgorithm_SHA2_256,
+			PublicKeyHint: []byte("someKeyHint"),
+		}
+		signer, err = sign.SignerKeypair(keypairOpts)
+		if err != nil {
+			log.Fatal(err)
+		}
+	} else {
+		// Assume arg is identity token for Fulcio
+		fulcioOpts := &sign.FulcioOptions{
+			BaseURL:        "https://fulcio.sigstage.dev",
+			IdentityToken:  flag.Arg(0),
+			Timeout:        time.Duration(30 * time.Second),
+			LibraryVersion: Version,
+		}
+
+		signer = sign.SignerFulcio(fulcioOpts)
 	}
 
-	// Print out privateKey.PublicKey() in PEM format
-	pubKeyBytes, err := x509.MarshalPKIXPublicKey(&privateKey.PublicKey)
-	if err != nil {
-		log.Fatal(err)
+	/*data := sign.PlainData{
+		Data: []byte("hello world"),
+	}*/
+
+	data := sign.DSSEData{
+		Data:        []byte(`{"_type":"https://in-toto.io/Statement/v0.1","subject":[{"name":"hello_world.txt","digest":{"sha256":"b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9"}}],"predicateType":"something","preidcate":{}}`),
+		PayloadType: "application/vnd.in-toto+json",
 	}
 
-	pemBlock := pem.Block{
-		Type:  "PUBLIC KEY",
-		Bytes: pubKeyBytes,
-	}
-	fmt.Println(string(pem.EncodeToMemory(&pemBlock)))
-
-	keypairOpts := &sign.KeypairOptions{
-		Signer:        privateKey,
-		HashAlgorithm: protocommon.HashAlgorithm_SHA2_256,
-		PublicKeyHint: []byte("someKeyHint"),
-	}
-	signer, err := sign.SignerKeypair(keypairOpts)
-	if err != nil {
-		log.Fatal(err)
-	}
-
-	bundle, err := signer.Sign([]byte("hello world"))
+	bundle, err := signer.Sign(data)
 	if err != nil {
 		log.Fatal(err)
 	}
--- a/examples/signing/main.go
+++ b/examples/signing/main.go
@@ -54,12 +54,12 @@ func main() {
 	}
 
 	if *intoto {
-		content = sign.DSSEData{
+		content = &sign.DSSEData{
 			Data:        data,
 			PayloadType: "application/vnd.in-toto+json",
 		}
 	} else {
-		content = sign.PlainData{
+		content = &sign.PlainData{
 			Data: data,
 		}
 	}
--- a/examples/signing/main.go
+++ b/examples/signing/main.go
@@ -15,79 +15,85 @@
 package main
 
 import (
-	"crypto/ecdsa"
-	"crypto/elliptic"
-	"crypto/rand"
-	"crypto/x509"
-	"encoding/pem"
 	"flag"
 	"fmt"
 	"log"
+	"os"
 	"time"
 
-	protocommon "github.com/sigstore/protobuf-specs/gen/pb-go/common/v1"
 	"google.golang.org/protobuf/encoding/protojson"
 
 	"github.com/sigstore/sigstore-go/pkg/sign"
 )
 
 var Version string
+var idToken *string
+var intoto *bool
+
+func init() {
+	idToken = flag.String("id-token", "", "OIDC token to send to Fulcio")
+	intoto = flag.Bool("in-toto", false, "Content to sign is in-toto document")
+	flag.Parse()
+	if flag.NArg() == 0 {
+		usage()
+		os.Exit(1)
+	}
+}
+
+func usage() {
+	fmt.Printf("Usage: %s [OPTIONS] FILE_TO_SIGN\n", os.Args[0])
+	flag.PrintDefaults()
+}
 
 func main() {
-	flag.Parse()
+	var content sign.Content
 
-	var signer sign.Signer
+	data, err := os.ReadFile(flag.Arg(0))
+	if err != nil {
+		log.Fatal(err)
+	}
 
-	if flag.NArg() == 0 {
-		// Assume we're signing with a keypair
-		privateKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
-		if err != nil {
-			log.Fatal(err)
-		}
-
-		// Print out privateKey.PublicKey() in PEM format
-		pubKeyBytes, err := x509.MarshalPKIXPublicKey(&privateKey.PublicKey)
-		if err != nil {
-			log.Fatal(err)
-		}
-
-		pemBlock := pem.Block{
-			Type:  "PUBLIC KEY",
-			Bytes: pubKeyBytes,
-		}
-		fmt.Println(string(pem.EncodeToMemory(&pemBlock)))
-
-		keypairOpts := &sign.KeypairOptions{
-			Signer:        privateKey,
-			HashAlgorithm: protocommon.HashAlgorithm_SHA2_256,
-			PublicKeyHint: []byte("someKeyHint"),
-		}
-		signer, err = sign.SignerKeypair(keypairOpts)
-		if err != nil {
-			log.Fatal(err)
+	if *intoto {
+		content = sign.DSSEData{
+			Data:        data,
+			PayloadType: "application/vnd.in-toto+json",
 		}
 	} else {
-		// Assume arg is identity token for Fulcio
+		content = sign.PlainData{
+			Data: data,
+		}
+	}
+
+	var signer sign.Signer
+	var keypair sign.Keypair
+
+	if *idToken != "" {
 		fulcioOpts := &sign.FulcioOptions{
 			BaseURL:        "https://fulcio.sigstage.dev",
-			IdentityToken:  flag.Arg(0),
+			IdentityToken:  *idToken,
 			Timeout:        time.Duration(30 * time.Second),
 			LibraryVersion: Version,
 		}
 
-		signer = sign.SignerFulcio(fulcioOpts)
+		signer = sign.NewFulcio(fulcioOpts)
+	} else {
+		// Create a keypair to sign with
+		var err error
+		keypair, err = sign.NewEphemeralKeypair(nil)
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		publicKeyPem, err := keypair.GetPublicKeyPem()
+		if err != nil {
+			log.Fatal(err)
+		}
+		fmt.Printf("Using public key:\n\n%s\n\n", publicKeyPem)
+
+		signer = &sign.KeySigner{}
 	}
 
-	/*data := sign.PlainData{
-		Data: []byte("hello world"),
-	}*/
-
-	data := sign.DSSEData{
-		Data:        []byte(`{"_type":"https://in-toto.io/Statement/v0.1","subject":[{"name":"hello_world.txt","digest":{"sha256":"b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9"}}],"predicateType":"something","preidcate":{}}`),
-		PayloadType: "application/vnd.in-toto+json",
-	}
-
-	bundle, err := signer.Sign(data)
+	bundle, err := signer.Sign(content, keypair)
 	if err != nil {
 		log.Fatal(err)
 	}
--- a/pkg/sign/certificate.go
+++ b/pkg/sign/certificate.go
@@ -1,0 +1,175 @@
+// Copyright 2024 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sign
+
+import (
+	"bytes"
+	"encoding/base64"
+	"encoding/json"
+	"encoding/pem"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+	"time"
+)
+
+type Fulcio struct {
+	options *FulcioOptions
+}
+
+type FulcioOptions struct {
+	// URL of Fulcio instance
+	BaseURL string
+	// Optional timeout for network requests
+	Timeout time.Duration
+	// Optional version string for user agent
+	LibraryVersion string
+}
+
+type jsonWebToken struct {
+	Sub string `json:"sub"`
+}
+
+type fulcioCertRequest struct {
+	PublicKeyRequest publicKeyRequest `json:"publicKeyRequest"`
+}
+
+type publicKeyRequest struct {
+	PublicKey         publicKey `json:"publicKey"`
+	ProofOfPossession string    `json:"proofOfPossession"`
+}
+
+type publicKey struct {
+	Algorithm string `json:"algorithm"`
+	Content   string `json:"content"`
+}
+
+type fulcioResponse struct {
+	SctCertWithChain signedCertificateEmbeddedSct `json:"signedCertificateEmbeddedSct"`
+}
+
+type signedCertificateEmbeddedSct struct {
+	Chain chain `json:"chain"`
+}
+
+type chain struct {
+	Certificates []string `json:"certificates"`
+}
+
+func NewFulcio(opts *FulcioOptions) *Fulcio {
+	return &Fulcio{options: opts}
+}
+
+func (f *Fulcio) GetCertificate(keypair Keypair, identityToken string) ([]byte, error) {
+	// Get JWT from identity token
+	//
+	// Note that the contents of this token are untrusted. Fulcio will perform
+	// the token verification.
+	tokenParts := strings.Split(identityToken, ".")
+	if len(tokenParts) < 2 {
+		return nil, errors.New("Unable to get subject from identity token")
+	}
+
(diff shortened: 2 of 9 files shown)


pkg/sign/content.go
```diff
@@ -0,0 +1,74 @@
...
+	"fmt"
+
+	protobundle "github.com/sigstore/protobuf-specs/gen/pb-go/bundle/v1"
+	protocommon "github.com/sigstore/protobuf-specs/gen/pb-go/common/v1"
+	protodsse "github.com/sigstore/protobuf-specs/gen/pb-go/dsse"
+)
+
+type Content interface {
+	// Return the data to be signed
+	Prepare() []byte
```
kommendorkapten: I think we should rename this to something... different... 😄 
As this preparation step is extremely critical to achieve good security, I think this should be more clear in the name. In the DSSE example below, DSSE PAE (Pre Auth Encoding) is used. And I'm thinking if the method name on the interface should be similar? `PAE`, or `Canonicalize`? 
My vote would be PAE, as it's a commonly used technique to prepare data before computing a digest or signing.

edit: spelling errors.

steiza: I updated this to be `PreAuthEncoding`

steiza:
This PR is kind of large, and it's likely it'll grow with some more testing. If it would help reviewing, I could separate out the Fulcio addition and the bundle content supporting Message Signature and DSSE Envelope - let me know.

examples/signing/main.go
```diff
@@ -33,34 +34,60 @@ import (
...
+		}
+		signer, err = sign.SignerKeypair(keypairOpts)
+		if err != nil {
+			log.Fatal(err)
+		}
+	} else {
+		// Assume arg is identity token for Fulcio
+		fulcioOpts := &sign.FulcioOptions{
+			BaseURL:        "https://fulcio.sigstage.dev",
+			IdentityToken:  flag.Arg(0),
```
Hayden-IO: Can we add flag variables following https://pkg.go.dev/flag, e.g `var idToken = flag.String("id-token", "", "help message")` rather than assume argument positions? I know we don't want to evolve this into a fully developed CLI, but I think it'll make it simpler to add more examples as we flesh out signing.

examples/signing/main.go
```diff
@@ -33,34 +34,60 @@ import (
...
+		signer = sign.SignerFulcio(fulcioOpts)
 	}
-	signer, err := sign.SignerKeypair(keypairOpts)
-	if err != nil {
-		log.Fatal(err)
+
+	/*data := sign.PlainData{
+		Data: []byte("hello world"),
+	}*/
+
```
Hayden-IO: Can we add a flag for signing a blob vs attestation?

pkg/sign/content.go
```diff
@@ -0,0 +1,74 @@
...
+
+	protobundle "github.com/sigstore/protobuf-specs/gen/pb-go/bundle/v1"
+	protocommon "github.com/sigstore/protobuf-specs/gen/pb-go/common/v1"
+	protodsse "github.com/sigstore/protobuf-specs/gen/pb-go/dsse"
+)
+
+type Content interface {
+	// Return the data to be signed
+	PreAuthEncoding() []byte
+	// Returns something that satisfies protobundle.isBundle_Content
```
Hayden-IO: nit: Should say "mutates bundle" or something similar rather than return

pkg/sign/content.go
```diff
@@ -0,0 +1,74 @@
...
+	Bundle(bundle *protobundle.Bundle, hashAlgorithm protocommon.HashAlgorithm, digest []byte, signature []byte)
```
Hayden-IO: Thoughts on `hashAlgorithm` and `digest` being fields on `PlainData` rather than variables to `Bundle()`? I think that would be a cleaner interface and would avoid ignoring those variables in DSSEData's impl of `Bundle`.

pkg/sign/timestamping.go
```diff
@@ -50,25 +51,27 @@ func (ta *TimestampAuthority) GetTimestamp(b *protobundle.Bundle) error {
...
+	dsseEnvelope := b.GetDsseEnvelope()
 
-	if messageSignature != nil {
+	switch {
+	case messageSignature != nil:
 		signatureHash := sha256.Sum256(messageSignature.Signature)
 		req.HashAlgorithm = crypto.SHA256
 		req.HashedMessage = signatureHash[:]
+	case dsseEnvelope != nil:
+		signatureHash := sha256.Sum256(dsseEnvelope.Signatures[0].Sig)
```
Hayden-IO: Hm, this is an interesting limitation of using timestamping with DSSE, since a timestamp is a counter-signature. This doesn't come up when using SETs since the timestamp is over the entire entry, which includes multiple signatures.

Should this hash be over the concatenation of all signatures instead? We just need to set the precedent and document how to verify multi-sig DSSEs, and check how other clients handle this.

pkg/sign/signer.go
```diff
@@ -15,29 +15,218 @@
...
+		return nil, err
+	}
+	requestBytes := bytes.NewBuffer(requestJSON)
+
+	var client http.Client
+	if f.options.Timeout != 0 {
+		client.Timeout = f.options.Timeout
+	}
+
+	request, err := http.NewRequest("POST", f.options.BaseURL+"/api/v2/signingCert", requestBytes)
```
Hayden-IO: Not blocking, but as a follow up, I'd like to use the client from Fulcio's repo. It doesn't use the v2 API currently though. I've got a WIP client that does, we can change this later on. Can you add a TODO?

pkg/sign/signer.go
```diff
@@ -15,29 +15,218 @@
...
+	if len(tokenParts) < 2 {
+		return nil, errors.New("Unable to get subject from identity token")
+	}
+
+	jwtString, err := base64.RawURLEncoding.DecodeString(tokenParts[1])
+	if err != nil {
+		return nil, err
+	}
+
+	var jwt jsonWebToken
```
Hayden-IO: Can we add a comment about not trusting this token? Something along the lines of: It isn't the library's responsibility to verify the signature on the token, we only extract the subject. The CA will verify it.

pkg/sign/signer.go
```diff
@@ -15,29 +15,218 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Generate ephemeral keypair
+	privateKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	if err != nil {
+		return nil, err
+	}
+
```
Hayden-IO: Can we use the sigstore/sigstore helper `MarshalPublicKeyToPEM`? https://github.com/sigstore/sigstore/blob/main/pkg/cryptoutils/publickey.go#L77. Shouldn't change the dependency tree and just cuts down on some repetition.

pkg/sign/signer.go
```diff
@@ -15,29 +15,218 @@
...
+	// Sign JWT subject with ephemeral key for proof of possession
+	subjectDigest := sha256.Sum256([]byte(jwt.Sub))
+	subjectSignature, err := privateKey.Sign(rand.Reader, subjectDigest[:], nil)
+	if err != nil {
+		return nil, err
+	}
+
+	// Make Fulcio certificate request
+	certRequest := fulcioCertRequest{
+		Credentials: identityToken{
```
Hayden-IO: This shouldn't be needed if you set the Authorization header (which you are doing below), or vice versa

pkg/sign/signer.go
```diff
@@ -15,29 +15,218 @@
...
+		return nil, err
+	}
+
+	if response.StatusCode != 200 {
+		return nil, fmt.Errorf("Fulcio returned %d: %s", response.StatusCode, string(body))
+	}
+
+	// Assemble bundle from Fulcio response
+	var fulcioResp fulcioResponse
+	err = json.Unmarshal(body, &fulcioResp)
```
Hayden-IO: Does this work? I thought oneOf and json unmarshalling don't work well together.

In the WIP client, I was using protojson - https://gist.github.com/haydentherapper/86274b9211685a34b4da6b16725870a9#file-fulcio-client-go-L186-L189

pkg/sign/signer.go
```diff
@@ -15,29 +15,218 @@
...
+	ProofOfPossession string    `json:"proofOfPossession"`
+}
+
+type publicKey struct {
+	Algorithm string `json:"algorithm"`
+	Content   string `json:"content"`
+}
+
+type fulcioResponse struct {
+	SCT signedCertificateEmbeddedSct `json:"signedCertificateEmbeddedSct"`
```
Hayden-IO: I would call this `CertificateWithSCTAndChain` or something similar (and less verbose than what I proposed), since it's not only an SCT.

pkg/sign/signer.go
```diff
@@ -15,29 +15,218 @@
...
+
+	data := content.PreAuthEncoding()
+	dataDigest := sha256.Sum256([]byte(data))
+	signature, err := ecdsa.SignASN1(rand.Reader, privateKey, dataDigest[:])
+	if err != nil {
+		return nil, err
+	}
+
+	bundle := &protobundle.Bundle{
+		MediaType: "application/vnd.dev.sigstore.bundle.v0.3+json",
```
Hayden-IO: Do we have a constant we can use for this string?

pkg/sign/signer.go
```diff
@@ -15,29 +15,218 @@
...
+	BaseURL string
+	// OIDC token to send to Fulcio instance
+	IdentityToken string
+	// Optional timeout for network requests
+	Timeout time.Duration
+	// Optional version string for user agent
+	LibraryVersion string
+}
+
+func SignerFulcio(opts *FulcioOptions) *Fulcio {
```
Hayden-IO: nit: `NewFulcioClient`?

pkg/sign/signer.go
```diff
@@ -15,29 +15,218 @@
...
+		return nil, err
+	}
+
+	var jwt jsonWebToken
+	err = json.Unmarshal([]byte(jwtString), &jwt)
+	if err != nil {
+		return nil, err
+	}
+
+	// Generate ephemeral keypair
```
Hayden-IO: I don't want to couple artifact signing with requesting Fulcio certificate requests, as this would mean that we wouldn't support self-managed keys and identity certificates.

I would propose that the Sign interface be responsible for signing `content` and producing a bundle, and optionally fetching a certificate. The implementations of that interface could include:
* `EphemeralSigner`, which initializes a key in memory (This is the only one we need for now)
* `FileSigner` (name TBD) which reads and uses a key from disk 
* `KMSSigner`, which reads and uses a key from KMS

Each signer can take in an optional `Fulcio` which can retrieve a certificate for the key material and include it in the bundle. How would this look in practice? Maybe the `Sign` method could instead be `Sign(content Content, opts Options)`? An`Options` struct could include:
* Signature key scheme and hash (from https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_common.proto#L62) (with a default of `PKIX_ECDSA_P256_SHA_256`)
* Optional `Fulcio{}`
* Future: Optional `Rekor{}`
* Maybe: Optional `TimestampAuthority`?

A future improvement could be to have `Fulcio` implement an interface like `CertificateRequest` so you could bring your own CA too. If it's not much work, I might suggest doing that now too if you like that approach.

cc @kommendorkapten @codysoyland @phillmv for your thoughts!

Hayden-IO (changes requested):
I'm good with the size and content of the PR! Was easy to read through and I think tests will be straightforward.

The main comment to discuss is the one about `Sign` being responsible for both certificate requesting and artifact signing, lemme know your thoughts. I don't think it's a significant change from this PR, mostly about providing an interface to request certificates as an optional step of signing.

pkg/sign/timestamping.go
kommendorkapten: How about countersign all signatures? The verification would be a bit tricker as you would need to perform N^2 verifications attempts to match timestamp signature against DSSE signature. But N is small so in practice it won't matter.

pkg/sign/signer.go
kommendorkapten: This works as the structure used is defined to match JSON serialization returned. I've been running similar code for another project.

pkg/sign/signer.go
kommendorkapten: Separating out the key management part from Fulcio is a good idea. I could envision something like this:

```go
func NewFulcioSigner (s crypto.Signer, url string) *FulcioSigner {...} // implements RawSigner

type RawSigner {
    crypto.Signer
    AddVerificationMaterial(pb *protobundle.Bundle)
}
```

Then we can have methods for preparing `RawSigner`s for various key backend such as kms, file or ephemeral. We can add a new layer that wraps a `RawSigner` for embedding public key or cert into the verification material too to avoid repetitive code.
The we create one for DSSE and one for message digest that only takes in a `crypto.Signer` and options (witnesses). I think that could simplify the code a bit as then TSA client would only do one thing, countersign without needing to bother where to get the signature to counter sign, and where to store it? Now that code is in `tsa` module.

Something like (not valid code but I think the idea of wrapping makes it somewhat clear):
```go

type CertSigner struct {
    s crypto.Signer
    cert *x509.Certificate
}

func (s *CertSigner) AddVerificationMaterial(b *protobundle.Bundle) {
    bundle.VerificationMaterial: &protobundle.VerificationMaterial{
			Content: &protobundle.VerificationMaterial_X509Cert{
                             cert
        }
}

type FulcioSigner struct
    CertSigner
    url string
}

func (f *FulcioSigner) GetCert(idToken string) {
        f.cert = f.getCert(url, f.s.PublicKey(), idToken)
}

func (f *FulcioSigner) Sign(...) {
    return f.s.Sign(...)
}

type DsseSigner struct {
    s RawSigner
}

func (d* DsseSigner) Sign(content, opts) error {
    bundle := &protobundle.Bundle{...}   
    sig := d.s.Sign(content.PreAuthEncode(), ...)
    d.s.AddVerificationMaterial(bundle)
    env := Dsse.Envelope{}
    // add content to envelope
    counterSignatures := [][]byte   
    for _, s := range env.Signatures {
        for _, w := range opts.witness {
            cs := w.CounterSign(s)
            bundle.VerificationMaterial.TimestampVerificationData = append(counterSignatures, newTimestampVerificationData(cs))
        }
    }
    bundle.Content = &protobundle.Bundle_DsseEnvelope{...}
}
```

so the idea is that we can compose the final signer with either DSSE/message signature with either key or cert signer, with the different key implementations from kms, file or ephemeral. Fulcio would thus be a super type of the CertSigner to manage to extract the cert (_any_ type of private key material). The client would only need to call the GetCert function after the public key is known, but before it starts to sign.

pkg/sign/timestamping.go
Hayden-IO: What would be the motivation over countersigning each over a concatenation of all? In either case, you're committing to the signatures. The latter does mean we need canonicalization, but would reduce verification cost.

BTW, we can punt this from this PR to avoid this getting larger, lets just follow up on an issue and explicitly reject multi-sig DSSEs.

pkg/sign/signer.go
Hayden-IO: Oh good! I think I ran into this because I was trying to handle unmarshalling either the detached or the embedded. I really just need to mark the former as deprecated...

pkg/sign/content.go
steiza: I agree, but I went with `Add`

pkg/sign/signer.go
Hayden-IO: This composition actually looks a lot like how Cosign implements signing now, wrapping a Sign method (without the bundle updates of course since that's missing from Cosign). Similarly, timestamping and Rekor are implemented to wrap Sign and do something with the result of it.

pkg/sign/timestamping.go
steiza: I based this off of how sigstore-go handles verification today: https://github.com/sigstore/sigstore-go/blob/75b405ded9aee87afa3c29facd08effbd35e43b2/pkg/bundle/signature_content.go#L100

We can change this, but if we decide to do so, we should probably change it in both places.

pkg/sign/signer.go
steiza: I didn't see one elsewhere in `sigstore`, `sigstore-go`, or `protobuf-specs`, but I moved it into a constant.

pkg/sign/timestamping.go
Hayden-IO: Can you file an issue to track this? This is worth standardizing across clients too.

pkg/sign/signer.go
steiza: I ended up going with `NewFulcio`, to not confuse this with something that's just focused on HTTP communication with Fulcio.

pkg/sign/signer.go
steiza: I didn't realize customers were using Fulcio with self-managed keys! I did a bit of a refactor to support that use case, and de-duplicate some code around handling keys.

I do like the interface @kommendorkapten suggested - right now I still feel like we're developing the primitives to talk to the Timestamp Authority, Fulcio, and Rekor. There will probably be some layer on top of what we have today that supports composing these primatives.

pkg/sign/content.go
steiza: Yeah, I don't love having `hashAlgorithm` and `digest` be arguments to `Bundle()` either.

The problem I'm trying to solve is that the `Signer` isn't necessarily going to know what kind of content it's signing, and the `hashAlgorithm` (and associated `digest`) comes from the `Keypair`, not the `Content`.

Actually, now that I type this out, maybe instead of having a public `PreAuthEncoding()` method, `Content` should have a `SetHashAlgorithm` that stores the digest and signature on itself, ready for a call to `Bundle()`.

steiza: Okay I took a pass at having a cleaner separation of responsibilities between `Content` and `Keypair`. The dataflow of the desired hash algorithm still feels a little weird, but the function arguments are more straightforward.

pkg/sign/timestamping.go
steiza: Tracking this on https://github.com/sigstore/sig-clients/issues/14

pkg/sign/signer.go
steiza: I added this above (around line 100).

pkg/sign/keys.go
```diff
@@ -0,0 +1,130 @@
...
+	protocommon "github.com/sigstore/protobuf-specs/gen/pb-go/common/v1"
+	"github.com/sigstore/sigstore/pkg/cryptoutils"
+)
+
+type Keypair interface {
+	GetHashAlgorithm() protocommon.HashAlgorithm
+	GetHint() []byte
+	GetKeyAlgorithm() string
+	GetPublicKeyPem() (string, error)
+	SignDigest(digest []byte) ([]byte, error)
```
Hayden-IO: This won't work with [ed25519](https://pkg.go.dev/crypto/ed25519), which computes the digest as part of signing. While we generate ecdsa keys for now, ideally we don't have to change the interface later to support cryptographic agility.

I see you switched this in the last commit, what was the reason?

pkg/sign/keys.go
```diff
@@ -0,0 +1,130 @@
...
+type EphemeralKeypairOptions struct {
+	// Optional hash algorithm to use to create digest of data provided to sign
+	HashAlgorithm protocommon.HashAlgorithm
+	// Optional hint of for signing key
+	Hint []byte
+}
+
+type EphemeralKeypair struct {
+	options    *EphemeralKeypairOptions
+	privateKey *ecdsa.PrivateKey
```
Hayden-IO: Can this be crypto.PrivateKey instead? Long-term we should take the key algorithm as part of the `Options` struct too, with ecdsa as the default, though I'm fine if we don't want to include that in this PR (maybe a TODO comment?).

pkg/sign/signer.go
```diff
@@ -15,99 +15,240 @@
...
+
+type Fulcio struct {
+	options *FulcioOptions
+}
+
+type FulcioOptions struct {
+	// URL of Fulcio instance
+	BaseURL string
+	// OIDC token to send to Fulcio instance
+	IdentityToken string
```
Hayden-IO: Mentioned above, but I think this should be a part of each request rather than initializing the Fulcio struct, as this makes the Fulcio object harder to reuse.

pkg/sign/signer.go
```diff
@@ -15,99 +15,240 @@
...
 
 	protobundle "github.com/sigstore/protobuf-specs/gen/pb-go/bundle/v1"
 	protocommon "github.com/sigstore/protobuf-specs/gen/pb-go/common/v1"
 )
 
+const bundleV03MediaType = "application/vnd.dev.sigstore.bundle.v0.3+json"
+
 type Signer interface {
-	Sign(data []byte) (*protobundle.Bundle, error)
+	Sign(content Content, k Keypair) (*protobundle.Bundle, error)
```
Hayden-IO: I think this is the wrong composition. This works fine for the expected PGI workflow, but it creates a binding between the key and CA. If a user needs to sign offline for example, it would not possible.

IMO, Fulcio should be responsible only for requesting a certificate. This is why I prefer the composition of either a higher-level object taking a `KeyPair` and `CA` interface or `Fulcio` struct (You had mentioned a similar composition in another comment), or an interface like what Fredrik proposed.

If we want to avoid doing a larger refactor to the interface, what do you think about something like:

```
type CA interface {
  CertificateRequest(key crypto.PublicKey, identity string, bundle *protobundle.Bundle) (error)
}
```

with the `Fulcio` struct implementing this method.

And in the `main.go` example:

```
// init fulcio opts struct
// init bundle
// get ID token
keypair := sign.NewEphemeralKeyPair()
fulcio := NewFulcio(opts)
fulcio.CertificateRequest(keypair.Public(), idtoken, bundle)
keypair.Sign(content, bundle)
```

Those latter steps can later be composed into a single method later on, as part of a `Signer` interface for example.

pkg/sign/keys.go
steiza: I switched this because I was double-hashing the JWT subject - but I put this back to be `SignData`. I think this also answers the question "should `SignData` return a digest" - as I understand it, it has to, to support ed25519.

pkg/sign/keys.go
steiza: I added a TODO - but I think changing this now is more trouble than it's worth, since `crypto.PrivateKey` doesn't include `Sign()`. Since it's unexported, it should be easy to change when we add support for more key types.

pkg/sign/signer.go
steiza: I reworked things quite a bit, so individual components like Fulcio or the TimestampAuthority no longer need to understand the bundle format. These changes also include providing the idToken separately when you request a certificate from Fulcio. It did end up simplifying things a bit, and I think it meets the criteria we've laid out, but let me know if there are further changes you want to see.

pkg/sign/certificate.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	// Assemble bundle from Fulcio response
+	var fulcioResp fulcioResponse
+	err = json.Unmarshal(body, &fulcioResp)
+	if err != nil {
+		return nil, err
+	}
+
+	certs := fulcioResp.SctCertWithChain.Chain.Certificates
```
Hayden-IO, line 165: Can be a TODO for later, should we verify the result? This would mean that the TUF repository needs to be initialized for signing, but I think this is reasonable, to prevent accidentally returning something unverifiable.

pkg/sign/certificate.go
```diff
@@ -0,0 +1,175 @@
...
+
+type chain struct {
+	Certificates []string `json:"certificates"`
+}
+
+func NewFulcio(opts *FulcioOptions) *Fulcio {
+	return &Fulcio{options: opts}
+}
+
+func (f *Fulcio) GetCertificate(keypair Keypair, identityToken string) ([]byte, error) {
```
Hayden-IO, line 78: Can you add a comment on the function that this returns the DER-encoded code signing certificate?

pkg/sign/signer.go
```diff
@@ -15,77 +15,13 @@
...
-type signedCertificateEmbeddedSct struct {
-	Chain chain `json:"chain"`
-}
-
-type chain struct {
-	Certificates []string `json:"certificates"`
-}
-
-func (f *Fulcio) Sign(content Content, keypair Keypair) (*protobundle.Bundle, error) {
+func Bundle(content Content, keypair Keypair, fulcio *Fulcio, idToken string, timestampAuthority *TimestampAuthority) (*protobundle.Bundle, error) {
```
Hayden-IO, line 24: I'm good with this for now while we're evolving this, but before stabilizing the API, we should figure out this method, since it wouldn't be possible to add additional steps without it being a breaking change.

pkg/sign/certificate.go
Hayden-IO: Actually, as I review the rest of this, maybe we should do verification in `sign.Bundle`? We can punt this, but should probably track this FR somewhere.

Hayden-IO (commented):
Thanks so much for the changes, this looks great. I have some thoughts on the top level `sign.Bundle` method, but I think it'll naturally evolve as we continue adding features. Just a couple last comments.

pkg/sign/certificate.go
steiza: 👍 I added a TODO comment for now in `signer.go`

pkg/sign/signer.go
steiza: Yeah, we'll probably want to encapsulate these in a `SignerOpts` / `BundleOpts` or similar.

Hayden-IO (dismissed):
Thanks again for all the discussion here! @kommendorkapten any other thoughts?

pkg/sign/keys.go
```diff
@@ -0,0 +1,135 @@
...
+	if opts == nil {
+		opts = &EphemeralKeypairOptions{}
+	}
+
+	privateKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	if err != nil {
+		return nil, err
+	}
+
+	if opts.HashAlgorithm == protocommon.HashAlgorithm_HASH_ALGORITHM_UNSPECIFIED {
```
kommendorkapten: I find this a bit unexpected. The hash algorithm can be specified, but we always create a key using the P256 curve. While this works (digests larger than bit length of the private key are truncate) I think it may be confusing. How about flipping this around? The options can specify a curve (P256 as default) and based on the curve we derive the hash function?

pkg/sign/signer.go
```diff
@@ -15,99 +15,74 @@
...
+const bundleV03MediaType = "application/vnd.dev.sigstore.bundle.v0.3+json"
 
-// TODO: implement
-// type Fulcio struct {
-//	baseURL       string
-//	identityToken string
-// }
+func Bundle(content Content, keypair Keypair, fulcio *Fulcio, idToken string, timestampAuthority *TimestampAuthority) (*protobundle.Bundle, error) {
+	// See if keypair was provided
+	if keypair == nil {
```
kommendorkapten: I'm not sure about this. For critical things such as signing I would suggest we require all parameters explicitly.

kommendorkapten:
I think this looks good for an incremental PR! I have two comments I would like to see answered before merging.

pkg/sign/keys.go
steiza: Fair point! Similar to how we're punting on different key types, I think for now we're going to just assume P256 and later decide how we want to support different key types, parameters, and hash algorithms.

pkg/sign/signer.go
steiza: Yeah, I was thinking that in the frequent Fulcio case, people would just want an ephemeral keypair.

But it's very easy to instantiate an ephemeral keypair with `pkg/sign/keys.go`, and I can see the point about this being better if its explicit, so I removed this.

pkg/sign/signer.go
```diff
@@ -22,16 +22,6 @@ import (
 const bundleV03MediaType = "application/vnd.dev.sigstore.bundle.v0.3+json"
 
 func Bundle(content Content, keypair Keypair, fulcio *Fulcio, idToken string, timestampAuthority *TimestampAuthority) (*protobundle.Bundle, error) {
```
Hayden-IO, line 24: Can we document what is and isn't required, and also err out when the required params aren't set? content and keypair are always, idtoken is required only when fulcio is set.

kommendorkapten (approved):

kommendorkapten:
Nice work @steiza !

pkg/sign/signer.go
steiza: I promise that I will do this in a sec when I post Rekor support.

Hayden-IO (approved):

PR #163, opened 2024-05-12
chore: fix undefined media type error in OCI verification example

vishal-chdhry:
#### Summary

`SigstoreBundleMediaType01` was removed in commit: https://github.com/sigstore/sigstore-go/commit/20c2ce9c55fc2d05bf01604c491edfe46e43e4dc but it is still referenced in OCI verification example, this PR fixes that by replacing it with `bundle.MediaTypeString()` function


vishal-chdhry:
cc @steiza

Hayden-IO (approved):

PR #172, opened 2024-05-13
Add Rekor support to signing

steiza:
#### Summary

Continuing to work towards https://github.com/sigstore/sigstore-go/issues/136 with small / incremental diffs.

#### Release Note

NONE - still need to review signing interface, add tests, and enable conformance testing

#### Documentation

NONE

### DIFF
--- a/pkg/sign/signer.go
+++ b/pkg/sign/signer.go
@@ -15,13 +15,31 @@
 package sign
 
 import (
+	"encoding/pem"
+
 	protobundle "github.com/sigstore/protobuf-specs/gen/pb-go/bundle/v1"
 	protocommon "github.com/sigstore/protobuf-specs/gen/pb-go/common/v1"
 )
 
 const bundleV03MediaType = "application/vnd.dev.sigstore.bundle.v0.3+json"
 
-func Bundle(content Content, keypair Keypair, fulcio *Fulcio, idToken string, timestampAuthority *TimestampAuthority) (*protobundle.Bundle, error) {
+type BundleOptions struct {
+	// Optional Fulcio instance to get code signing certificate from.
+	//
+	// Resulting bundle will contain a certificate for its verification
+	// material content, instead of a public key.
+	Fulcio *Fulcio
+	// Optional OIDC JWT to send to Fulcio; required if using Fulcio
+	IDToken string
+	// Optional list of timestamp authorities to contact for inclusion in bundle
+	TimestampAuthorities []*TimestampAuthority
+	// Optional Rekor instance to get transparency log entry from.
+	//
+	// Supports hashedrekord and dsse entry types
+	Rekor *Rekor
+}
+
+func Bundle(content Content, keypair Keypair, opts BundleOptions) (*protobundle.Bundle, error) {
 	bundle := &protobundle.Bundle{MediaType: bundleV03MediaType}
 
 	// Sign content and add to bundle
@@ -33,8 +51,9 @@ func Bundle(content Content, keypair Keypair, fulcio *Fulcio, idToken string, ti
 	content.Bundle(bundle, signature, digest, keypair.GetHashAlgorithm())
 
 	// Add verification information to bundle
-	if fulcio != nil {
-		certBytes, err := fulcio.GetCertificate(keypair, idToken)
+	var pubKeyPEM []byte
+	if opts.Fulcio != nil && opts.IDToken != "" {
+		pubKeyBytes, err := opts.Fulcio.GetCertificate(keypair, opts.IDToken)
 		if err != nil {
 			return nil, err
 		}
@@ -42,10 +61,15 @@ func Bundle(content Content, keypair Keypair, fulcio *Fulcio, idToken string, ti
 		bundle.VerificationMaterial = &protobundle.VerificationMaterial{
 			Content: &protobundle.VerificationMaterial_Certificate{
 				Certificate: &protocommon.X509Certificate{
-					RawBytes: certBytes,
+					RawBytes: pubKeyBytes,
 				},
 			},
 		}
+
+		pubKeyPEM = pem.EncodeToMemory(&pem.Block{
+			Type:  "CERTIFICATE",
+			Bytes: pubKeyBytes,
+		})
 
 		// TODO: do verification of Fulcio certificate
 	} else {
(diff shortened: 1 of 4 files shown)


pkg/sign/signer.go
```diff
@@ -15,13 +15,31 @@
...
+	// material content, instead of a public key.
+	Fulcio *Fulcio
+	// Optional OIDC JWT to send to Fulcio; required if using Fulcio
+	IDToken string
+	// Optional list of timestamp authorities to contact for inclusion in bundle
+	TimestampAuthorities []*TimestampAuthority
+	// Optional Rekor instance to get transparency log entry from.
+	//
+	// Supports hashedrekord and dsse entry types
+	Rekor *Rekor
```
Hayden-IO: Can we make this a list too?

pkg/sign/signer.go
```diff
@@ -15,13 +15,31 @@
...
+}
+
+func Bundle(content Content, keypair Keypair, opts BundleOptions) (*protobundle.Bundle, error) {
 	bundle := &protobundle.Bundle{MediaType: bundleV03MediaType}
 
 	// Sign content and add to bundle
```
Hayden-IO, line 54: Can we fail early if keypair == nil?

pkg/sign/signer.go
```diff
@@ -33,20 +51,26 @@ func Bundle(content Content, keypair Keypair, fulcio *Fulcio, idToken string, ti
 	content.Bundle(bundle, signature, digest, keypair.GetHashAlgorithm())
 
 	// Add verification information to bundle
-	if fulcio != nil {
-		certBytes, err := fulcio.GetCertificate(keypair, idToken)
+	var pubKeyPEM []byte
+	if opts.Fulcio != nil && opts.IDToken != "" {
```
Hayden-IO, line 64: Same here, can we return an error if IDToken is not set?

pkg/sign/signer.go
```diff
@@ -33,20 +51,26 @@ func Bundle(content Content, keypair Keypair, fulcio *Fulcio, idToken string, ti
 	content.Bundle(bundle, signature, digest, keypair.GetHashAlgorithm())
 
 	// Add verification information to bundle
-	if fulcio != nil {
-		certBytes, err := fulcio.GetCertificate(keypair, idToken)
+	var pubKeyPEM []byte
```
Hayden-IO: nit on naming, can we say "verifierPEM" or something similar? This was one issue I had with Cosign naming, that it perpetuated the use of public keys when certificates are what we prefer.

The naming for the Rekor client LGTM since we do call the verifier a `PublicKey`.

pkg/sign/transparency.go
```diff
@@ -0,0 +1,198 @@
...
+	if err != nil {
+		return err
+	}
+
+	entry := resp.Payload[resp.ETag]
+	keyID, err := hex.DecodeString(*entry.LogID)
+	if err != nil {
+		return err
+	}
+
```
Hayden-IO, line 132: Can you use https://github.com/sigstore/rekor/blob/92cf4d5d3682ffcbf4138341f1bc5f7a66766549/pkg/tle/tle.go#L33 instead to construct the TLE struct?

Hayden-IO (commented):
Just a few small comments, this looks great!

How do you think this should be tested? I would like to see us reuse what we built in Cosign with hermetic infrastructure.

pkg/sign/transparency.go
```diff
@@ -0,0 +1,140 @@
...
+	Timeout time.Duration
+	// Optional version string for user agent
+	LibraryVersion string
+}
+
+func NewRekor(opts *RekorOptions) *Rekor {
+	return &Rekor{options: opts}
+}
+
+func (r *Rekor) GetTransparencyLog(pubKeyPEM []byte, b *protobundle.Bundle) error {
```
Hayden-IO: naming nit, can this be `GetTransparencyLogEntry`?

pkg/sign/transparency.go
```diff
@@ -0,0 +1,140 @@
...
+		return err
+	}
+
+	resp, err := client.Entries.CreateLogEntry(params)
+	if err != nil {
+		return err
+	}
+
+	entry := resp.Payload[resp.ETag]
+	tlogEntry, err := tle.GenerateTransparencyLogEntry(entry)
```
Hayden-IO, line 128: Unrelated to this PR, I think I noticed a bug - https://github.com/sigstore/rekor/blob/92cf4d5d3682ffcbf4138341f1bc5f7a66766549/pkg/tle/tle.go#L88 doesn't base64 decode the SET before setting it, which is different than what you had here. Can you verify if this is a bug?

Hayden-IO: Hm, maybe not? https://github.com/sigstore/cosign/blob/121115774e8c662648165f2924e7354292365ae8/pkg/cosign/tlog.go#L492 doesn't base64-decode it.

steiza: I did generate a bundle with these changes, and it verified with a policy that required an observer timestamp (and it didn't have a timestamp authority entry), so I believe this is working:

```
$ cat bundle.json | jq -r ".verificationMaterial.tlogEntries[0].inclusionPromise.signedEntryTimestamp" | base64 -d
0E J??s??U?P??7????43?$D
                        <s;?b???!ԟ?:
                                    ?Ljj-K??>S?j??mv??????*	%
```

steiza:
> How do you think this should be tested? I would like to see us reuse what we built in Cosign with hermetic infrastructure.

Yeah, I think I've put off testing for about as long as I can. Now that the interfaces are mostly settled, I can write some unit tests that exercise code paths (but aren't necessarily looking at cryptographic correctness - I think that will mostly be driven by the conformance tests).

I poked around Cosign a bit but I'm not sure I understand what you mean by hermetic infrastructure - can you link me to some examples?

pkg/sign/transparency.go
Hayden-IO: If I were to guess, during marshalling, it doesn't re-encode it as base64. I'll play around with it offline, but if this is working, perfect.

Hayden-IO (approved):

Hayden-IO:
@cmurphy has refactored Cosign's end-to-end tests to rely on local instances of Fulcio, Rekor, an identity provider, and a timestamp authority. Set up is in https://github.com/sigstore/cosign/blob/main/test/e2e_test.sh, and runs tests in https://github.com/sigstore/cosign/blob/main/test/e2e_test.go (and a few other e2e_test_.* files). 

End-to-end tests work well for the signing path so that we don't have to deal with mocked clients.

I think it'll be relatively straightforward to reuse this in sigstore-go, initializing the Rekor and Fulcio clients with the local instances spun up from the shell script.

pkg/sign/transparency.go
Hayden-IO: Just to close out on this, the difference between your previous change and this was that you base64 decoded `e.Verification.SignedEntryTimestamp.String()`, which outputs a base64-encoded SET. Without calling `MarshalJSON` or `String`, you're working on the bytes.

PR #173, opened 2024-05-15
Add clarifying comments around TUF caching options

### LINKED ISSUE #162: bug: fix WithCacheValidity
**Description**

I'm planning to use the tuf client's `WithCacheValidity()` option. When I use`.WithCacheValidity(365 * 2025)`, it works, but not with `2024`. It seems like the initial `LastTimestamp` is [initially](https://github.com/sigstore/sigstore-go/blob/75b405ded9aee87afa3c29facd08effbd35e43b2/pkg/tuf/config.go#L25-L35) set to `0001-01-01 00:00:00 +0000 UTC` when using the [embedded](https://github.com/sigstore/sigstore-go/blob/75b405ded9aee87afa3c29facd08effbd35e43b2/pkg/tuf/options.go#L27-L28) tuf root.

I'm thinking it should maybe be set to the same day the embedded root was updated, so instead I could just invoke with `.WithCacheValidity(7)`, to say that I want to use my local cache up to 7 days after after refresh. Or if it's the embedded cache, 7 days after the root was published or embedded by maintainers.

**Version**

sigstore-go@v0.3.0

Hayden-IO:
This changes the behavior of the TUF client to use the cached metadata if it's unexpired on first initialization. Previously, if no config which included the last update timestamp was found, it would force a reinitialization. Now, the first init with cached metadata will be offline.

Fixes https://github.com/sigstore/sigstore-go/issues/162

### DIFF
--- a/pkg/tuf/client_test.go
+++ b/pkg/tuf/client_test.go
@@ -167,6 +167,23 @@ func TestCache(t *testing.T) {
 	target, err = c.GetTarget("foo")
 	assert.NoError(t, err)
 	assert.Equal(t, target, []byte("foo version 2"))
+
+	r.AddTarget("foo", []byte("foo version 3"))
+
+	// Delete config to show that client fetches fresh metadata when no config is present
+	if err = os.Remove(c.configPath()); err != nil {
+		t.Fatal(err)
+	}
+
+	// Create another new client with the same cache path
+	c, err = New(opt)
+	assert.NotNil(t, c)
+	assert.NoError(t, err)
+
+	// Cache contains new version
+	target, err = c.GetTarget("foo")
+	assert.NoError(t, err)
+	assert.Equal(t, target, []byte("foo version 3"))
 }
 
 func TestExpiredTimestamp(t *testing.T) {
--- a/pkg/tuf/options.go
+++ b/pkg/tuf/options.go
@@ -36,13 +36,20 @@ const (
 
 // Options represent the various options for a Sigstore TUF Client
 type Options struct {
-	// CacheValidity period in days (default 0). Note that the client will
-	// always refresh the cache if the metadata is expired, so this is not an
-	// optimal control for air-gapped environments. Use const MaxCache to only
-	// update the cache when the metadata is expired.
+	// CacheValidity period in days (default 0). The client will persist a
+	// timestamp with the cache after refresh. Note that the client will
+	// always refresh the cache if the metadata is expired or if the client is
+	// unable to find a persisted timestamp, so this is not an optimal control
+	// for air-gapped environments. Use const MaxCache to update the cache when
+	// the metadata is expired, though the first initialization will still
+	// refresh the cache.
 	CacheValidity int
 	// ForceCache controls if the cache should be used without update
-	// as long as the metadata is valid
+	// as long as the metadata is valid. Use ForceCache over CacheValidity
+	// if you want to always use the cache up until its expiration. Note that
+	// the client will refresh the cache once the metadata has expired, so this
+	// is not an optimal control for air-gapped environments. Clients instead
+	// should provide a trust root file directly to the client to bypass TUF.
 	ForceCache bool
 	// Root is the TUF trust anchor
 	Root []byte


Hayden-IO:
cc @ramonpetgrave64

Hayden-IO:
@kommendorkapten, lemme know your thoughts on this. I agree with [your comment](https://github.com/sigstore/sigstore-go/issues/162#issuecomment-2111650002) that a good solution for offline verification is to pass a trust root file out of band, which Ramon and I discussed, but I think the behavior in this PR is a nice default so that users do not need to be aware of the timestamp config.

pkg/tuf/client.go
```diff
@@ -105,9 +105,14 @@ func (c *Client) loadMetadata() error {
 	} else if c.opts.CacheValidity > 0 {
 		cfg, err := LoadConfig(c.configPath())
 		if err != nil {
-			// Config may not exist, don't error
-			// create a new empty config
-			cfg = &Config{}
+			// Config will not exist during the first initialization
+			// of unexpired metadata
+			cfg = &Config{
+				LastTimestamp: time.Now(),
```
ramonpetgrave64: I haven't tried yet, but does this mean that the client will assume that my local cache is already fresh, without first checking `~/.sigstore/root/tuf-repo-cdn.sigstore.dev.json`?

Hayden-IO: If the cache is already fresh, meaning `c.up.Refresh()` succeeded, and no configuration file exists, then no additional refresh will occur. This addition persists the current time so that on the next metadata load, the config will be checked.

ramonpetgrave64: By fresh, I meant that my local copy up-to-date **before** creating the client.

Hayden-IO: `c.up.Refresh` when `UnsafeLocalRefresh` is set does not make any network calls, it simply loads the metadata from disk. It will err out if the timestamp is out of date, that we cannot and should not bypass.

Does that answer your question?

kommendorkapten:
I don't see any real issues with this PR (but I would prefer to only write the timestamp file upon sigstore client initiated tuf refresh), but I do think that the _cache is being misused_ if there is a concern about **one** egress TUF update on the first invocation. 

The cache is not about prohibiting network calls to be made, as the timestamp can expire at any time. It's only about avoiding excessive tuf updates if a program is called within a "tight loop".

Consider the scenario where the tuf cache is populated by some process at time `X_1`, timestamp expires at `X_2` and the the sigstore client is started at `X_3`. This will force a full TUF update as timetsamp is expired, so trying to use the cache to avoid network calls is doomed to fail.

ramonpetgrave64:
> but I do think that the cache is being misused if there is a concern about one egress TUF update on the first invocation.

That's understandable. For the use-cases I'm imagining, I would be using a separate process to provide fresh caches.

pkg/tuf/client.go
ramonpetgrave64: Those options seem like the right thing I could use. Would you make those publicly accessible?

Hayden-IO:
> The cache is not about prohibiting network calls to be made, as the timestamp can expire at any time. It's only about avoiding excessive tuf updates if a program is called within a "tight loop".

If I set cache validity, to me, I want to continue to use TUF metadata up until the cache is expired as per the validity timestamp. `ForceCache` is the extreme of that, which is "trust the metadata until it's expired". Currently, if I set that flag without a locally saved timestamp, then it works as intended. If I set `CacheValidity`, it doesn't, as it will fetch TUF metadata on first client initialization.

So either:
* We should go through with this PR to reconcile the above, or
* Update documentation on what `CacheValidity`'s purpose is

pkg/tuf/client.go
Hayden-IO: I don't think we should expose `UnsafeLocalRefresh`, but instead make sure we have caching options that match expected use cases. Which I think we do have currently, besides what's in this PR.

The more I look at this, I think that `ForceCache` is what most users should be using, and that `CacheValidity` doesn't have a clear use case.

pkg/tuf/client.go
```diff
@@ -105,9 +105,14 @@ func (c *Client) loadMetadata() error {
 	} else if c.opts.CacheValidity > 0 {
 		cfg, err := LoadConfig(c.configPath())
```
ramonpetgrave64: I tested this with a fresh local cache (in a previous invocation, I let the tuf client do and update, like normally). But in the next invocation, `LoadConfig` does not error, like I think your comment is suggesting. Printed, It actually has the value `&{0001-01-01 00:00:00 +0000 UTC}`.

kommendorkapten:
> If I set cache validity, to me, I want to continue to use TUF metadata up until the cache is expired as per the validity timestamp

I agree, and I argue it does except for the first time as we have no idea on how old the cache actually is. The concern I have here, is that _if_ there is an existing TUF cache, we don't know how old it is. We read it, and if it's valid sets the `lastTimestamp` as if it was updated at this point in time, which may not be the case. That's why it current performs a TUF refresh before writing this timestamp. Looking at the extreme scenario here (metadata cache is written a long time before the sigstore tuf client is invoked), if an existing cache is present, we may use the cached metadata that is expired given the configured cache validity (assuming metadata is not expired) with the proposed change.

Clarifying this in the documentation is what I would prefer.

steiza:
There is a lot going on in this thread! I agree it isn't exactly clear who should own what, and how to do things like offline verification.

What we decided [for the gh CLI](https://github.com/cli/cli/pull/8997) is that if you specify a `trusted_root.json` file we won't instantiate TUF clients and thus won't do network requests to support offline verification.

That pull request also sets the TUF cache to 1 day to reduce (but not eliminate) TUF client network calls.

Hayden-IO:
Thanks all for the comments. I think the behavior of the initial network call might be a little unexpected still, but I recognize there isn't really a perfect way to do this - either we just implicitly cache like in this PR or we fetch the latest state.

Tomorrow, I'll update this PR to be comment updates that specify a) the purpose of CacheValidity, b) when to use it vs ForceCache, c) the preferred way to do offline verification by providing a trusted_root file.

kommendorkapten:
thanks @haydentherapper, especially around using an `trusted_root` file for true offline mode, as that's something that has been discussed in many places/repos and we always come to the conclusion that using TUF in those scenarios is not the best choice. Having a single place with documentation to refer to would be great! 
And clarify that caching is only "best effort" and that network calls are inevitable, and if those fails (due to air-gapped environment) the program will crash.

Hayden-IO:
@kommendorkapten Updated the comments, and I left in the test I had added since I think it's useful to demonstrate that CacheValidity will still update a local cache without a timestamp.

ramonpetgrave64 (approved):

kommendorkapten (approved):
Thanks!

PR #179, opened 2024-05-16
Add conformance testing and unit testing for signing

steiza:
#### Summary

We're working towards signing support (#136) - this adds unit testing to signing as well as conformance testing.

#### Release Note

#### Documentation

N/A

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -42,10 +42,8 @@ jobs:
       - uses: sigstore/sigstore-conformance@ee4de0e602873beed74cf9e49d5332529fe69bf6 # v0.0.11
         with:
           entrypoint: ${{ github.workspace }}/conformance
-          skip-signing: true
 
       - uses: sigstore/sigstore-conformance@ee4de0e602873beed74cf9e49d5332529fe69bf6 # v0.0.11
         with:
           entrypoint: ${{ github.workspace }}/conformance
-          skip-signing: true
           environment: staging
--- a/cmd/conformance/main.go
+++ b/cmd/conformance/main.go
@@ -23,25 +23,33 @@ import (
 	"os"
 	"path"
 	"runtime"
+	"time"
 
 	protobundle "github.com/sigstore/protobuf-specs/gen/pb-go/bundle/v1"
 	protocommon "github.com/sigstore/protobuf-specs/gen/pb-go/common/v1"
+	"google.golang.org/protobuf/encoding/protojson"
 
 	"github.com/sigstore/sigstore-go/pkg/bundle"
 	"github.com/sigstore/sigstore-go/pkg/root"
+	"github.com/sigstore/sigstore-go/pkg/sign"
 	"github.com/sigstore/sigstore-go/pkg/tuf"
 	"github.com/sigstore/sigstore-go/pkg/verify"
 )
 
+var Version string
 var bundlePath *string
 var certPath *string
 var certOIDC *string
 var certSAN *string
+var identityToken *string
 var signaturePath *string
+var staging = false
 var trustedRootPath *string
 
 func usage() {
 	fmt.Println("Usage:")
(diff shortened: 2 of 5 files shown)


cmd/conformance/main.go
```diff
@@ -93,33 +101,135 @@ func getTrustedRoot(staging bool) root.TrustedMaterial {
 	return tr
 }
 
+func parseArgs() {
+	for i := 2; i < len(os.Args); {
```
Hayden-IO, line 105: Could we use [flags](https://pkg.go.dev/flag) rather than parsing the arguments?

.github/workflows/conformance.yml
```diff
@@ -42,10 +42,8 @@ jobs:
       - uses: sigstore/sigstore-conformance@ee4de0e602873beed74cf9e49d5332529fe69bf6 # v0.0.11
         with:
           entrypoint: ${{ github.workspace }}/conformance
-          skip-signing: true
 
       - uses: sigstore/sigstore-conformance@ee4de0e602873beed74cf9e49d5332529fe69bf6 # v0.0.11
         with:
           entrypoint: ${{ github.workspace }}/conformance
```
Hayden-IO, line 48: Just wondering, I assume that some tests will still fail because we don't have conformance hooked up for DSSE signing, correct? How does this surface?

cmd/conformance/main.go
```diff
@@ -23,25 +23,33 @@ import (
...
 var certOIDC *string
 var certSAN *string
+var identityToken *string
 var signaturePath *string
+var staging = false
 var trustedRootPath *string
 
 func usage() {
 	fmt.Println("Usage:")
+	fmt.Printf("\t%s sign --identity-token TOKEN --signature FILE --certificate FILE [--staging] FILE", os.Args[0])
```
Hayden-IO, line 51: Is this needed because we have conformance tests for detached metadata?

Hayden-IO (dismissed):

cmd/conformance/main.go
steiza: It's part of the [conformance CLI protocol](https://github.com/sigstore/sigstore-conformance/blob/main/docs/cli_protocol.md) and it looks like there are tests that [test signing with detached materials](https://github.com/sigstore/sigstore-conformance/blob/main/test/test_signature_verify.py#L67-L80).

.github/workflows/conformance.yml
steiza: This is my first time really digging in the signing side of conformance tests. We do [test DSSE verification](https://github.com/sigstore/sigstore-conformance/blob/main/test/test_bundle.py#L33-L44) but I don't think we test DSSE signing today, since there [isn't a way to specify it in the CLI protocol](https://github.com/sigstore/sigstore-conformance/blob/main/docs/cli_protocol.md).

I created https://github.com/sigstore/sigstore-conformance/issues/145 to track this request.

cmd/conformance/main.go
steiza: I thought about this, but I don't think it'll work because the [first argument in the CLI protocol](https://github.com/sigstore/sigstore-conformance/blob/main/docs/cli_protocol.md) (e.g. `sign` / `sign-bundle` / ...) is not a flag.

.github/workflows/conformance.yml
Hayden-IO: Ah! This actually wasn't what I was asking originally, but I'm glad this was noticed. I was thinking that because in our conformance test, we are only setting up message signing, we would see a failure and need an XFAIL.

cmd/conformance/main.go
Hayden-IO: SG

.github/workflows/conformance.yml
steiza: The conformance tests at v0.0.11 all pass for me locally - I think this will work.

Do note that the signing conformance tests aren't running on this pull request, since this hasn't landed in `main` yet.

Hayden-IO (approved):

Issue #74, opened 2024-01-09
Allow configurable signing algorithms


ret2libc:
**Description**

*@tetsuo-cpp filed similar issues under Cosign and Rekor. We realise there's a lot of overlap in maintainers, but wanted to make sure that we discuss each project that we plan to touch. Apologies if this feels a bit spammy.*

Hi there! At Trail of Bits, we're looking at potentially implementing part of the [Configurable Crypto Algorithms](https://docs.google.com/document/d/1-wF1t6lmJO37BzStbyz8ZkDhNm86HyvKZSaDl7uyHxI) proposal (specifically Phase 1). We wanted to float this idea to each of the relevant Sigstore sub-projects so we can hash out the details in a more concrete way.

Across the Sigstore stack, we default to using ECDSA for signatures and SHA256 for hashing. There's more detail in the linked proposal but there are a number of motivations for wanting to customise the signatures that are generated, including paving the way for post-quantum signatures. The proposed design includes having a "supported algorithm" registry (perhaps this can go in the Protobuf specs) that enumerates the approved signature/hash algorithm combinations. We specifically don't want to allow arbitrary mixing and matching of signature and hash algorithm to avoid some of the security pitfalls listed in the proposal.

For sigstore-go, we want to support this set of approved signing algorithms. This can be as simple as a --signing-algorithm flag. As a first pass, we'd like to support ECDSA with SHA256 and SHA384 as well as EdDSA, with ECDSA-SHA256 remaining as the default.

steiza:
Hello! I think this should be relatively straightforward as `sigstore-go` today just supports verification, not signing. We can add additional signing algorithm support as needed to things like `pkg/verify/signature.go` and `pkg/root/trusted_root.go`.

woodruffw:
xref: https://github.com/sigstore/protobuf-specs/issues/189 will ultimately standardize the full registry of signing algorithms, but doesn't block initial work here (since we know a couple of ECDSA variants that'll need verification support already).

tetsuo-cpp:
One requirement of this work is that verifying clients will need to take information about what signing algorithm was used to generate the signature since we can no longer assume SHA-256.

I think that this information can be part of the Sigstore bundle's `VerificationMaterials`, but each client including `sigstore-go` will require a `--signing-algorithm` flag for the non-bundle verification path. In the bundle spec, it might make sense to make this field optional and have clients fallback to ECDSA-SHA256 to avoid breaking changes.

[...]

PR #181, opened 2024-05-18
Support additionnal hash algorithms for artifact signing

alexiscampan:


Hayden-IO (changes requested):
Hey, thanks for the PR. The issue for log key types wasn’t about more hash algorithms for the log, it was to support different key algorithms. sha256 is fairly standard for log deployment at this point, and I’d prefer to not add support for other digests if it’s not easily possible to deploy such a log. 

As for the second part of this change to add more hash algorithms for artifact signing, this looks good at a glance. Can you add tests for this too?

alexiscampan:
Hey @haydentherapper, thanks for the review. I've just added some tests for the artifact signing part and refined the scope of this PR. I'll do the support of new algorithms for the verify part in another PR.

cmd/conformance/main.go
```diff
@@ -117,6 +118,8 @@ func main() {
 				trustedRootPath = &os.Args[i+1]
 			case "--staging":
 				staging = true
+			case "--signing-algorithm":
```
steiza, line 121: We want the conformance command line options to match https://github.com/sigstore/sigstore-conformance/blob/main/docs/cli_protocol.md. If we decide to continue with this change, we could add this as an option to `examples/signing/main.go` (for signing) or `cmd/sigstore-go/main.go` (for verifying).

pkg/sign/keys.go
```diff
@@ -39,8 +40,8 @@ type Keypair interface {
 
 type EphemeralKeypairOptions struct {
 	// Optional hint of for signing key
-	Hint []byte
-	// TODO: support additional key algorithms
+	Hint          []byte
+	HashAlgorithm protocommon.HashAlgorithm
```
steiza, line 44: So this was [actually proposed before](https://github.com/sigstore/sigstore-go/pull/158/files/241055779db16fd93ddd2d64551a79ff52b6c62f#diff-04ec1e79d9617f37f1eada53044d0c9cfaac07ed9ffd7a35e8c22d1bde7e0d36R42), and [we decided not to allow hash algorithm options for ECDSA P256](https://github.com/sigstore/sigstore-go/pull/158#discussion_r1598413552), because even if you specify a hash with more bits it will get truncated to match the key size.

I guess to take a step back, what problem are we trying to solve here? What use-case are we trying to enable?

steiza (changes requested):

pkg/sign/keys.go
Hayden-IO: I'd like to see cryptographic agility. Agreed that specifying the hash here doesn't make sense.

Instead of `HashAlgorithm`, should we be taking in `PublicKeyDetails` - https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_common.proto#L62, and initializing both a signing algorithm and hash algorithm based on that.

Hayden-IO (changes requested):

pkg/sign/keys.go
steiza: Ah, yeah, we can support some additional keys in https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_common.proto#L62, especially for key types and hashes we expect to be very common.

But I don't think that `pkg/sign/keys.go` should implement every possible combination. The point of the `Keypair` interface is so that `sigstore-go` doesn't have to implement every key type - you can provide your own implementation to that interface.

That's why I was curious what @alexiscampan's use-case is. If they were anticipating lots of clients needing to sign with a specific key type and hash, or if it would make more sense for them to implement their own `Keypair` when they use `sigstore-go`.

Hayden-IO: That makes sense for KMS keys or other key interfaces that pull in dependencies or experimental algorithms, e.g. PQC. Or for signing where keys are managed, e.g on disk or HSMs.

For algorithms that are natively supported in Go and where keys are used in ephemeral signing, why not support them here too? The default of ECDSA-P256 is reasonable, but in the spirit of cryptographic agility (https://github.com/sigstore/sigstore-go/issues/74), I would like to add in a way to configure RSA, ed25519, and other elliptic curves. I recognize this is a case of YAGNI, but it's straightforward to add this support and removes a barrier for potential integrators while still directing them to the default, preferred Sigstore signing workflow.

alexiscampan: > Ah, yeah, we can support some additional keys in https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_common.proto#L62, especially for key types and hashes we expect to be very common.
> 
> But I don't think that `pkg/sign/keys.go` should implement every possible combination. The point of the `Keypair` interface is so that `sigstore-go` doesn't have to implement every key type - you can provide your own implementation to that interface.
> 
> That's why I was curious what @alexiscampan's use-case is. If they were anticipating lots of clients needing to sign with a specific key type and hash, or if it would make more sense for them to implement their own `Keypair` when they use `sigstore-go`.

Hey @steiza, I have no particular use case for this specific feature. With my company we are going to need the signing part, we could use other librairies like the one in python to sign our images but all our applications are written in Go so we wanted to stay consistant. I also think that the default ECDSA-P256 is sufficient for our use case. I just wanted to do a first contribution before contributing on the signing part. I will close this PR for now as the priority of this feature is not high and we don't really know if it will be useful. I will try to give some support on the signing part which is the main feature of this library. Thanks for your fast review.

PR #184, opened 2024-05-21
Add optional retries for signing with Fulcio, Rekor, and Timestamp Authority

steiza:
#### Summary

As we work towards releasing signing support (#136) we don't want an intermittent network issue to stop a bundle from being created.

#### Release Note

#### Documentation

N/A

### DIFF
--- a/pkg/sign/certificate.go
+++ b/pkg/sign/certificate.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"net/http"
 	"strings"
 	"time"
@@ -36,6 +37,8 @@ type FulcioOptions struct {
 	BaseURL string
 	// Optional timeout for network requests
 	Timeout time.Duration
+	// Optional number of times to retry on HTTP 5XX
+	Retries uint
 	// Optional version string for user agent
 	LibraryVersion string
 }
@@ -122,7 +125,6 @@ func (f *Fulcio) GetCertificate(keypair Keypair, identityToken string) ([]byte, 
 	if err != nil {
 		return nil, err
 	}
-	requestBytes := bytes.NewBuffer(requestJSON)
 
 	// TODO: For now we are using our own HTTP client
 	//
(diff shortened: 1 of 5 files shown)


pkg/sign/certificate.go
```diff
@@ -133,17 +135,31 @@ func (f *Fulcio) GetCertificate(keypair Keypair, identityToken string) ([]byte,
...
+		request.Header.Add("Authorization", "Bearer "+identityToken)
+		request.Header.Add("Content-Type", "application/json")
+		request.Header.Add("User-Agent", constructUserAgent(f.options.LibraryVersion))
+
+		response, err = client.Do(request)
+		if err != nil {
+			return nil, err
+		}
+
+		if !(response.StatusCode >= 500 && response.StatusCode < 600) {
```
Hayden-IO: 429 as well, as that's a recoverable error

pkg/sign/certificate.go
```diff
@@ -133,17 +135,31 @@ func (f *Fulcio) GetCertificate(keypair Keypair, identityToken string) ([]byte,
...
+			// Not a HTTP 5XX error, don't retry
+			break
+		}
+
+		attempts++
+		delay := time.Duration(math.Pow(2, float64(attempts)))
```
Hayden-IO, line 161: Small thing, I would consider starting with 2^0=1 second, so attempts - 1, otherwise 2 seconds adds a lot of delay initially.

FWIW, I'd consider using https://github.com/hashicorp/go-retryablehttp which implements this already. This also makes the backoff function configurable.

codysoyland: https://github.com/cenkalti/backoff is another library I've used for this type of thing.

I'd also add context to these calls.

steiza: I had thought about https://github.com/hashicorp/go-retryablehttp (I hadn't seen https://github.com/cenkalti/backoff), but I was trying to balance the here we have access to `http.Client`, but we don't in `timestamping.go`, so we have to do something else there anyways.

I have been meaning to learn about Go channels, and adding context was a great excuse to do so!

steiza: Oh, I also bumped `attempts++` down so the first sleep is shorter.

Hayden-IO (approved):

Hayden-IO:
I'd like to see a test for this as well, though that can be a part of later e2e testing.

steiza:
Yeah, I really need to start working on testing. Although testing retry behavior is going to be easier with mocks than e2e...

PR #186, opened 2024-05-22
Add the ability to verify signed bundle against trusted root

steiza:
#### Summary

As we work towards releasing signing support (https://github.com/sigstore/sigstore-go/issues/136), we want the option to verify that the bundle we signed is valid before returning it to the user.

At first I thought about requiring a trusted root, but I came up with at least two examples where that wasn't feasible:
1. Staging doesn't include a timestamp authority that is reachable from the general internet (arguably this could be fixed)
2. Conformance tests that use detached materials would require a custom verification policy, at which point you might as well use `verify` directly

#### Release Note

#### Documentation

N/A


Hayden-IO:
> Staging doesn't include a timestamp authority that is reachable from the general internet (arguably this could be fixed)

I didn't catch that we were using the [sigstore/timestamp-authority](https://github.com/sigstore/timestamp-authority/blob/main/pkg/client/timestamp_client.go) client in the [signer](https://github.com/sigstore/sigstore-go/blob/f10ccede7a7d21baced945ab6b697418123cf9de/pkg/sign/timestamping.go#L63). This limitation of this client is that the API of the TSA must match the one defined in Sigstore's implementation which is not standardized. The only standardization is what the API must accept and produce. This is why in Cosign, we implemented [a client](https://github.com/sigstore/cosign/blob/main/internal/pkg/cosign/tsa/client/client.go) where you must pass the API path rather than just the root URL, which means that we can support any TSA. We can upstream the [Cosign TSA client](https://github.com/sigstore/cosign/blob/main/internal/pkg/cosign/tsa/client/client.go) to sigstore/timestamp-authority as well.

If we do this, then you can use https://freetsa.org/index_en.php for staging.

Edit: Concretely, all that needs to change is that when you specify `BaseURL` when [initializing the TSA client](https://github.com/sigstore/sigstore-go/blob/f10ccede7a7d21baced945ab6b697418123cf9de/examples/signing/main.go#L96), you would just provide the entire API, e.g. `https://timestamp.githubapp.com/api/v1/timestamp`, which is a reasonable tradeoff to support all TSAs imo.

Edit edit: I also realized that doesn't exactly answer your question, so in addition to this change, we'd need to append the custom timestamp authority to the trust material.

Edit edit edit: I keep finding more things :) In the signing client config, we also define a TSA URL as its whole path - https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_trustroot.proto#L176-L178. I would suggest we make this change (in another PR) to align here and with the other clients.

examples/signing/staging_root.json
Hayden-IO, line 1: Should we move this into `pkg/tuf/repository`? I think it would be reasonable to ship sigstore-go with support for both the prod and staging public good instances, but I'm not strongly opinionated on this.

pkg/sign/signer.go
```diff
@@ -87,7 +95,7 @@ func Bundle(content Content, keypair Keypair, opts BundleOptions) (*protobundle.
 			Bytes: pubKeyBytes,
 		})
 
-		// TODO: do verification of Fulcio certificate
+		verifierOptions = append(verifierOptions, verify.WithSignedCertificateTimestamps(1))
```
Hayden-IO: The Fulcio instance isn't guaranteed to be configured to write to a CT log. We should condition this on the SCT extension (`asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 11129, 2, 4, 2}`, from https://github.com/sigstore/fulcio/blob/e813f854fe72d22fa9a9af8961dcda234d8d8f6f/pkg/ca/baseca/baseca.go#L37C22-L37C77) is set in the certificate. Marshal `pubKeyBytes` to a `x509.Certificate`, and:

```
for _, ext := range cert.Extensions {
    if reflect.DeepEqual(ext.Id, asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 11129, 2, 4, 2}) {
        verifierOptions = append(verifierOptions, verify.WithSignedCertificateTimestamps(1))
        break
    }
}
```

kommendorkapten: I think we even would be fine to ignore the SCT check as part of the signing workflow. 

The reason for verifying the signature after signing is to 1) verify that the signature is actually correct and 2) that we signed with the expected identity.

Verifying the SCT is a policy decision for the verifier IMO, so we can ignore that here.

Edit: conditioning SCT verification on the existence of an SCT extension is fine too of course, as that allows us to abort if the SCT signature is invalid, but it's not as crucial verify IMO, the CA should have done that for us.

examples/signing/staging_root.json
steiza: Sounds good to me! This simplifies things quite a bit both here and in the conformance tests.

pkg/sign/signer.go
steiza: Yeah, I think it makes sense to leave off the SCT check. We'll still make sure the Fulcio certificate verifies with what's in the trusted root.

kommendorkapten (approved):
LGTM

PR #187, opened 2024-05-23
Update timestamping to use full URL

steiza:
#### Summary

As requested by https://github.com/sigstore/sigstore-go/pull/186#issuecomment-2125782482.

By supplying the full URL, in the future we can add freetsa.org to the staging Sigstore TUF root, which would then let us verify bundles that are produced by `examples/signing/main.go` with `-tsa`.

#### Release Note

NONE - signing is not released yet

#### Documentation

N/A

### DIFF
--- a/examples/signing/main.go
+++ b/examples/signing/main.go
@@ -93,7 +93,7 @@ func main() {
 
 	if *tsa {
 		tsaOpts := &sign.TimestampAuthorityOptions{
-			BaseURL:        "https://timestamp.githubapp.com",
+			URL:            "https://timestamp.githubapp.com/api/v1/timestamp",
 			Timeout:        time.Duration(30 * time.Second),
 			Retries:        1,
 			LibraryVersion: Version,
--- a/pkg/sign/timestamping.go
+++ b/pkg/sign/timestamping.go
@@ -30,7 +30,7 @@ import (
 
 type TimestampAuthorityOptions struct {
 	// URL of Timestamp Authority instance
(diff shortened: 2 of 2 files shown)


codysoyland (approved):

Hayden-IO (commented):
Have you tested this locally? I don’t think this works, the client is specific to the Sigstore TSA API, and when you call “GetTinestampResponse”, it will append the API path again. You would need a client like what I linked as well.

steiza:
Ah, I misunderstood your comment:

> Edit: Concretely, all that needs to change is that when you specify BaseURL when [initializing the TSA client](https://github.com/sigstore/sigstore-go/blob/f10ccede7a7d21baced945ab6b697418123cf9de/examples/signing/main.go#L96), you would just provide the entire API, e.g. https://timestamp.githubapp.com/api/v1/timestamp, which is a reasonable tradeoff to support all TSAs imo.

It does work, but because https://github.com/sigstore/timestamp-authority/blob/main/pkg/client/timestamp_client.go#L27-L33 takes the provided URL, calls `url.Parse`, and just uses `url.Host`.

So we want to use a client that actually uses the entire URL, not just the host. Here's a few options for how we could do that:

- Instead of calling `GetTimestampClient()`, we could make our own wrapper around `sigstore/timestamp-authority/pkg/generated/client` that uses the entire URL
- We could upstream the cosign TSA client into sigstore/timestamp-authority and then use that
- ... other ideas welcome!

Issue #53, opened 2023-12-19
Add unit tests for online log verification and inclusion proofs


Hayden-IO:
**Description**

The mock Sigstore data generates a Rekor log entry with a SignedEntryTimestamp but no inclusion proof (See [`generateTlogEntry`](https://github.com/sigstore/sigstore-go/blob/9a59c7f51a05d2305a8ba4ab9343416aaaeb16d6/pkg/testing/ca/ca.go#L248)).

Additionally, we don't test the online verification path since we don't have a fake Rekor client. Should be straightforward to add with something like https://github.com/sigstore/rekor-monitor/blob/main/pkg/rekor/mock/mock_rekor_client.go

Hayden-IO:
Additionally, we should either add unit tests for https://github.com/sigstore/sigstore-go/blob/main/pkg/tlog/entry.go or confirm that the unit/integration tests for https://github.com/sigstore/sigstore-go/blob/main/pkg/verify/tlog.go cover entry.go

cmurphy:
@vishal-chdhry are you still interested in [taking this up](https://github.com/sigstore/sigstore-go/pull/188#issuecomment-2149435433) or should someone else have a go?

vishal-chdhry:
@cmurphy sure I can take this up, I forgot about it

vishal-chdhry:
\assign

PR #188, opened 2024-05-25
feat: use GetLogEntryByIndex to query rekor

### LINKED ISSUE #52: Use GetLogEntryByIndex when querying Rekor
**Description**

Relevant code: https://github.com/sigstore/sigstore-go/blob/9a59c7f51a05d2305a8ba4ab9343416aaaeb16d6/pkg/verify/tlog.go#L135-L138

SearchLogQuery is used when querying Rekor to get an entry, which returns a list of entries matching the provided index. This will return a single entry since we are requesting a single index, but we can instead use GetLogEntryByIndex which will always return a single log entry.

(This is a low priority improvement, not a functional one)

vishal-chdhry:
#### Summary

Closes: #52

Replaces SearchLogQuery with GetLogEntryByIndex which always returns a single log entry.

#### TODO
Add unit tests
#### Release Note
NONE

#### Documentation
NONE

### DIFF
--- a/pkg/verify/tlog.go
+++ b/pkg/verify/tlog.go
@@ -26,7 +26,6 @@ import (
 	rekorClient "github.com/sigstore/rekor/pkg/client"
 	rekorGeneratedClient "github.com/sigstore/rekor/pkg/generated/client"
 	rekorEntries "github.com/sigstore/rekor/pkg/generated/client/entries"
-	rekorModels "github.com/sigstore/rekor/pkg/generated/models"
 	rekorVerify "github.com/sigstore/rekor/pkg/verify"
 	"github.com/sigstore/sigstore/pkg/signature"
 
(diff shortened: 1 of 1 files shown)


vishal-chdhry:
@steiza I could not find any unit test for online set to `true`, any tips on how should I test it?

steiza (dismissed):
Yeah, testing could definitely be easier. Until we fix #53, I think the best way to test this is to get a bundle and run a verification manually.

There's a bundle in `examples/bundle-provenance.json`. You can also get bundles from many npm packages with something like:

```
$ curl https://registry.npmjs.org/-/npm/v1/attestations/semver@7.6.1 | jq ".attestations[0]"
```

Once you have the bundle, you can inspect the certificate to find the SAN and the issuer (`https://token.actions.githubusercontent.com` for GitHub Actions), and then attempt to verify the bundle with online tlog lookup via:

```
sigstore-go$ go run cmd/sigstore-go/main.go -expectedSAN "https://github.com/sigstore/sigstore-js/.github/workflows/release.yml@refs/heads/main" -expectedIssuer "https://token.actions.githubusercontent.com" -onlineTlog examples/bundle-provenance.json
...
Verification successful!
```

LGTM!

vishal-chdhry:
Okay, I will look into #53 and also add a test for this with that

Hayden-IO (dismissed):
Thanks for the change!

Hayden-IO:
Before I merge, to confirm, was this tested locally?

vishal-chdhry:
tested it out with the example provided by @steiza and it passed
```
go run cmd/sigstore-go/main.go -expectedSAN "https://github.com/sigstore/sigstore-js/.github/workflows/release.yml@refs/heads/main" -expectedIssuer "https://token.actions.githubusercontent.com" -onlineTlog examples/bundle-provenance.json
```

But I think we should add tests for this. Since this is not functional, we can pause it, I will look into #53 in the weekend and add tests for this too

Hayden-IO:
Thank you!

pkg/verify/tlog.go
```diff
@@ -148,7 +144,7 @@ func VerifyArtifactTransparencyLog(entity SignedEntity, trustedMaterial root.Tru
 				return nil, errors.New("too many log entries returned")
```
Hayden-IO: @vishal-chdhry can you remove this error condition, as it's no longer relevant?

vishal-chdhry:
@haydentherapper done

Hayden-IO (approved):

PR #189, opened 2024-05-26
feat: add validation of required fields in the bundle

### LINKED ISSUE #63: Additional verification of required fields in the bundle
**Description**

We currently check for an inclusion proof or promise in a [validation function](https://github.com/sigstore/sigstore-go/blob/f7518a68b37b7887b69e1607b9ddb668ddec117d/pkg/bundle/bundle.go#L74-L94) when constructing the bundle.

Further verification for required fields would be ideal. Using proto reflection, we could use the field annotation `field_behavior` to determine if the field has been annotated as required.

vishal-chdhry:
#### Summary
Ref #63

Sigstore bundle [protobuf spec](https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_bundle.proto) has `field_behavior` annotation which is used  to describe field behavior, such as a field being required or immutable as described in [AIP-203](https://google.aip.dev/203). 

This PR adds verification of top-level required fields in the bundle.

#### TODO

Fix validation of `oneof` fields

#### Release Note
```
Add validation of required fields in the bundle
```


pkg/bundle/bundle_test.go
```diff
@@ -124,3 +126,62 @@ func TestMediaTypeString(t *testing.T) {
...
+							PublicKey: &v1.PublicKeyIdentifier{},
+						},
+						TimestampVerificationData: &protobundle.TimestampVerificationData{},
+					},
+				},
+			},
+			errMsg:  "",
+			wantErr: false,
+		},
+	}
```
vishal-chdhry: This test case fails because `public_key` is present and `x509_certificate_chain` is not present, when only one has to be present

Hayden-IO:
Overall LGTM, pending figuring out the issue with required one_of fields. Thanks for adding this!

vishal-chdhry:
updated the PR @haydentherapper

kommendorkapten (dismissed):

Hayden-IO (dismissed):
This is a good start, though the linked issue isn't completed by this. The library you were using originally is what we need to do verification of nested fields, otherwise we're just validating top level fields.

I'm fine with merging this as-is, I've changed the PR description to not close the issue.

vishal-chdhry:
> The library you were using originally is what we need to do verification of nested fields, otherwise we're just validating top level fields.

Would the better solution be to use the library after it has better support for oneof fields (The AIP does not state anything specific about how to handle oneof so I am not sure if we can get something added to the library)
or, 
Should I just add validation for all the [imported fields](https://github.com/sigstore/protobuf-specs/blob/3d01ad8fa60bdd0152bea2963b4d35282c2e263a/protos/sigstore_bundle.proto#L21-L23)?

Hayden-IO:
I think the library would be the best path forward. Maybe we can post an issue there to see if they have suggestions for handling one-ofs in a way that is AIP compliant?

vishal-chdhry:
logged an issue for this in the library: https://github.com/einride/aip-go/issues/300

Hayden-IO:
@vishal-chdhry Do you want to rebase for the tests, and we'll merge this as is?

Hayden-IO (approved):

PR #191, opened 2024-05-29
Greatly increase test coverage for pkg/sign

steiza:
#### Summary

As we work towards signing support (#136) we want to ensure there's good test coverage of the code, in addition to conformance tests. This increases the built-in test coverage (e.g. not conformance tests) from ~23% to ~65%.

#### Release Note

#### Documentation

N/A

### DIFF
--- a/pkg/sign/certificate.go
+++ b/pkg/sign/certificate.go
@@ -37,7 +37,7 @@ type Fulcio struct {
 type FulcioOptions struct {
 	// URL of Fulcio instance
 	BaseURL string
-	// Optional timeout for network requests (default 30s)
+	// Optional timeout for network requests (default 30s; use negative value for no timeout)
 	Timeout time.Duration
 	// Optional number of times to retry on HTTP 5XX
 	Retries uint
@@ -80,12 +80,16 @@ type chain struct {
 func NewFulcio(opts *FulcioOptions) *Fulcio {
 	fulcio := &Fulcio{options: opts}
 	fulcio.client = &http.Client{
-		Timeout:   opts.Timeout,
 		Transport: opts.Transport,
 	}
-	if fulcio.client.Timeout == 0 {
-		fulcio.client.Timeout = 30 * time.Second
+
+	if opts.Timeout >= 0 {
+		if opts.Timeout == 0 {
+			opts.Timeout = 30 * time.Second
+		}
+		fulcio.client.Timeout = opts.Timeout
 	}
+
 	return fulcio
 }
 
--- a/pkg/sign/certificate.go
+++ b/pkg/sign/certificate.go
@@ -29,6 +29,10 @@ import (
 	"time"
 )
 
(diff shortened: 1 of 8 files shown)


pkg/sign/certificate.go
```diff
@@ -42,6 +46,8 @@ type FulcioOptions struct {
 	Retries uint
 	// Optional version string for user agent
 	LibraryVersion string
+	// Optional HTTP client (for dependency injection)
+	Client HTTPClient
```
codysoyland: Nit: I suggest using [`http.RoundTripper`](https://pkg.go.dev/net/http#RoundTripper) for dependency injection.

pkg/sign/certificate_test.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/stretchr/testify/assert"
+
+	"github.com/sigstore/sigstore-go/pkg/testing/ca"
+)
+
+var virtualSigstore *ca.VirtualSigstore
+
+func getFulcioResponse() (*http.Response, error) {
+	var err error
+	if virtualSigstore == nil {
```
codysoyland, line 37: Mutating a package level var can result in bugs if we go add another test that uses this same var (as tests can run in parallel), so I would guard this check with [`sync.Once`](https://pkg.go.dev/sync#Once.Do), which is like a single-use mutex.

pkg/sign/certificate.go
Hayden-IO: I was reading a bit about the idiomatic way to do testing in Golang, and a common [suggestion](https://medium.com/zus-health/mocking-outbound-http-requests-in-go-youre-probably-doing-it-wrong-60373a38d2aa#:~:text=There%20is%20an%20easier%2C%20idiomatic%20way) was https://pkg.go.dev/net/http/httptest

pkg/sign/certificate.go
```diff
@@ -75,6 +81,14 @@ type chain struct {
 }
 
 func NewFulcio(opts *FulcioOptions) *Fulcio {
+	if opts.Client == nil {
+		client := &http.Client{}
+		if opts.Timeout != 0 {
```
kommendorkapten: If no timeout is provided I would still configure a timeout (30s?), as the default client blocks "forever" and that can lead to very unexpected behaviours.

pkg/sign/certificate.go
codysoyland: Interesting, I've used the pkg httptest for testing handlers before, but not outbound requests. I think for this usecase, I still prefer replacing the transport with a custom `RoundTripper` as it seems simpler and doesn't open a port in tests.

codysoyland (dismissed):
nice work!

Hayden-IO (dismissed):

Hayden-IO (approved):

codysoyland (approved):

PR #195, opened 2024-05-31
Add certificate interface to allow for non-Fulcio certificate authorities

steiza:
#### Summary

We are working our way down the punch list to enable signing (#136). We want the interface to be pretty stable before then, so we're making an interface for certificate authorities generally, in case someone wants to use something other than Fulcio.

#### Release Note

#### Documentation

N/A

### DIFF
--- a/pkg/sign/certificate.go
+++ b/pkg/sign/certificate.go
@@ -29,8 +29,13 @@ import (
 	"time"
 )
 
-type Certificate interface {
-	GetCertificate(context.Context, Keypair, string) ([]byte, error)
+type CertificateAuthorityOptions struct {
+	// Optional OIDC JWT to send to certificate authority; required for Fulcio
+	IDToken string
+}
+
+type CertificateAuthority interface {
+	GetCertificate(context.Context, Keypair, *CertificateAuthorityOptions) ([]byte, error)
 }
 
 type Fulcio struct {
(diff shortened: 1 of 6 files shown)


pkg/sign/certificate.go
```diff
@@ -29,6 +29,10 @@ import (
 	"time"
 )
 
+type Certificate interface {
```
Hayden-IO: Can we call this `CertificateAuthority`?

pkg/sign/certificate.go
```diff
@@ -29,6 +29,10 @@ import (
...
+	GetCertificate(context.Context, Keypair, string) ([]byte, error)
```
Hayden-IO: can you document what the string is supposed to be?

Hayden-IO (changes requested):

pkg/sign/certificate.go
Hayden-IO: We might want to make `string` an `CertificateRequestOptions` struct so that it can be extended in the future. A `KeyPair` will always be needed, but an identity token may not.

Hayden-IO (dismissed):

Hayden-IO:
Looks great, thanks!

kommendorkapten (dismissed):
Nice!

pkg/sign/certificate.go
kommendorkapten: Thinking more about this. For the Fulcio use-case `CertificateAuthority` makes sense. But consider a scenario where an existing certificate is provided (i.e it's already signed by a CA before artifact signing begins). In this case the name `CertificateAuthority` is a bit misleading. How about changing this to `CertificateProvider` or something similar?

Hayden-IO (approved):

PR #197, opened 2024-06-02
feat: add support for additional transparency log key types

### LINKED ISSUE #65: Support for additional transparency log key types
**Description**

Code: https://github.com/sigstore/sigstore-go/blob/f7518a68b37b7887b69e1607b9ddb668ddec117d/pkg/root/trusted_root.go#L137-L146

The public instance uses ECDSA P256, but private instances can use RSA, other ECDSA curves, and ed25519. Should be a straightforward change, as `ParsePKIXPublicKey` can parse these key types. Support for PKCS1 encoded keys can be added with `ParsePKCS1PublicKey`.

vishal-chdhry:
#### Summary
Closes:  #65

Adds support for transparency log keys with RSA, other ECDSA curves, and ed25519 

#### Release Note

```
Adds support for transparency log keys with RSA, other ECDSA curves, and ed25519 
```

### DIFF
--- a/pkg/root/trusted_root.go
+++ b/pkg/root/trusted_root.go
@@ -145,7 +145,10 @@ func ParseTransparencyLogs(tlogs []*prototrustroot.TransparencyLogInstance) (tra
 		}
 
 		switch tlog.GetPublicKey().GetKeyDetails() {
-		case protocommon.PublicKeyDetails_PKIX_ECDSA_P256_SHA_256:
+		case protocommon.PublicKeyDetails_PKIX_ECDSA_P256_SHA_256,
+			protocommon.PublicKeyDetails_PKIX_ECDSA_P384_SHA_384,
+			protocommon.PublicKeyDetails_PKIX_ECDSA_P521_SHA_512,
+			protocommon.PublicKeyDetails_PKIX_ECDSA_P256_HMAC_SHA_256:
 			key, err := x509.ParsePKIXPublicKey(tlog.GetPublicKey().GetRawBytes())
 			if err != nil {
 				return nil, err
@@ -180,9 +183,9 @@ func ParseTransparencyLogs(tlogs []*prototrustroot.TransparencyLogInstance) (tra
 			if err != nil {
 				return nil, err
 			}
-			var edKey *ed25519.PublicKey
+			var edKey ed25519.PublicKey
 			var ok bool
-			if edKey, ok = key.(*ed25519.PublicKey); !ok {
+			if edKey, ok = key.(ed25519.PublicKey); !ok {
 				return nil, fmt.Errorf("tlog public key is not RSA: %s", tlog.GetPublicKey().GetKeyDetails())
 			}
 			tlogEntry.PublicKey = edKey
--- a/pkg/root/trusted_root.go
+++ b/pkg/root/trusted_root.go
@@ -139,9 +139,10 @@ func ParseTransparencyLogs(tlogs []*prototrustroot.TransparencyLogInstance) (tra
 		}
 
 		tlogEntry := &TransparencyLog{
-			BaseURL:  tlog.GetBaseUrl(),
-			ID:       tlog.GetLogId().GetKeyId(),
-			HashFunc: hashFunc,
+			BaseURL:           tlog.GetBaseUrl(),
+			ID:                tlog.GetLogId().GetKeyId(),
+			HashFunc:          hashFunc,
+			SignatureHashFunc: crypto.SHA256,
 		}
 
 		switch tlog.GetPublicKey().GetKeyDetails() {
@@ -159,7 +160,6 @@ func ParseTransparencyLogs(tlogs []*prototrustroot.TransparencyLogInstance) (tra
 				return nil, fmt.Errorf("tlog public key is not ECDSA: %s", tlog.GetPublicKey().GetKeyDetails())
 			}
 			tlogEntry.PublicKey = ecKey
-			tlogEntry.SignatureHashFunc = crypto.SHA256
 		// This key format has public key in PKIX RSA format and PKCS1#1v1.5 or RSASSA-PSS signature
 		case protocommon.PublicKeyDetails_PKIX_RSA_PKCS1V15_2048_SHA256,
 			protocommon.PublicKeyDetails_PKIX_RSA_PKCS1V15_3072_SHA256,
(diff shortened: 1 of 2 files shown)


pkg/root/trusted_root.go
```diff
@@ -145,32 +153,54 @@ func ParseTransparencyLogs(tlogs []*prototrustroot.TransparencyLogInstance) (tra
...
+			if err != nil {
+				return nil, err
+			}
+			var edKey *ed25519.PublicKey
+			var ok bool
+			if edKey, ok = key.(*ed25519.PublicKey); !ok {
+				return nil, fmt.Errorf("tlog public key is not RSA: %s", tlog.GetPublicKey().GetKeyDetails())
+			}
+			tlogEntry.PublicKey = edKey
+			tlogEntry.SignatureHashFunc = crypto.SHA512 // see RFC8032
```
vishal-chdhry: Is `hashFunc` related to `SignatureHashFunc`? ED25519 uses SHA512 signature scheme and only SHA256 is supported as a hash function: https://github.com/sigstore/sigstore-go/blob/3a6a109900dfe00c55d51d18556c37859fbae3b8/pkg/root/trusted_root.go#L132-L137

pkg/root/trusted_root.go
```diff
@@ -136,6 +138,12 @@ func ParseTransparencyLogs(tlogs []*prototrustroot.TransparencyLogInstance) (tra
...
 		}
 
+		tlogEntry := &TransparencyLog{
+			BaseURL:  tlog.GetBaseUrl(),
+			ID:       tlog.GetLogId().GetKeyId(),
+			HashFunc: hashFunc,
+		}
+
 		switch tlog.GetPublicKey().GetKeyDetails() {
 		case protocommon.PublicKeyDetails_PKIX_ECDSA_P256_SHA_256:
```
Hayden-IO: Can we support the other ECDSA curves as well?

pkg/root/trusted_root.go
Hayden-IO: No, these are different. hashFunc is the digest used for hashing the leaves in the merkle tree, SignatureHashFunc would be the digest used for signing SETs or checkpoints.

Hayden-IO (changes requested):
Thanks! Can you add tests for this?

vishal-chdhry:
@haydentherapper added the changes, please let me know if this is okay.

sorry for the delay on this, I was working on other things (btw, we also added support for sigstore bundle verification in Kyverno, it will be a part of the next release)

pkg/root/trusted_root.go
```diff
@@ -136,41 +138,69 @@ func ParseTransparencyLogs(tlogs []*prototrustroot.TransparencyLogInstance) (tra
...
+			HashFunc:          hashFunc,
+			SignatureHashFunc: crypto.SHA256,
+		}
+
 		switch tlog.GetPublicKey().GetKeyDetails() {
-		case protocommon.PublicKeyDetails_PKIX_ECDSA_P256_SHA_256:
+		case protocommon.PublicKeyDetails_PKIX_ECDSA_P256_SHA_256,
+			protocommon.PublicKeyDetails_PKIX_ECDSA_P384_SHA_384,
+			protocommon.PublicKeyDetails_PKIX_ECDSA_P521_SHA_512,
+			protocommon.PublicKeyDetails_PKIX_ECDSA_P256_HMAC_SHA_256: //nolint:staticcheck
```
Hayden-IO: We don't need to support `PublicKeyDetails_PKIX_ECDSA_P256_HMAC_SHA_256`, we've marked this as deprecated.

pkg/root/trusted_root.go
```diff
@@ -136,41 +138,69 @@ func ParseTransparencyLogs(tlogs []*prototrustroot.TransparencyLogInstance) (tra
 			return nil, fmt.Errorf("unsupported hash function for the tlog")
 		}
 
+		tlogEntry := &TransparencyLog{
+			BaseURL:           tlog.GetBaseUrl(),
+			ID:                tlog.GetLogId().GetKeyId(),
+			HashFunc:          hashFunc,
+			SignatureHashFunc: crypto.SHA256,
```
Hayden-IO, line 146: The signature hash algorithm will differ depending on the signature scheme. I'd recommend a function like https://github.com/slsa-framework/slsa-verifier/blob/d96b9777090694fa5096ee1b9c710a46b5a66f5e/cli/slsa-verifier/verify/verify_vsa.go#L95-L117 to handle the mapping

tldr - RSA can always be SHA256, ECDSA P256 and P384 should be SHA256, P521 SHA512, and ed25519 is SHA512.

pkg/root/trusted_root.go
```diff
@@ -136,41 +138,69 @@ func ParseTransparencyLogs(tlogs []*prototrustroot.TransparencyLogInstance) (tra
...
 			if ecKey, ok = key.(*ecdsa.PublicKey); !ok {
-				return nil, fmt.Errorf("tlog public key is not ECDSA P256")
+				return nil, fmt.Errorf("tlog public key is not ECDSA: %s", tlog.GetPublicKey().GetKeyDetails())
+			}
+			tlogEntry.PublicKey = ecKey
+		// This key format has public key in PKIX RSA format and PKCS1#1v1.5 or RSASSA-PSS signature
+		case protocommon.PublicKeyDetails_PKIX_RSA_PKCS1V15_2048_SHA256,
+			protocommon.PublicKeyDetails_PKIX_RSA_PKCS1V15_3072_SHA256,
+			protocommon.PublicKeyDetails_PKIX_RSA_PKCS1V15_4096_SHA256,
+			protocommon.PublicKeyDetails_PKIX_RSA_PSS_2048_SHA256,
```
Hayden-IO: We can skip support for `PSS`, that would require also changing how we call `LoadVerifier`, which would require conditionally adding an option to use PSS over PKCS1v1.5. Given loading this verifier is in another package, it'd be a bit too much refactoring to try to connect these for not much gain.

pkg/root/trusted_root.go
```diff
@@ -136,41 +138,69 @@ func ParseTransparencyLogs(tlogs []*prototrustroot.TransparencyLogInstance) (tra
...
-				HashFunc:          hashFunc,
-				PublicKey:         ecKey,
-				SignatureHashFunc: crypto.SHA256,
+			var rsaKey *rsa.PublicKey
+			var ok bool
+			if rsaKey, ok = key.(*rsa.PublicKey); !ok {
+				return nil, fmt.Errorf("tlog public key is not RSA: %s", tlog.GetPublicKey().GetKeyDetails())
 			}
+			tlogEntry.PublicKey = rsaKey
+		case protocommon.PublicKeyDetails_PKIX_ED25519, protocommon.PublicKeyDetails_PKIX_ED25519_PH: //nolint:staticcheck
```
Hayden-IO: Same here, we can skip `PKIX_ED25519_PH`, the prehash variant of ed25519 isn't needed, and it would require making changes to `LoadVerifier`.

Hayden-IO (changes requested):
Thanks for making these updates! A few comments on reducing the number of supported algorithms and updating the digest.

pkg/root/trusted_root.go
vishal-chdhry: I have removed PKIX-PSS but I havent removed PKIX-PKCS1v1.5

Hayden-IO (approved):
Thank you!

Issue #196, opened 2024-05-31
Support TSAs other than sigstore/timestamp-authority


Hayden-IO:
RFC3161 does not define a specification for the API path of a TSA, only standardizing the input and output as timestamp request and response structures. The Sigstore TSA implementation defines the path to request timestamps as `api/v1/timestamp`.

As discussed in https://github.com/sigstore/sigstore-go/pull/187, the current timestamp authority client, which comes from [generated code in sigstore/timestamp-authority](https://github.com/sigstore/timestamp-authority/blob/main/pkg/generated/client/timestamp_authority_client.go), only supports the Sigstore TSA implementation. When a URL is provided to the client, it is truncated to only the URL host, and then the API path above is appended.

In order to support arbitrary TSAs, we need to use a custom client that a) uses the provide URL with both the host and path, b) takes in a timestamp request, and c) parses a returned timestamp response.

We have implemented this already in Cosign as part of https://github.com/sigstore/cosign/pull/2708:

* https://github.com/sigstore/cosign/blob/main/internal/pkg/cosign/tsa/client/client.go for the client
* https://github.com/sigstore/cosign/blob/main/internal/pkg/cosign/tsa/mock/mock_tsa_client.go for a mocked client for testing
* https://github.com/sigstore/cosign/blob/main/internal/pkg/cosign/tsa/signer.go#L38-L45 as an example of this being used

I would recommend we upstream this client to sigstore/timestamp-authority or duplicate it in sigstore/sigstore-go.

steiza:
This was fixed with https://github.com/sigstore/sigstore-go/pull/202

PR #202, opened 2024-06-03
Support other TSAs for signing

steiza:
#### Summary

For #196. This is a light wrapper around github.com/digitorus/timestamp, and additionally supports retries.

Now you can put things like `https://freetsa.org/tsr` on https://github.com/sigstore/sigstore-go/blob/main/examples/signing/main.go#L120 and it still works!

#### Release Note

NONE - we aren't ready to release signing yet (although maybe this is the last signing change before we do?)

#### Documentation

N/A

### DIFF
--- a/pkg/sign/timestamping.go
+++ b/pkg/sign/timestamping.go
@@ -19,18 +19,14 @@ import (
 	"context"
 	"crypto"
 	"crypto/sha256"
+	"fmt"
 	"io"
 	"math"
+	"net/http"
 	"time"
 
 	"github.com/digitorus/timestamp"
-	tsaclient "github.com/sigstore/timestamp-authority/pkg/client"
-	tsagenclient "github.com/sigstore/timestamp-authority/pkg/generated/client/timestamp"
 )
-
-type TSAClient interface {
-	GetTimestampResponse(params *tsagenclient.GetTimestampResponseParams, writer io.Writer, opts ...tsagenclient.ClientOption) (*tsagenclient.GetTimestampResponseCreated, error)
-}
 
 type TimestampAuthorityOptions struct {
 	// URL of Timestamp Authority instance
@@ -41,16 +37,29 @@ type TimestampAuthorityOptions struct {
 	Retries uint
 	// Optional version string for user agent
 	LibraryVersion string
-	// Optional client (for dependency injection)
-	Client TSAClient
+	// Optional Transport (for dependency injection)
+	Transport http.RoundTripper
 }
 
 type TimestampAuthority struct {
 	options *TimestampAuthorityOptions
+	client  *http.Client
 }
 
 func NewTimestampAuthority(opts *TimestampAuthorityOptions) *TimestampAuthority {
-	return &TimestampAuthority{options: opts}
+	ta := &TimestampAuthority{options: opts}
+	ta.client = &http.Client{
+		Transport: opts.Transport,
+	}
+
+	if opts.Timeout >= 0 {
+		if opts.Timeout == 0 {
+			opts.Timeout = 30 * time.Second
+		}
+		ta.client.Timeout = opts.Timeout
+	}
+
+	return ta
 }
 
 func (ta *TimestampAuthority) GetTimestamp(ctx context.Context, signature []byte) ([]byte, error) {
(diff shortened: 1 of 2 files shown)


pkg/sign/timestamping.go
```diff
@@ -104,16 +107,21 @@ func (ta *TimestampAuthority) GetTimestamp(ctx context.Context, signature []byte
 		attempts++
 	}
 
+	body, err := io.ReadAll(response.Body)
```
kommendorkapten, line 109: Nit but I would move the consumption of the body below where we have verified that we got an expected response (200 or 201).

kommendorkapten (dismissed):
Nice one!

pkg/sign/timestamping.go
steiza: I'm undecided - if the error response contains a body with useful content, then we can return that in the error on line 115. I might leave this as-is.

pkg/sign/timestamping.go
```diff
@@ -41,23 +37,35 @@ type TimestampAuthorityOptions struct {
...
+	}
+
+	return ta
 }
 
 func (ta *TimestampAuthority) GetTimestamp(ctx context.Context, signature []byte) ([]byte, error) {
 	signatureHash := sha256.Sum256(signature)
 
 	req := &timestamp.Request{
-		Certificates:  true,
```
steiza, line 60: I went down quite the rabbit hole investigating this, but following sigstore-js' example I don't think we want to include the leaf certificate from the timestamp authority in the bundle.

Hayden-IO: It's been a bit since we wrote this, but according to https://github.com/sigstore/timestamp-authority/blob/main/pkg/verification/verify.go#L272-L280 and looking at https://github.com/digitorus/pkcs7/blob/3a137a8743524b3683ca4e11608d0dde37caee99/verify.go#L161, there is a requirement that the certificate is present in the signed timestamp struct to be verified. However, if you're using the sigstore/timestamp-authority verification library, we add in the provided leaf certificate to the struct [before verification](https://github.com/sigstore/timestamp-authority/blob/main/pkg/verification/verify.go#L278-L280) so it's not strictly required iirc.

What I don't recall is if setting this to false affects other values. Going down the rabbit hole, I believe no, that it's just the included certs - https://github.com/digitorus/pkcs7/blob/3a137a8743524b3683ca4e11608d0dde37caee99/sign.go#L194-L199 

Have you tested this locally to confirm that a signed timestamp generated here can be verified by both sigstore-go and sigstore-js? My preference would also be to leave out the cert to save space.

I think sigstore-python and sigstore-java are just starting to look at RFC3161 support, so we should let them know as well. This is one of those things that's likely to differ between rfc3161 implementations.

pkg/sign/timestamping.go
```diff
@@ -19,19 +19,15 @@ import (
...
-	tsaclient "github.com/sigstore/timestamp-authority/pkg/client"
-	tsagenclient "github.com/sigstore/timestamp-authority/pkg/generated/client/timestamp"
 )
 
-type TSAClient interface {
-	GetTimestampResponse(params *tsagenclient.GetTimestampResponseParams, writer io.Writer, opts ...tsagenclient.ClientOption) (*tsagenclient.GetTimestampResponseCreated, error)
-}
-
 type TimestampAuthorityOptions struct {
 	// URL of Timestamp Authority instance
```
Hayden-IO: Can you update the comment to say it should be the full URL?

Hayden-IO (commented):
Just one "comment" comment, thanks!

Long-term, I'd like to use the client we have in Cosign since it has support for mTLS as well, but this is good for now.

pkg/sign/timestamping.go
kommendorkapten: > What I don't recall is if setting this to false affects other values.

According to the RFC it should _only_ affect the inclusion of the certificate (which may contain more than one certificate) in the response.

kommendorkapten:
> Long-term, I'd like to use the client we have in Cosign since it has support for mTLS as well, but this is good for now.

Would it make sense to move that client to `sigstore/sigstore` and not have the client defined in Cosign?

edit: spelling errors

pkg/sign/timestamping.go
Hayden-IO: Perfect, that syncs up with what's in code as well.

pkg/sign/timestamping.go
steiza: I updated the comment - also we call this a `URL` instead of a `BaseURL` like we have for `RekorOptions` and `FulcioOptions`.

kommendorkapten (approved):
🚀

Hayden-IO (approved):

Hayden-IO:
> > Long-term, I'd like to use the client we have in Cosign since it has support for mTLS as well, but this is good for now.
> 
> Would it make sense to move that client to `sigstore/sigstore` and not have the client defined in Cosign?
> 
> edit: spelling errors

I was thinking moving it to the TSA repo, but yea, in either case it needs to be moved out of cosign.

PR #203, opened 2024-06-07
Update README and documentation for signing support

### LINKED ISSUE #136: Support Signing

steiza:
#### Summary

Closes #136 🎉 by finishing documentation, updating README, and fixing a bug uncovered while writing those things.

#### Release Note

* Supports signing; see docs/signing.md and examples/sigstore-go-signing/main.go for details

#### Documentation

N/A - I think we want to wait for cosign support to update https://docs.sigstore.dev/signing/overview/


pkg/sign/signer.go
```diff
@@ -45,7 +45,7 @@ type BundleOptions struct {
 	// Optional list of Rekor instances to get transparency log entry from.
 	//
 	// Supports hashedrekord and dsse entry types.
-	Rekors []*Rekor
+	Transparencies []Transparency
```
Hayden-IO: WDYT about TransparencyLog(s)? Or RekorLogs? I don’t think using Rekor is a problem, because this implementation only works with the Rekor API. 

I do prefer adding Log/Logs to the name over just Rekor/Rekors

kommendorkapten: TransparencyLogs is what I would prefer, it's used in a few other places too.

Hayden-IO (approved):
This looks great!

Issue #143, opened 2024-04-08
Set custom user agent for outbound requests


Hayden-IO:
**Description**

It would help when debugging traffic to distinguish between different Golang clients. We may need to make a change to go-tuf to pass the custom user agent.

steiza:
This was done in https://github.com/sigstore/sigstore-go/pull/216

PR #216, opened 2024-06-25
Add user agent to online Rekor requests and TUF requests.

steiza:
#### Summary

For https://github.com/sigstore/sigstore-go/issues/143.

One way to test this is to set up a fake TUF server locally with something like `nc -l 8000` and then attempt a verification with something like:

```bash
go run cmd/sigstore-go/main.go -artifact examples/sigstore-go-signing/hello_world.txt -tufRootURL http://127.0.0.1:8000 -expectedSANRegex ".+" -expectedIssuer "https://token.actions.githubusercontent.com" examples/bundle-provenance.json
```

Note that this verification won't succeed, this is just to see that the User-Agent is being set.

Depends on https://github.com/theupdateframework/go-tuf/pull/642.

#### Release Note

* Rekor verification and TUF client additionally send User-Agent header with requests

#### Documentation

N/A


steiza:
Note that some checks will fail until https://github.com/theupdateframework/go-tuf/pull/642 lands, which this depends on.

Hayden-IO (dismissed):

codysoyland:
The `LibraryVersion` value won't be set when this is used as a library (it must be set with LDFLAGS, so projects that import sigstore-go must set that value). Can we switch to the [debug/buildinfo](https://pkg.go.dev/debug/buildinfo) package for this?

codysoyland (changes requested):
See last comment

steiza:
> Can we switch to the [debug/buildinfo](https://pkg.go.dev/debug/buildinfo) package for this?

Sure! That turns out to have simplified things considerably.

Hayden-IO (approved):
oo nice, this cleans up well!

codysoyland (approved):
Nice! One small caveat with that `debug.ReadBuildInfo()` solution is that the version is not read from git tags when using `go build` to build the sigstore-go CLI (it outputs just `(devel)` in that case). However, when using it as a package, or when installing with `go install pkg@version`, it should contain the tagged version. I'm okay with that caveat personally.

PR #218, opened 2024-06-26
Improve error message when verifying certificate identity

### LINKED ISSUE #89: Improve error message for verifying certificate identify
**Description**

encounter verify failing with error `failed to verify certificate identity: no matching certificate identify`, when verifying the CertificateIdentity of the bundle. However, suggest that the error to be more clear something like 
```
SAN OIDSourceRepositoryURI "username" does not match expected pattern "userName"
```

codysoyland:
This modifies the API for verifying certificate identity (SAN and Extensions) to return error values instead of booleans, in order to provide a more helpful error message when the identity cannot be verified.

Fixes #89 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /sigstore-go
 /tufdata
 /conformance
+/pkg/tuf/testing.local.json
--- a/pkg/fulcio/certificate/summarize.go
+++ b/pkg/fulcio/certificate/summarize.go
@@ -17,6 +17,7 @@ package certificate
 import (
 	"crypto/x509"
 	"errors"
+	"fmt"
 	"reflect"
 )
 
@@ -34,7 +35,7 @@ const (
 )
 
 type SubjectAlternativeName struct {
-	Type  SubjectAlternativeNameType `json:"type,omitempty""`
+	Type  SubjectAlternativeNameType `json:"type,omitempty"`
 	Value string                     `json:"value,omitempty"`
 }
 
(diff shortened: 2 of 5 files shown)


steiza (dismissed):
LGTM!

Part of me worries very slightly that by telling people what value we're expecting when there's an error, they will just copy / paste the expected value without thinking about if it actually was a problem (like a typo confusion attack) or not.

But you could equally argue that this is more user-friendly and helps people discover what these expected values should be.

codysoyland:
Thanks for the review! Good to think about that scenario, but I think providing more info to the user is the right balance here.

I pushed a few more lines of test code that test the error Unwrap method.

Hayden-IO (approved):
thanks!

I was thinking more on Zach's comment and if this is an issue here. There's a class of attacks related to oracles, for example on authentication where you return a different error for an account that does not exist vs one where the password is wrong. Given that the metadata we're working with has integrity and authenticity but not confidentiality, I don't see this being an issue. 

I agree with the small risk of someone copy-pasting the "malicious" value, but hopefully (eventually) this would be a part of a verification workflow without user input where errors are unexpected and mostly just for debugging during development.

codysoyland:
> Given that the metadata we're working with has integrity and authenticity but not confidentiality, I don't see this being an issue.

@haydentherapper Good thoughts! The verification inputs should be outside of user control in any "enforcement" scenario, and as you said, the bundle contents are authenticated, so I think this is safe. There's been a lot of confusion with "no matching identity" during verification debugging, so this will hopefully improve UX/DX.

PR #226, opened 2024-07-10
Dsse one sig

### LINKED ISSUE #225: Improper verification of number of DSSE signatures
**Description**

During verification where a DSSE envelope is embedded in the bundle, sigstore-go accepts envelopes a signature count other than 1, which is in violation with the [protobuf-bundle spec](https://github.com/sigstore/protobuf-specs/blob/7acff8d6c718b44672e677a52e70205c6e0716b5/protos/sigstore_bundle.proto#L143). 

This does not allow for any integrity or authenticity threats, as counter signatures are required, and the verifier must provide a list of accepted verification materials. But this can cause DoS style attacks where a bundle is modified to contain an invalid signature first, as the [first signature](https://github.com/sigstore/sigstore-go/blob/main/pkg/bundle/signature_content.go#L100) is returned.

**Version**

All versions.

kommendorkapten:
#### Summary
Closes #225 

This is implemented by verifying that the envelope has _exactly_ one signature during the verification process. This is a bug-fix, as the protobuf spec for the bundle requires exactly one signature to be present in the envelope.

Note that I did not chose to modify the `Signature` [function](https://github.com/sigstore/sigstore-go/blob/main/pkg/bundle/signature_content.go#L100). This could be done too, but I think it is better to have the verification process be more strict, as it is a security feature.

#### Release Note
* Signature verification of bundles with DSSE envelopes now verifies that the envelope have exactly one signature.

#### Documentation
N/A


pkg/verify/signature.go
```diff
@@ -31,6 +31,8 @@ import (
 	"github.com/sigstore/sigstore/pkg/signature/options"
 )
 
+var ErrInvSigCount = errors.New("exactly one signature is required")
```
codysoyland: For consistency with other sentinel errors which use unabbreviated names, could you rename this `ErrInvalidSignatureCount`?

pkg/verify/dsse_test.go
```diff
@@ -0,0 +1,124 @@
...
+type envelopeContent struct {
+	EnvelopeContent
+	e *dsse.Envelope
+}
+
+func (e *envelopeContent) RawEnvelope() *dsse.Envelope {
+	return e.e
+}
+
+func TestVerifyEnvelope(t *testing.T) {
```
codysoyland: I would like to be explicit what the purpose of the test is, for example (or add a comment if you prefer):
```suggestion
func TestVerifyEnvelopeSignatureCount(t *testing.T) {
```

codysoyland (changes requested):

pkg/verify/signature.go
Hayden-IO: Can we also make this specific to DSSE as well, mentioning DSSE in the var name and error message?

kommendorkapten: Makes sense. Updated.

pkg/verify/dsse_test.go
kommendorkapten: Yeah, good call out. Name is too generic.

codysoyland (approved):

PR #228, opened 2024-07-10
Fix ProtobufBundle.MinVersion

### LINKED ISSUE #227: MinVersion not compatible with Bundle v0.3
**Description**

`ProtobufBundle.MinVersion` is not compatible with the new v0.3 Bundle MediaType which ends with `+json`.

This is used by the flag `-minBundleVersion` flag in the CLI.

codysoyland:
Fixes #227

### DIFF
--- a/pkg/bundle/bundle_test.go
+++ b/pkg/bundle/bundle_test.go
@@ -106,6 +106,39 @@ func Test_getBundleVersion(t *testing.T) {
 	}
 }
 
+func TestMinVersion(t *testing.T) {
+	t.Parallel()
+	for _, tc := range []struct {
+		name            string
+		mediaType       string
+		expectedVersion string
+		ret             bool
+	}{
+		{"old-format", "application/vnd.dev.sigstore.bundle+json;version=0.1", "v0.1", true},
+		{"old-format-unexpected", "application/vnd.dev.sigstore.bundle+json;version=0.1", "v0.2", false},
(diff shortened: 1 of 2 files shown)


pkg/bundle/bundle_test.go
```diff
@@ -106,6 +106,39 @@ func Test_getBundleVersion(t *testing.T) {
...
+	t.Parallel()
+	for _, tc := range []struct {
+		name            string
+		mediaType       string
+		expectedVersion string
+		ret             bool
+	}{
+		{"old-format", "application/vnd.dev.sigstore.bundle+json;version=0.1", "v0.1", true},
+		{"old-format-unexpected", "application/vnd.dev.sigstore.bundle+json;version=0.1", "v0.2", false},
+		{"old-format-without-v", "application/vnd.dev.sigstore.bundle+json;version=0.1", "0.1", true},
```
Hayden-IO, line 119: Do we want to support matching both with and without the preceding `v`, or should we standardize around one?

kommendorkapten: I would favour the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle) here, i.e what's implemented. 
As this is a simple helper function, requiring stricter formatting does not seem to aid in any correctness IMHO.

edit: grammer fix.

kommendorkapten (approved):
🚀

PR #229, opened 2024-07-12
Updates for SAN parsing

### LINKED ISSUE #55: Support additional GeneralNames in the SubjectAlternativeName
**Description**

See https://github.com/sigstore/sigstore-go/blob/main/pkg/fulcio/certificate/summarize.go#L64

- [ ] Support OtherName, used by Fulcio for user names in certificates (not currently in use in production)
- [ ] Support parsing additional GeneralNames like DNS, for BYO PKI use cases

### LINKED ISSUE #175: Proposal to remove Type for SubjectAlternativeName
**Description**

A user should not need to be aware of which "type" or `GeneralName` the subject is set in. Removing Type would simplify how a certificate identity is represented to be comprised of a subject and issuer only. This is also aligned with other Sigstore client implementations.

A similar conversation occurred in Fulcio previously (https://github.com/sigstore/fulcio/issues/716#issuecomment-1204549133), and the threat of "type confusion" was mitigated through CA enforcement that URIs look like URIs and emails look like emails, rather than client enforcement.

Relevant code:
* https://github.com/sigstore/sigstore-go/blob/main/pkg/fulcio/certificate/summarize.go#L37
* https://github.com/sigstore/sigstore-go/blob/main/pkg/verify/certificate_identity.go#L72-L76

I'd like to discuss this, I'm fine if we ultimately decide there is value in keeping this, but with the goal of making breaking changes before a 1.0, wanted to raise this.

cmurphy:
# Remove Type for SubjectAlternativeName

Simplify certificate identities by removing the Type attribute from the
SubjectAlternativeName type. The CA is expected to handle validating
that SAN values match their designated types.

Fixes https://github.com/sigstore/sigstore-go/issues/175

# Support OtherName SAN in Fulcio cert

Add support for parsing and verifying a Fulcio certificate with a
username identity issued from an OIDC provider. See [1].

[1] https://github.com/sigstore/fulcio/blob/main/docs/oid-info.md#1361415726417--othername-san

Fixes https://github.com/sigstore/sigstore-go/issues/55

#### Release Note

API change - `github.com/sigstore/sigstore-go/pkg/fulcio/certificate` : removed the `SubjectAlternativeNameType` type and changed the type of the `SubjectAlternativeName` field in the `Summary` struct to be a string instead of a struct.
API change - `github.com/sigstore/sigstore-go/pkg/verify`: changed the `SubjectAlternativeName` field of the `SubjectAlternativeNameMatcher` strcut to be a string instead of a struct. Changed the signature for functions `NewSANMatcher` and `NewShortCertificateIdentity`.
Feature: added support for verifying Fulcio certificates that use the OtherName GeneralName (RFC 5280 4.2.1.6) as a certificate identity.


pkg/verify/signed_entity.go
```diff
@@ -497,11 +499,26 @@ func (v *SignedEntityVerifier) Verify(entity SignedEntity, pb PolicyBuilder) (*V
...
 
+		if len(leafCert.UnhandledCriticalExtensions) > 0 {
+			var unhandledExts []asn1.ObjectIdentifier
+			for _, oid := range leafCert.UnhandledCriticalExtensions {
+				if !oid.Equal(cryptoutils.SANOID) {
+					unhandledExts = append(unhandledExts, oid)
+				}
+			}
+			leafCert.UnhandledCriticalExtensions = unhandledExts
+		}
```
steiza, line 553: Minor: this is probably worth a comment, as it took me some time to figure out! Maybe something like:

```
We handle the Subject Alternative Name (`SANOID`) in pkg/fulcio/certificate, so we remove it from the unhandled critical extensions list here, otherwise x.509 verification will fail
```

Feel free to edit / improve on that suggested wording!

pkg/verify/certificate_identity.go
```diff
@@ -141,8 +133,8 @@ func NewCertificateIdentity(sanMatcher SubjectAlternativeNameMatcher, extensions
 // NewShortCertificateIdentity provides a more convenient way of initializing
 // a CertificiateIdentity with a SAN and the Issuer OID extension. If you need
 // to check more OID extensions, use NewCertificateIdentity instead.
-func NewShortCertificateIdentity(issuer, sanValue, sanType, sanRegex string) (CertificateIdentity, error) {
-	sanMatcher, err := NewSANMatcher(sanValue, sanType, sanRegex)
+func NewShortCertificateIdentity(issuer, sanValue, sanRegex string) (CertificateIdentity, error) {
```
steiza, line 136: I don't want to expand the scope of this pull request, but should sigstore-go support an issuer regex in the future?

I noticed recently that cosign has that option: https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/options/certificate.go#L28

steiza (commented):
One minor comment / request for a explanatory code comment.

Otherwise I think this is good! This is a breaking change, but it's pretty easy for clients to update and remove that argument from the few places they call things like `NewSANMatcher()` or `NewShortCertificateIdentity()`. For example, in https://github.com/cli/cli/blob/trunk/pkg/cmd/attestation/verify/policy.go#L27-L43 we just pass an empty string for the `sanType` currently.

pkg/verify/certificate_identity.go
steiza: I made https://github.com/sigstore/sigstore-go/issues/234 to move the discussion out of this pull request, which I don't want to block.

pkg/verify/signed_entity.go
cmurphy: Done, thanks for pointing this out

pkg/verify/certificate_identity.go
cmurphy: Definitely in favor of supporting an issuer regex :+1:

cmurphy:
> This is a breaking change

Added a release note to the PR description

steiza (dismissed):
Nice work! LGTM

steiza (dismissed):
Can I approve my resolving merge conflicts?

cmurphy:
@steiza I went ahead and rebased to remove the merge commit

steiza (approved):
Thanks! Sorry, I should have landed this first this morning

Hayden-IO (commented):
Late LGTM, thank you!

kommendorkapten:
Thank you!

Issue #105, opened 2024-02-22
Policy for verifying with key


Hayden-IO:
**Description**

Currently, when verifying with a key, you must specify `WithoutIdentitiesUnsafe()` when creating the policy verifier because no identity is provided. It would be more accurate to have a method such as `WithKeyHint` or `WithKeyID`, since identities are not needed when verifying with a key. Either this method could take a key fingerprint, or it could take no fingerprint given the bundle should contain a key hint.

If verifying with a Sigstore bundle, a key hint should be provided as a [public key identifier](https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_bundle.proto#L94). As long as the trust root contains the public key, we can match based on this key hint (currently implemented [here](https://github.com/sigstore/sigstore-go/blob/main/pkg/root/trusted_material.go#L164). If this is not populated, either a) we should err out, or b) we should use a hint provided via the policy method.

Related: https://github.com/sigstore/protobuf-specs/issues/236 to simplify providing trusted keys.

cc @kommendorkapten @codysoyland @rdimitrov

kommendorkapten:
Having either `WithKeyHint` or `WithKey` as the identity for a policy is what I would expect, as that's how the verification options in the protobuf specs repo are described: https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_verification.proto#L92

For a specific verification, I would prefer `WithKey` as it minimizes the risk of confusion. If `WithKeyHint` is used, I think it's preferable if that selection happens in a client controlled function, as an implementation may only have a single key, and always return that, regardless what the key hint is (the key hint not authenticated in the bundle). 

That said, providing a matcher based on key hint (as @haydentherapper linked to) is a good utility option to provide, as it may be enough for certain scenarios.

kommendorkapten:
See my comment in https://github.com/sigstore/protobuf-specs/issues/236#issuecomment-1960814948 for more reasoning. 
The short story is that public keys are only weakly references via unauthenticated hints, certificate identities are strongly as the matching happens on authenticated data in the certificate.

steiza:
This was added in #235

PR #235, opened 2024-07-16
Add Policy WithKey() for verifing content you know is signed with a key

steiza:
#### Summary

To address #105.

Right now, if you have a bundle signed with a key, your verification policy needs to include `WithoutIdentitiesUnsafe`. This isn't great though because:
- bundles signed with keys aren't inherently unsafe
- there isn't a way to **require** your bundle was signed with a key - a bundle signed with a certificate would pass `WithoutIdentitiesUnsafe`

#### Release Note

* Added `WithKey()` policy option to require a bundle is signed with a key instead of a certificate

#### Documentation

N/A


kommendorkapten (dismissed):
Awesome!

kommendorkapten:
How about adding a brief overview/example of the steps to verify with a key (creating the trusted material and configuring the policy) to the `docs/verification.md`, that is not covered right now.

kommendorkapten (approved):
Love it!

Hayden-IO (commented):
Late LGTM, thanks!

PR #236, opened 2024-07-18
Add the ability to specify certificate identity via a regular expression

### LINKED ISSUE #234: Verify certificates by issuer regex
**Description**

When we verify certificates, today we accept:
- An issuer string (e.g. `https://token.actions.githubusercontent.com/`)
- A SAN string (the user or workload identity used in the certificate)
- A SAN regex (in case you want to support a range of SAN strings)

Noticeably absent is an issuer regex, which is a feature [supported by cosign](https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/options/certificate.go#L28). We should consider adding this capability to sigstore-go.

See https://github.com/sigstore/sigstore-go/pull/229#discussion_r1679457620 for the initial discussion.

steiza:
#### Summary

Fixes #234

Cosign allows you to verify bundles signed with a certificate by providing a regular expression to match the certificate identity. Previously, sigstore-go required you to specify the exact certificate identity. This change adds support for sigstore-go to match the certificate identity with a regular expression.

#### Release Note

* Allows specifying certificate identity with a regular expression, for verifying bundles
* Breaking change: additional string argument to `verify.NewShortCertificateIdentity()` and `verify.NewCertificateIdentity()`

#### Documentation

N/A

### DIFF
--- a/cmd/sigstore-go/main.go
+++ b/cmd/sigstore-go/main.go
@@ -41,6 +41,7 @@ var artifact *string
 var artifactDigest *string
 var artifactDigestAlgorithm *string
 var expectedOIDIssuer *string
+var expectedOIDIssuerRegex *string
 var expectedSAN *string
 var expectedSANRegex *string
 var requireTimestamp *bool
@@ -58,6 +59,7 @@ func init() {
 	artifactDigest = flag.String("artifact-digest", "", "Hex-encoded digest of artifact to verify")
 	artifactDigestAlgorithm = flag.String("artifact-digest-algorithm", "sha256", "Digest algorithm")
 	expectedOIDIssuer = flag.String("expectedIssuer", "", "The expected OIDC issuer for the signing certificate")
+	expectedOIDIssuerRegex = flag.String("expectedIssuerRegex", "", "The expected OIDC issuer for the signing certificate")
 	expectedSAN = flag.String("expectedSAN", "", "The expected identity in the signing certificate's SAN extension")
 	expectedSANRegex = flag.String("expectedSANRegex", "", "The expected identity in the signing certificate's SAN extension")
 	requireTimestamp = flag.Bool("requireTimestamp", true, "Require either an RFC3161 signed timestamp or log entry integrated timestamp")
@@ -120,7 +122,7 @@ func run() error {
 		verifierConfig = append(verifierConfig, verify.WithOnlineVerification())
 	}
 
-	certID, err := verify.NewShortCertificateIdentity(*expectedOIDIssuer, "", *expectedSAN, *expectedSANRegex)
+	certID, err := verify.NewShortCertificateIdentity(*expectedOIDIssuer, *expectedOIDIssuerRegex, *expectedSAN, *expectedSANRegex)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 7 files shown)


cmd/sigstore-go/main.go
```diff
@@ -120,7 +120,7 @@ func run() error {
 		verifierConfig = append(verifierConfig, verify.WithOnlineVerification())
 	}
 
-	certID, err := verify.NewShortCertificateIdentity(*expectedOIDIssuer, *expectedSAN, *expectedSANRegex)
+	certID, err := verify.NewShortCertificateIdentity(*expectedOIDIssuer, "", *expectedSAN, *expectedSANRegex)
```
cmurphy: Maybe a `expectedOIDIssuerRegex` flag should be added to the binary?

examples/oci-image-verification/main.go
```diff
@@ -134,7 +134,7 @@ func run() error {
 	}
 
 	if *expectedOIDIssuer != "" || *expectedSAN != "" || *expectedSANRegex != "" {
```
cmurphy: and also to the  OCI example

pkg/verify/certificate_identity.go
```diff
@@ -116,15 +117,19 @@ func (s SubjectAlternativeNameMatcher) Verify(actualCert certificate.Summary) er
 	return nil
 }
 
-func NewCertificateIdentity(sanMatcher SubjectAlternativeNameMatcher, extensions certificate.Extensions) (CertificateIdentity, error) {
+func NewCertificateIdentity(sanMatcher SubjectAlternativeNameMatcher, issuerRegexp *regexp.Regexp, extensions certificate.Extensions) (CertificateIdentity, error) {
```
cmurphy: It might be more consistent with `SubjectAlternativeNameMatcher.Regexp` to not use a pointer here.

pkg/verify/certificate_identity.go
```diff
@@ -133,13 +138,23 @@ func NewCertificateIdentity(sanMatcher SubjectAlternativeNameMatcher, extensions
...
+		}
+	}
+
 	sanMatcher, err := NewSANMatcher(sanValue, sanRegex)
 	if err != nil {
 		return CertificateIdentity{}, err
 	}
 
-	return NewCertificateIdentity(sanMatcher, certificate.Extensions{Issuer: issuer})
+	return NewCertificateIdentity(sanMatcher, r, certificate.Extensions{Issuer: issuer})
```
Hayden-IO: nit, should we set `Issuer` only if `issuer != ""`? `CompareExtensions` would skip it anyways (https://github.com/sigstore/sigstore-go/blob/8554eb6de5ac145044cd1f0f869fa536d59a1d46/pkg/fulcio/certificate/summarize.go#L79) though

Hayden-IO: It's inconsistent to have a SANMatcher for one property and separate issuer matching into separate fields. An alternative could be to have an IssuerMatcher struct as well, and then omit `Issuer` from the extensions?

pkg/verify/certificate_identity.go
steiza: Makes sense; this got replaced with `IssuerMatcher`

Hayden-IO (approved):
Thanks!

Issue #48, opened 2023-12-15
Include skipped signatures in VerificationResult


Hayden-IO:
**Description**

https://github.com/sigstore/sigstore-go/pull/47 and https://github.com/sigstore/sigstore-go/pull/45 introduce skipping log and TSA signatures respectively that the trust bundle cannot verify. This information is not passed back to the verifier, the signatures are just silently skipped over.

We can update [VerificationResults](https://github.com/sigstore/sigstore-go/blob/9a59c7f51a05d2305a8ba4ab9343416aaaeb16d6/pkg/verify/signed_entity.go#L160) to pass this information back, which could be helpful for debugging that the bundle contains the expected trust root material.

Hayden-IO:
We might also want to split out parsing errors from verification errors. For example, for timestamp verification, we will skip a timestamp either if it is an invalid RFC3161 structure or if verification fails.

Hayden-IO:
At the sigstore-go meeting, we discussed that `VerificationResult` currently contains metadata that is verified. @phillmv pointed out that this struct could be passed to a policy controller for further evaluation. Adding a field like `UnverifiedResults` into this struct might mislead verifiers.

dmitris:
I had a related experience and asked a question on [Slack](https://sigstore.slack.com/archives/C0440BFT43H/p1729613321322739):

```
a question about this error handling instance:
https://github.com/sigstore/sigstore-go/blob/main/pkg/verify/tsa.go#L113-L116
		// Ensure timestamp responses are from trusted sources
		timestamp, err := tsaverification.VerifyTimestampResponse(signedTimestamp, bytes.NewReader(dsseSignatureBytes), trustedRootVerificationOptions)
		if err != nil {
			continue
		}
don't we want to expose / return those errors in some form?  For example, I'm now struggling with an ASN.1 parsing error  from [github.com/digitorus/timestamp.ParseResponse](http://github.com/digitorus/timestamp.ParseResponse) (called from [github.com/sigstore/timestamp-authority/pkg/verification.VerifyTimestampResponse](http://github.com/sigstore/timestamp-authority/pkg/verification.VerifyTimestampResponse)):

error parsing response into Timestamp: asn1: structure error: tags don't match (16 vs {class:1 tag:13 length:73 isCompound:false}) {optional:false explicit:false application:false private:false defaultValue:<nil> tag:<nil> stringType:0 timeType:0 set:false omitEmpty:false} response @2

[...]

PR #237, opened 2024-07-20
include skipped signatures in VerificationResult

ggrangel:
#### Summary
Solves #48.

Ideally, I would return the unverified signatures from [here (TSA)](https://github.com/sigstore/sigstore-go/blob/main/pkg/verify/tsa.go#L62-L65) and [here (logs)](https://github.com/sigstore/sigstore-go/blob/main/pkg/verify/sct.go#L50C1-L54C1). However, this would require me to change the signature of exported functions, which I'm not sure if I should.

To avoid this, for each signature type, I've created a function that loops through all available signatures and mark the ones that weren't verified.

Additionally, I couldn't find a reliable way to address [this comment](https://github.com/sigstore/sigstore-go/issues/48#issuecomment-1911006482) since the [timestamp-authority package](https://github.com/sigstore/timestamp-authority/blob/main/pkg/verification/verify.go#L213C1-L219C3) doesn't generate distinct errors for different failure cases. We would have to parse the error strings which I consider a weak approach because they're easily changed as code base evolves. Also, this would likely require changes to the API of exported functions, which, again, I'm not sure if it's desirable.

#### Release Note

NONE

#### Documentation

No documentation update required

### DIFF
--- a/pkg/verify/signed_entity.go
+++ b/pkg/verify/signed_entity.go
@@ -15,12 +15,15 @@
 package verify
 
 import (
+	"crypto/x509"
 	"encoding/asn1"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"io"
 	"time"
 
+	"github.com/google/certificate-transparency-go/x509util"
 	"github.com/in-toto/in-toto-golang/in_toto"
 	"github.com/sigstore/sigstore-go/pkg/fulcio/certificate"
 	"github.com/sigstore/sigstore-go/pkg/root"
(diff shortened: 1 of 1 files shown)


cmurphy (commented):
> Ideally, I would return the unverified signatures from [here (TSA)](https://github.com/sigstore/sigstore-go/blob/main/pkg/verify/tsa.go#L62-L65) and [here (logs)](https://github.com/sigstore/sigstore-go/blob/main/pkg/verify/sct.go#L50C1-L54C1). However, this would require me to change the signature of exported functions, which I'm not sure if I should.

This library is pre-1.0, and this issue is one of several that are slotted to be addressed before the 1.0 in case there are any breaking changes, so if it is a better solution to change the function signature, please go for it. Also, please mention the breaking changes in the Release Note section of the PR description.

Could you add some unit tests?

Could you update the example serialized result in the [docs](https://github.com/sigstore/sigstore-go/blob/main/docs/verification.md#go-api)?

PR #245, opened 2024-07-25
handle case insensitivity for compare extensions

ejahnGithub:
#### Summary

ref: https://github.com/sigstore/sigstore-go/issues/89

This modifies the API for verifying certificate identity (SAN and Extensions) to turn CompareExtensions case insensitivity in order to handle cases that `OIDSourceRepositoryURI` or other fields can have different cases. 

Signed-off-by: Eugene Jahn [ejahngithub@github.com](mailto:ejahngithub@github.com)

### DIFF
--- a/pkg/fulcio/certificate/summarize.go
+++ b/pkg/fulcio/certificate/summarize.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"strings"
 
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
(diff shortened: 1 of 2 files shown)


Hayden-IO (changes requested):
This seems risky, as we don't know if the extension value should be case sensitive (e.g maybe `UserX` is different than `userx`). If this is meant for a single extension, I would recommend we either a) create a policy to dictate whether an ext is case sensitive or not, b) allow for passing multiple values for an ext.

PR #253, opened 2024-08-01
Support intermediate certificates in bundle

### LINKED ISSUE #132: Support intermediate certificates in VerificationMaterial certificate chain
**Description**

Currently, in https://github.com/sigstore/sigstore-go/blob/main/pkg/bundle/bundle.go when we get a bundle that includes a certificate chain in the verification materials, we assume there is only 1 leaf certificate in the "chain" and no intermediates.

This needs to change if we're going to support certificate chains that include intermediates. 

This came up in https://github.com/sigstore/sigstore-go/pull/101#discussion_r1543215948.

cmurphy:
Add support for processing and verifying a v0.3 bundle that contains a `X509CertificateChain` rather than a single X.509 certificate or public key.

Fixes #132

### DIFF
--- a/pkg/bundle/bundle.go
+++ b/pkg/bundle/bundle.go
@@ -108,15 +108,6 @@ func (b *Bundle) validate() error {
 		}
 	}
 
-	// if bundle version >= v0.3, require verification material to not be X.509 certificate chain (only single certificate is allowed)
-	if semver.Compare(bundleVersion, "v0.3") >= 0 {
-		certs := b.Bundle.VerificationMaterial.GetX509CertificateChain()
-
-		if certs != nil {
-			return errors.New("verification material cannot be X.509 certificate chain (for bundle v0.3)")
-		}
-	}
-
 	// if bundle version is >= v0.4, return error as this version is not supported
 	if semver.Compare(bundleVersion, "v0.4") >= 0 {
 		return fmt.Errorf("%w: bundle version %s is not yet supported", ErrUnsupportedMediaType, bundleVersion)
@@ -250,14 +241,18 @@ func (b *Bundle) VerificationContent() (verify.VerificationContent, error) {
 		if len(certs) == 0 || certs[0].RawBytes == nil {
 			return nil, ErrMissingVerificationMaterial
 		}
-		parsedCert, err := x509.ParseCertificate(certs[0].RawBytes)
-		if err != nil {
-			return nil, ErrValidationError(err)
+		parsedCerts := make([]*x509.Certificate, len(certs))
+		var err error
+		for i, c := range certs {
+			parsedCerts[i], err = x509.ParseCertificate(c.RawBytes)
+			if err != nil {
+				return nil, ErrValidationError(err)
+			}
 		}
-		cert := &Certificate{
-			Certificate: parsedCert,
+		certChain := &CertificateChain{
+			Certificates: parsedCerts,
 		}
-		return cert, nil
+		return certChain, nil
 	case *protobundle.VerificationMaterial_Certificate:
 		if content.Certificate == nil || content.Certificate.RawBytes == nil {
 			return nil, ErrMissingVerificationMaterial
@@ -266,8 +261,8 @@ func (b *Bundle) VerificationContent() (verify.VerificationContent, error) {
 		if err != nil {
 			return nil, ErrValidationError(err)
 		}
-		cert := &Certificate{
-			Certificate: parsedCert,
+		cert := &CertificateChain{
+			Certificates: []*x509.Certificate{parsedCert},
 		}
 		return cert, nil
 	case *protobundle.VerificationMaterial_PublicKey:
--- a/pkg/bundle/verification_content.go
+++ b/pkg/bundle/verification_content.go
@@ -23,8 +23,8 @@ import (
 	"github.com/sigstore/sigstore-go/pkg/verify"
 )
 
-type Certificate struct {
-	*x509.Certificate
+type CertificateChain struct {
+	Certificates []*x509.Certificate
 }
 
 type PublicKey struct {
@@ -35,25 +35,40 @@ func (pk PublicKey) Hint() string {
 	return pk.hint
 }
 
-func (c *Certificate) CompareKey(key any, _ root.TrustedMaterial) bool {
+func (c *CertificateChain) CompareKey(key any, _ root.TrustedMaterial) bool {
+	if len(c.Certificates) < 1 {
+		return false
+	}
 	x509Key, ok := key.(*x509.Certificate)
 	if !ok {
 		return false
 	}
-
-	return c.Certificate.Equal(x509Key)
+	return c.Certificates[0].Equal(x509Key)
 }
 
-func (c *Certificate) ValidAtTime(t time.Time, _ root.TrustedMaterial) bool {
-	return !(c.Certificate.NotAfter.Before(t) || c.Certificate.NotBefore.After(t))
+func (c *CertificateChain) ValidAtTime(t time.Time, _ root.TrustedMaterial) bool {
+	if len(c.Certificates) < 1 {
+		return false
+	}
+	return !(c.Certificates[0].NotAfter.Before(t) || c.Certificates[0].NotBefore.After(t))
 }
 
-func (c *Certificate) GetCertificate() *x509.Certificate {
-	return c.Certificate
+func (c *CertificateChain) GetLeafCertificate() *x509.Certificate {
+	if len(c.Certificates) < 1 {
+		return nil
+	}
+	return c.Certificates[0]
 }
 
-func (c *Certificate) HasPublicKey() (verify.PublicKeyProvider, bool) {
+func (c *CertificateChain) HasPublicKey() (verify.PublicKeyProvider, bool) {
 	return PublicKey{}, false
+}
+
+func (c *CertificateChain) GetIntermediates() []*x509.Certificate {
+	if len(c.Certificates) < 2 {
+		return nil
+	}
+	return c.Certificates[1:]
 }
 
 func (pk *PublicKey) CompareKey(key any, tm root.TrustedMaterial) bool {
@@ -79,10 +94,14 @@ func (pk *PublicKey) ValidAtTime(t time.Time, tm root.TrustedMaterial) bool {
 	return verifier.ValidAtTime(t)
 }
 
-func (pk *PublicKey) GetCertificate() *x509.Certificate {
+func (pk *PublicKey) GetLeafCertificate() *x509.Certificate {
 	return nil
 }
 
 func (pk *PublicKey) HasPublicKey() (verify.PublicKeyProvider, bool) {
 	return *pk, true
+}
+
+func (pk *PublicKey) GetIntermediates() []*x509.Certificate {
+	return nil
 }
--- a/pkg/bundle/bundle_test.go
+++ b/pkg/bundle/bundle_test.go
@@ -520,7 +520,7 @@ func Test_validate(t *testing.T) {
 					Content: &protobundle.Bundle_MessageSignature{},
 				},
 			},
-			wantErr: true,
+			wantErr: false,
 		},
 		{
 			name: "v0.3 without x.509 certificate chain",
@@ -584,11 +584,12 @@ func TestVerificationContent(t *testing.T) {
 	leafDer, err := x509.CreateCertificate(rand.Reader, leafCert, caCert, &leafKey.PublicKey, caKey)
 	require.NoError(t, err)
 	tests := []struct {
-		name            string
-		pb              Bundle
-		wantCertificate bool
-		wantPublicKey   bool
-		wantErr         bool
+		name                 string
+		pb                   Bundle
+		wantCertificateChain bool
+		wantCertificate      bool
+		wantPublicKey        bool
+		wantErr              bool
 	}{
 		{
 			name: "no verification material",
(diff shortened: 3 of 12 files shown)


pkg/bundle/verification_content.go
```diff
@@ -86,3 +94,39 @@ func (pk *PublicKey) GetCertificate() *x509.Certificate {
...
+		return nil
+	}
+	return cc.Certificates[0]
+}
+
+func (cc *CertificateChain) HasPublicKey() (verify.PublicKeyProvider, bool) {
+	return PublicKey{}, false
+}
+
+func (cc *CertificateChain) GetCertificateChain() []*x509.Certificate {
```
steiza: Minor: maybe this should be `GetIntermediateCerts()`, as it doesn't return the whole chain?

pkg/bundle/bundle.go
```diff
@@ -98,15 +98,6 @@ func (b *ProtobufBundle) validate() error {
...
 	}
 
-	// if bundle version >= v0.3, require verification material to not be X.509 certificate chain (only single certificate is allowed)
-	if semver.Compare(bundleVersion, "v0.3") >= 0 {
-		certs := b.Bundle.VerificationMaterial.GetX509CertificateChain()
-
-		if certs != nil {
-			return errors.New("verification material cannot be X.509 certificate chain (for bundle v0.3)")
-		}
-	}
```
steiza, line 118: This is required [if your verification materials are from PGI and you're using keyless signing](https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_bundle.proto#L91-L92), but I don't know of an easy way to test for that here...

pkg/bundle/verification_content.go
```diff
@@ -35,6 +35,10 @@ func (pk PublicKey) Hint() string {
 	return pk.hint
 }
 
+type CertificateChain struct {
+	Certificates []*x509.Certificate
+}
```
steiza: I know there's a difference in `protobundle` between `VerificationMaterial_X509CertificateChain` and `VerificationMaterial_Certificate`, but if we had both use the `CertificateChain` type here, and had `.GetCertificateChain()` (or `.GetIntermediateCerts()` as I propose we call it) return an empty list if there's just a leaf certificate, I don't think we'd need `CertificateChain` and `Certificate` types?

@haydentherapper do you really want two separate types here, or have you [changed your mind](https://github.com/sigstore/sigstore-go/pull/101/files#r1543211390)?

pkg/bundle/verification_content.go
```diff
@@ -86,3 +94,39 @@ func (pk *PublicKey) GetCertificate() *x509.Certificate {
...
+
+func (cc *CertificateChain) ValidAtTime(t time.Time, tm root.TrustedMaterial) bool {
+	if len(cc.Certificates) < 1 {
+		return false
+	}
+	return (&Certificate{cc.Certificates[0]}).ValidAtTime(t, tm)
+}
+
+func (cc *CertificateChain) GetCertificate() *x509.Certificate {
```
steiza: Minor: would `GetLeafCertificate()` be more clear?

steiza (commented):
I don't totally understand the use-case of including intermediate certificates in your bundle verification materials, but maybe there's some private deployments that do so for some reason? It seems like you're better off with those intermediates in your trust root, for more control over revocation?

At any rate, I'm not opposed to these changes, but I do have some clarifying questions / naming suggestions.

cmurphy:
> I don't totally understand the use-case of including intermediate certificates in your bundle verification materials, but maybe there's some private deployments that do so for some reason?

@haydentherapper could speak more to this, but my understanding is it's wanted for some private deployments. But I don't know if anyone has specifically asked for it. If no one has actually said they need this, I wouldn't mind punting on it for a while.

pkg/bundle/bundle.go
cmurphy: Could we check the certificate issuer in the leaf certificate to see if it matches `O=sigstore.dev, CN=sigstore-intermediate` and do this validation in that case?

Also, is it critical to validate this right here? If we had a trusted root from PGI and a certificate chain in the bundle, wouldn't it become clear in `VerifyLeafCertificate` that the intermediates in the bundle hadn't been issued by PGI?

Hayden-IO:
For the ask, yes, we had someone from nvidia asking about support for this in Cosign. Given it's already supported in Cosign, I'd love to see this supported here in sigstore-go as we migrate Cosign's internals over to sigstore-go.

There's maybe a bit of a meta question which is, do we want to try to add BYO-PKI features into the existing API, or create separate APIs that still live in sigstore-go but are intentionally designed to be more flexible and less bound to the Sigstore spec?

pkg/bundle/bundle.go
Hayden-IO: This came up when we were writing these requirements previously - I think this is more of a promise to clients that a trust root coming from PGI will meet the stated requirements, for if you're writing a client only for the public instance. I don't think we need to enforce that here.

pkg/bundle/verification_content.go
Hayden-IO: I think that'd be fine to merge the types here.

steiza:
> There's maybe a bit of a meta question which is, do we want to try to add BYO-PKI features into the existing API, or create separate APIs that still live in sigstore-go but are intentionally designed to be more flexible and less bound to the Sigstore spec?

Ideally there's one API, and the verification policies are flexible enough to cover BYO-PKI, although maybe we'll have to see how convoluted the BYO-PKI cases end up being.

But for these changes, I think that they can work with the existing API.

pkg/bundle/verification_content.go
cmurphy: Merged Certificate to CertificateChain

pkg/bundle/verification_content.go
cmurphy: done

pkg/bundle/verification_content.go
cmurphy: Changed to `GetIntermediates`

steiza (approved):
Looks good to me!

I had to check what order we expect `GetIntermediates()` to return certificates in, but it looks like this is compatible with our call to `ctutil`.

codysoyland:
This seems like a step back from the decision in https://github.com/sigstore/protobuf-specs/issues/190 / https://github.com/sigstore/protobuf-specs/pull/191 that v0.3 bundles may only contain a leaf cert. Did I misunderstand that spec change? I do see the current [client spec](https://github.com/sigstore/architecture-docs/blob/main/client-spec.md#41-inputs) allows for intermediates in the verification materials, but I thought it was explicitly disallowed in a previous version.

> There's maybe a bit of a meta question which is, do we want to try to add BYO-PKI features into the existing API, or create separate APIs that still live in sigstore-go but are intentionally designed to be more flexible and less bound to the Sigstore spec?

Is there an issue with providing a custom `root.TrustedMaterial` that contains additional intermediates? We could provide a helper function that accepts an existing `root.TrustedMaterial` and a list of additional certificates to augment it, for example:

```go
func AugmentTrustedMaterialWithIntermediates(root.TrustedMaterial, []*x509.Certificate) root.TrustedMaterial
```

steiza:
> This seems like a step back from the decision in https://github.com/sigstore/protobuf-specs/issues/190 / https://github.com/sigstore/protobuf-specs/pull/191 that v0.3 bundles may only contain a leaf cert.

I think where this landed is that we want users of PGI to only have the leaf cert in the bundle, however there are private deployments that expect to have intermediate certificates in the bundle, even for bundle v0.3.

From https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_bundle.proto:

>         // This allows key material to be conveyed in one of three forms:
>         ...
>         // 3. A single X.509 certificate, which MUST be a leaf certificate conveying
>         //    the signing key.
>         ...
>         // When used in a `0.3` bundle with the PGI and "keyless" signing,
>         // form (3) MUST be used.

My read of that is v0.3 bundles that **are not** using PGI may include intermediate certificates, which I think was the motivation for requesting this change.

We could bring back the check that v0.3 bundles only have leaf certs, but we'd need some way to distinguish between PGI and not - maybe checking the certificate subject?

cmurphy:
Would the concerns about violating the spec be alleviated if the bundle wasn't providing the intermediates but something else was?

@steiza how were you envisioning the intermediates be supplied when you wrote https://github.com/sigstore/sigstore-go/issues/132?

steiza:
> @steiza how were you envisioning the intermediates be supplied when you wrote https://github.com/sigstore/sigstore-go/issues/132?

I created that issue because @haydentherapper asked me to 😆 https://github.com/sigstore/sigstore-go/pull/101#discussion_r1543215948

I'm fine with either not having the check (like how the code currently is) or adding the check back in conditional on the certificate subject - but I'm open to other suggestions here!

Hayden-IO:
> My read of that is v0.3 bundles that are not using PGI may include intermediate certificates, which I think was the motivation for requesting this change.
> We could bring back the check that v0.3 bundles only have leaf certs, but we'd need some way to distinguish between PGI and not - maybe checking the certificate subject?

My two cents, the comments are meant as guidance to clients for what to implement. If a client wants to only support PGI, then it doesn't need to handle the non-PGI cases. The bundle being flexible allows it to be used more broadly for signing and verification - we actually got this feedback from chatting with a sigstore integrator that wanted to use their own PKI.

There is one attack that would be possible without a check - an intermediate CA is removed from the trust root due to compromise, with a malicious bundle provided that still references that intermediate. However, for the public good instance, this is a fairly unlikely scenario, as we would rotate the root and intermediate unless we were extremely confident that we could detect where the compromise occurred.

As a compromise, what do you think about a policy flag like `TrustProvidedIntermediates`, with a default value of false? For clients that only support the well-lit path for verification, no need to set it. For Cosign, when you set `--private-infrastructure`, we can set this flag as well.

Hayden-IO (approved):
Also I'm good to merge this as-is and follow up with a PR for a flag if there's consensus.

phillmv:
Drive by comment, I know this probably got discussed synchronously earlier today but I feel this contradicts the effort that went into adding the `certificate` field to the protobuf spec in the first place.

Q for whatever entity wants this feature: if they're BYO-PKI, presumably in a closed corporate environment, what is the blocker for distributing/trusting the intermediate certificate?

phillmv:
(I don't really have a position on the security properties of `certChain` vs `certificate`, putting it behind a policy flag seems cromulent)

codysoyland:
>There is one attack that would be possible without a check - an intermediate CA is removed from the trust root due to compromise, with a malicious bundle provided that still references that intermediate. However, for the public good instance, this is a fairly unlikely scenario, as we would rotate the root and intermediate unless we were extremely confident that we could detect where the compromise occurred.
>
>As a compromise, what do you think about a policy flag like TrustProvidedIntermediates, with a default value of false? For clients that only support the well-lit path for verification, no need to set it. For Cosign, when you set --private-infrastructure, we can set this flag as well.

This is my understanding of the threat as well, and I support adding a policy flag like this. 👍🏻

kommendorkapten:
> There is one attack that would be possible without a check - an intermediate CA is removed from the trust root due to compromise, with a malicious bundle provided that still references that intermediate. However, for the public good instance, this is a fairly unlikely scenario, as we would rotate the root and intermediate unless we were extremely confident that we could detect where the compromise occurred.

> As a compromise, what do you think about a policy flag like `TrustProvidedIntermediates`, with a default value of false? For clients that only support the well-lit path for verification, no need to set it. For Cosign, when you set `--private-infrastructure`, we can set this flag as well.

I'm thinking more about the implementations here. As we don't offer any kind of certificate verification within sigstore-go (all certificates from the bundle or the trust root are deemed valid). I think this is the right choice right now:
* The PGI leaf certificates are short lived -- low risk of them being compromised
* The trust root is managed via TUF -- extremly fast failover due to compromised certificates (intermediates or root)

To support the other scenario, where untrusted intermediates are provided via the bundle for chain building. Instead of a flag/policy, how about a certificate verification function provided by the client? For the pgi use-case, this could be a simple function

```go
func ValidateCertificate(c *x509.Certifiicate) (bool, error) {
        return true, nil
}
```

Buf for clients using intermediates in the bundle, they **want** to verify the certs during chain building, either by comparing against a CRL or relying on OCSP. (Note that such functions can be hard to implement securely, see [CVE-2023-2422](https://github.com/advisories/GHSA-3qh5-qqj2-c78f) as an example where client certificate validation can fail. I don't suggest we implement it, only provide such a hook.

If we offer such an interface during verification, we can make sure that even when untrusted intermediates are being provided, a client implementing that have the necessary hooks to perform certificate verification.

The other option is the force the client to perform out of band verification of all certificates, which works, but the UX is slightly worse as they would need to parse the bundle and the trust root. A callback function is nicer IMHO, and pushes the complexity _out of_ sigstore-go into the client where it belongs.

edig: spelling error

Hayden-IO:
I'd be supportive of that design. It would make it easier to support requests like https://github.com/sigstore/cosign/issues/2568, asking for CRLs.

For the interface, would we want to include intermediate and root certificates as well in the function parameters?

From Cosign's perspective, if we're only dealing with untrusted intermediates, these solutions - a policy flag or implementing a validate function - are roughly equivalent. If we go with a validate function, I'd suggest we put it in sigstore/sigstore as shared logic between Golang implementations.

kommendorkapten:
Created this issue for discussion of certificate verification: https://github.com/sigstore/sigstore-go/issues/298

Hayden-IO:
Closing - https://github.com/sigstore/sigstore-go/issues/132#issuecomment-2859823066

Thank you for working on this! We'll revisit in Cosign whenever we tackle redoing the BYO PKI features in Cosign.

PR #258, opened 2024-08-06
Fix verify DSSE bundles (after signing)

### LINKED ISSUE #257: Bundle signer does not correctly verify artifact if bundle content is DSSE
**Description**

As noted in https://github.com/sigstore/sigstore-go/issues/256:

If `BundleOptions.TrustedRoot` is set to a trusted root, then `sign.Bundle` will attempt to verify the bundle against the given trusted root. If the bundle content is DSSE, the artifact provided to the verifier is [the PAE bytes of the envelope](https://github.com/sigstore/sigstore-go/blob/6f7e99d5908575d5faaddb5a688f83430ba989b1/pkg/sign/signer.go#L152), however, the verifier requires that if `WithArtifact`/`WithArtifactDigest` is provided, the digest is [compared against the subjects in the attestation](https://github.com/sigstore/sigstore-go/blob/6f7e99d5908575d5faaddb5a688f83430ba989b1/pkg/verify/signature.go#L183-L193), not the PAE bytes of the DSSE.

The routine that does this verification should be updated to extract the subject digest and use `WithArtifactDigest` when the provided content is a DSSE.

**Version**

steiza:
#### Summary

Fixes #257.

When signing, if you optionally supply a trusted root we will attempt to verify the bundle before we return it.

Previously we were using the wrong artifact digest for DSSE signing. We could add a way to specify the artifact referred to in DSSE, but we are already trusting the certificate identity in the user-supplied id token.

You can test with something like this (which previously would fail):

```
$ go run examples/sigstore-go-signing/main.go -id-token "..." -in-toto -rekor examples/sigstore-go-signing/intoto.txt 
```

#### Release Note

NONE

#### Documentation

N/A

### DIFF
--- a/pkg/sign/signer.go
+++ b/pkg/sign/signer.go
@@ -15,7 +15,6 @@
 package sign
 
 import (
-	"bytes"
 	"context"
 	"encoding/pem"
 	"errors"
@@ -149,7 +148,7 @@ func Bundle(content Content, keypair Keypair, opts BundleOptions) (*protobundle.
 			return nil, err
 		}
 
-		artifactOpts := verify.WithArtifact(bytes.NewReader(content.PreAuthEncoding()))
+		artifactOpts := verify.WithoutArtifactUnsafe()
 		policy := verify.NewPolicy(artifactOpts, verify.WithoutIdentitiesUnsafe())
 		_, err = sev.Verify(protobundle, policy)
 		if err != nil {
--- a/pkg/sign/signer.go
+++ b/pkg/sign/signer.go
@@ -15,6 +15,7 @@
 package sign
 
 import (
+	"bytes"
 	"context"
 	"encoding/pem"
 	"errors"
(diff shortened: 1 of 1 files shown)


pkg/sign/signer.go
```diff
@@ -149,7 +148,7 @@ func Bundle(content Content, keypair Keypair, opts BundleOptions) (*protobundle.
 			return nil, err
 		}
 
-		artifactOpts := verify.WithArtifact(bytes.NewReader(content.PreAuthEncoding()))
+		artifactOpts := verify.WithoutArtifactUnsafe()
```
Hayden-IO, line 156: Can we add a comment that says we are just doing a signature check, and that this isn’t safe to do for verification, or something similar?

pkg/sign/signer.go
```diff
@@ -149,7 +149,15 @@ func Bundle(content Content, keypair Keypair, opts BundleOptions) (*protobundle.
 			return nil, err
 		}
 
-		artifactOpts := verify.WithArtifact(bytes.NewReader(content.PreAuthEncoding()))
+		// Generally, you should provide an artifact when verifying.
+		//
+		// However, we just signed the DSSE object trusting the user has
+		// referenced the artifact(s) they intended.
+		artifactOpts := verify.WithoutArtifactUnsafe()
+		if bundle.GetMessageSignature() != nil {
```
Hayden-IO, line 157: Why is this conditional necessary, and we only skip artifact verification?

Hayden-IO (approved):

pkg/sign/signer.go
steiza: sigstore-go requires that some artifact reference is provided when verifying a message signature: https://github.com/sigstore/sigstore-go/blob/6f7e99d5908575d5faaddb5a688f83430ba989b1/pkg/verify/signature.go#L48

Hayden-IO: Oh yea, that makes sense, because you need an artifact for a signature, but not for a DSSE since that's just checking subject claims

ramonpetgrave64:
Since this is a bugfix, how soon can you do another release?

steiza:
> Since this is a bugfix, how soon can you do another release?

Usually we cut release as-needed, so this morning I made https://github.com/sigstore/sigstore-go/releases/tag/v0.6.0

PR #260, opened 2024-08-08
Fix crash with missing checkpoint

### LINKED ISSUE #259: Verification crashes for `v0.1` bundles missing checkpoints
## Description

I observed this behavior while trying to cross-check `sigstore-python`'s handling of `v0.1` bundles with other clients.

Reproduction steps:

```
wget https://www.python.org/ftp/python/3.12.5/Python-3.12.5.tgz
wget https://www.python.org/ftp/python/3.12.5/Python-3.12.5.tgz.sigstore
sigstore-go -artifact Python-3.12.5.tgz -expectedSAN 'thomas@python.org' Python-3.12.5.tgz.sigstore 
```

Running that fails with:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x30 pc=0x10293ec40]

goroutine 1 [running]:
github.com/sigstore/sigstore-go/pkg/tlog.ParseEntry(0x140003c8f50)
	/Users/william/go/pkg/mod/github.com/sigstore/sigstore-go@v0.5.1/pkg/tlog/entry.go:132 +0x500
github.com/sigstore/sigstore-go/pkg/bundle.(*ProtobufBundle).TlogEntries(0x140003d4860)
	/Users/william/go/pkg/mod/github.com/sigstore/sigstore-go@v0.5.1/pkg/bundle/bundle.go:252 +0xb0
github.com/sigstore/sigstore-go/pkg/bundle.(*ProtobufBundle).validate(0x140003d4860)
	/Users/william/go/pkg/mod/github.com/sigstore/sigstore-go@v0.5.1/pkg/bundle/bundle.go:84 +0x100
github.com/sigstore/sigstore-go/pkg/bundle.(*ProtobufBundle).UnmarshalJSON(0x140003d4860, {0x14000389500, 0x14ad, 0x14ae})
	/Users/william/go/pkg/mod/github.com/sigstore/sigstore-go@v0.5.1/pkg/bundle/bundle.go:189 +0xa4
github.com/sigstore/sigstore-go/pkg/bundle.LoadJSONFromPath({0x16d997928, 0x33})
	/Users/william/go/pkg/mod/github.com/sigstore/sigstore-go@v0.5.1/pkg/bundle/bundle.go:170 +0x8c
main.run()
	/Users/william/go/pkg/mod/github.com/sigstore/sigstore-go@v0.5.1/cmd/sigstore-go/main.go:94 +0x58
main.main()
	/Users/william/go/pkg/mod/github.com/sigstore/sigstore-go@v0.5.1/cmd/sigstore-go/main.go:87 +0x1c
```

From a quick triage, that looks like it fails on this `swag.String` ctor:

https://github.com/sigstore/sigstore-go/blob/004c4250f082b69a03df43a95930fd72cef63549/pkg/tlog/entry.go#L132

...which I *suspect* fails because `InclusionProof.Checkpoint` is completely missing from the bundle, which gets silently ignored during unpacking because `protobuf` is very malleable about missing items.

## Version

I tested this with `go install github.com/sigstore/sigstore-go/cmd/sigstore-go@latest`, which I _believe_ should be installing the latest tag (`v0.5.1`).

## Additional context

This is arguably a knock-on bug: the Sigstore bundle in question was generated by `sigstore-python` in the `1.x` series, which didn't include the `checkpoint` field in its bundles (since it wasn't clear from the `v0.1` bundle spec that it was required). 

I'm tracking the associated behavior in sigstore-python here: https://github.com/sigstore/sigstore-python/issues/1088

Hayden-IO:
#### Summary

It was assumed that if an inclusion proof is present, then a checkpoint would be present too. Some bundles exist where the checkpoint was not present, particularly for v0.1. This is a bug, so we should err out, but we shouldn't crash if the checkpoint is missing.

Fixes https://github.com/sigstore/sigstore-go/issues/259

#### Release Note

Fixed crash when log checkpoint was missing

### DIFF
--- a/pkg/bundle/bundle_test.go
+++ b/pkg/bundle/bundle_test.go
@@ -25,7 +25,6 @@ import (
 
 	protobundle "github.com/sigstore/protobuf-specs/gen/pb-go/bundle/v1"
 	protocommon "github.com/sigstore/protobuf-specs/gen/pb-go/common/v1"
-	v1 "github.com/sigstore/protobuf-specs/gen/pb-go/common/v1"
 	rekorv1 "github.com/sigstore/protobuf-specs/gen/pb-go/rekor/v1"
 	_ "github.com/sigstore/rekor/pkg/types/hashedrekord"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


steiza (approved):
LGTM - thanks!

woodruffw:
Thanks @haydentherapper!

PR #271, opened 2024-08-14
Add deprecation message for ProtobufBundle

codysoyland:
#### Summary

In https://github.com/sigstore/sigstore-go/pull/251, we renamed `ProtobufBundle` to `Bundle`. This PR adds a type alias and constructor with [deprecation messages](https://go.dev/wiki/Deprecated) so that users will know to rename their calling code to use the new type names when upgrading. Many tools and IDEs will show deprecation messages.

Example in VS Code:
image


Hayden-IO (approved):
Thanks!

PR #277, opened 2024-08-19
Verify certificate validity with only current time, bump conformance tests

### LINKED ISSUE #276: Certificate time is always valid when no timestamp is provided
Following up from https://github.com/sigstore/sigstore-go/issues/193#issuecomment-2297158885, we are using the [certificate's timestamp to verify itself](https://github.com/sigstore/sigstore-go/blob/d19dc5bd33f20e43bf49eaafcaeded0173c6b526/pkg/verify/signed_entity.go#L733-L738). This will always pass the time verification because a certificate's "not before" time always falls in the range of "not before" to "not after".

Instead, we should only use current time to verify a certificate's validity. I've started making the change, but some tests are failing, so I wanted to check if I'm missing something and this behavior is what you had expected from the policy flag?

@codysoyland @steiza @cmurphy

Hayden-IO:
Using a certificate's NBF will always pass the time verification. We should be using only the current time to try to verify a certificate's validity. This is likely to only work with long-lived certificates or where verification happens immediately after signing.

Updated conformance testing to drop detached testing. 

Fixes #276

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -42,11 +42,11 @@ jobs:
 
       - run: go build -o conformance cmd/conformance/main.go
 
-      - uses: sigstore/sigstore-conformance@6bd1c54e236c9517da56f7344ad16cc00439fe19 # v0.0.13
+      - uses: sigstore/sigstore-conformance@b0635d4101f11dbd18a50936568a1f7f55b17760 # v0.0.14
         with:
           entrypoint: ${{ github.workspace }}/conformance
 
-      - uses: sigstore/sigstore-conformance@6bd1c54e236c9517da56f7344ad16cc00439fe19 # v0.0.13
+      - uses: sigstore/sigstore-conformance@b0635d4101f11dbd18a50936568a1f7f55b17760 # v0.0.14
         with:
           entrypoint: ${{ github.workspace }}/conformance
           environment: staging
--- a/cmd/conformance/main.go
+++ b/cmd/conformance/main.go
@@ -15,10 +15,7 @@
 package main
 
 import (
-	"crypto/sha256"
-	"encoding/base64"
 	"encoding/hex"
-	"encoding/pem"
 	"fmt"
 	"log"
 	"os"
@@ -26,7 +23,6 @@ import (
 	"time"
 
 	protobundle "github.com/sigstore/protobuf-specs/gen/pb-go/bundle/v1"
-	protocommon "github.com/sigstore/protobuf-specs/gen/pb-go/common/v1"
 	"github.com/theupdateframework/go-tuf/v2/metadata/fetcher"
 	"google.golang.org/protobuf/encoding/protojson"
 
@@ -39,19 +35,15 @@ import (
 )
 
 var bundlePath *string
-var certPath *string
 var certOIDC *string
 var certSAN *string
 var identityToken *string
-var signaturePath *string
 var staging = false
 var trustedRootPath *string
 
 func usage() {
 	fmt.Println("Usage:")
-	fmt.Printf("\t%s sign --identity-token TOKEN --signature FILE --certificate FILE [--staging] FILE", os.Args[0])
 	fmt.Printf("\t%s sign-bundle --identity-token TOKEN --bundle FILE [--staging] FILE", os.Args[0])
-	fmt.Printf("\t%s verify --signature FILE --certificate FILE --certificate-identity IDENTITY --certificate-oidc-issuer URL [--trusted-root FILE] [--staging] FILE\n", os.Args[0])
 	fmt.Printf("\t%s verify-bundle --bundle FILE --certificate-identity IDENTITY --certificate-oidc-issuer URL [--trusted-root FILE] [--staging] FILE\n", os.Args[0])
 }
 
@@ -100,9 +92,6 @@ func parseArgs() {
 		case "--bundle":
 			bundlePath = &os.Args[i+1]
 			i += 2
-		case "--certificate":
-			certPath = &os.Args[i+1]
-			i += 2
 		case "--certificate-oidc-issuer":
 			certOIDC = &os.Args[i+1]
 			i += 2
@@ -111,9 +100,6 @@ func parseArgs() {
 			i += 2
 		case "--identity-token":
 			identityToken = &os.Args[i+1]
-			i += 2
-		case "--signature":
-			signaturePath = &os.Args[i+1]
 			i += 2
 		case "--staging":
 			staging = true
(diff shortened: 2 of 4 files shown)


Hayden-IO:
@woodruffw Do you know if we have any conformance tests around verifying certificates against current time?

Hayden-IO:
Well, right after I wrote that, a conformance test failed, so I guess yes.

Hayden-IO:
I think the issue is we're conflating two things in sigstore-go - verification of the certificate and that a signing event occurred when the certificate is valid.

In Cosign, the way this was dealt with is that the certificate is always verified with [`NotBefore`](https://github.com/sigstore/cosign/blob/8a4f39046605e0072cda5da67a457fcb57b5e767/pkg/cosign/verify.go#L1337), and then later when we compare the time of signing against the certificate, we use [current time](https://github.com/sigstore/cosign/blob/8a4f39046605e0072cda5da67a457fcb57b5e767/pkg/cosign/verify.go#L792-L802) as a fallback.

I think we should do the same here. This mirrors what sigstore-python does (I assume other clients as well given this is in the conformance suite) - sigstore-python uses [NBF](https://github.com/sigstore/sigstore-python/blob/main/sigstore/verify/verifier.py#L160) to verify the certificate with a later check against [provided timestamps](https://github.com/sigstore/sigstore-python/blob/main/sigstore/verify/verifier.py#L209).

Hayden-IO:
The other question is what we want this fallback behavior to be. If we require current time to be used as a fallback, that would mean that you MUST provide a signed timestamp for short lived certs, which I think is the behavior we want. If the fallback would use NBF, then we are effectively skipping the timestamp check for a certificate - we do say this is an unsafe behavior as per the policy flag. I’d like to hear more from clients if we want to have this - it seems only useful for testing, which I would rather not have policy flags for just that.

Hayden-IO:
Another question - how should conformance pass when given only a signature and certificate? I guess all clients are using NBF, because otherwise it wouldn’t be possible to verify a short lived cert?

woodruffw:
> @woodruffw Do you know if we have any conformance tests around verifying certificates against current time?

I don't think we do yet, unfortunately -- I believe everything in the conformance suite currently verifies using the signed time source when present (or the client is expected to retrieve it, if not present). But the latter isn't enforceable, so in theory some clients could be allowing verification to succeed based on NBF being correct.

> how should conformance pass when given only a signature and certificate? I guess all clients are using NBF, because otherwise it wouldn’t be possible to verify a short lived cert?

At least for sigstore-python, we currently perform an online lookup if we're given only detached materials, so we *don't* use just the `notBefore` time -- we use whatever signed time we get back from the transparency service. But I'm not sure about other clients 🙂

Hayden-IO:
In chatting with some other clients and thinking more on this, there are a few cases to handle:

* Verifying short-lived certificates as part of conformance tests
* Verifying short-lived certificates when no timestamp is presented
* Verifying long-lived certificates (not a part of Sigstore's spec, but for bring-your-own PKI or modified Fulcio instances)

For the first and second, I think the right solution is to require a signed timestamp. For conformance tests, either

1. we need to keep the current code to include a policy to skip the timestamp check (exposing a significant footgun to typical users)
2. don't integrate with the detached material conformance test (maybe this is fine, the bundle tests should be sufficient)
3. follow what sigstore-java is doing and fetch an SET on demand during conformance to get a signed timestamp

For the third, this is where I think we should fallback to current time when no time is provided. Either we could remove `WithoutAnyObserverTimestampsUnsafe` and fallback by default, or rename/alias this to `WithCurrentTime` since it isn't unsafe as long as it is doing some timestamp check.

@steiza @codysoyland is there a use case for the `WithoutAnyObserverTimestampsUnsafe` policy flag that I'm unaware of?

kommendorkapten:
This is great @haydentherapper!

I'm not aware of any use cases for `WithoutAnyObserverTimestampsUnsafe`. And my reasoning is aligned with yours, if you are using "classical" long-lived certificates, you do want to verify against current time, as that's more of a typical verification flow when using certificates.

Replacing `WithoutAnyObserverTimestampsUnsafe` with a new policy `WithCurrentTime` is a good idea, as it's explicit on what it does, and does not offer a method to bypass a timestamp check.

steiza:
There's a lot here! Let me try to walk through the [Sigstore client spec](https://docs.google.com/document/d/1kbhK2qyPPk8SLavHzYSDM8-Ueul9_oxIMVFuWMWKz0E/edit?usp=sharing), conformance tests, and then we can talk about what we should do here.

From the Sigstore client spec, I think there was some misunderstanding when we were implementing sigstore-go as to what was meant under "Establishing a Time for the Signature":

> The Verifier MUST perform certification path validation ([RFC 5280 §6](https://datatracker.ietf.org/doc/html/rfc5280#section-6)) of the certificate chain with the pre-distributed Fulcio root certificate(s) as a trust anchor, but with a fake “current time.” If a timestamp from the timestamping service is available, the Verifier MUST perform path validation using the timestamp from the Timestamping Service. If a timestamp from the Transparency Service is available, the Verifier MUST perform path validation using the timestamp from the Transparency Service. If both are available, the Verifier performs path validation twice. If either fails, verification fails.

At the very least, we should probably call out here what to do if there's no timestamp from the timestamping service or the transparency service. Some options for sigstore-go to consider:

- Attempt to query the transparency service online (it sounds like this is what sigstore-python does). This would allow the conformance test to pass, but maybe we want to move away from online lookups from the transparency service?

- Fail (which is what this pull request does). This isn't a bad option, although we should probably update the conformance test to not use detached materials ([which we want to do anyways](https://github.com/sigstore/sigstore-conformance/issues/122)). 

- Fall back to ensuring the current time just matches the certificate's NBF. If we decide to go down this route, we should definitely update the client specification.

It's not clear to me what the next step is here. Maybe we should agree on what the behavior should be, make any necessary updates to the client specification and / or conformance tests, and then update the client behavior?

I'm leaning towards "fail", although it'll make sigstore-go harder to use with detached materials, which will exist for a while even as we move away from them. I think we can work around it by ensuring cosign does online transparency service queries when constructing a bundle to verify.

woodruffw:
> I'm leaning towards "fail", although it'll make sigstore-go harder to use with detached materials, which will exist for a while even as we move away from them. I think we can work around it by ensuring cosign does online transparency service queries when constructing a bundle to verify.

Big +1 for this approach: this is what we settled on for sigstore-python, as a compromise between loosening the client's default timeliness guarantees and requiring an online lookup as part of verification for detached materials.

kommendorkapten:
> At the very least, we should probably call out here what to do if there's no timestamp from the timestamping service or the transparency service

This is interesting, I need to refresh my memory on the client spec. But this is the scenario when neither a RFC 3161 signed timestamp nor a SET exists. I'm tempted to update the client spec to explicitly state that the verification should fail. As the client spec should define the behaviour of a default sigstore client.

But that does not mean that sigstore-go or other client's does not allow for policies like `WithCurrentTime` as that can be useful for certain deployments like BYOPKI where more classical X.509 PKI is used.

Hayden-IO:
It sounds like we're all in agreement! So to confirm:

1. We will update the client spec to be explicit that at least one signed timestamp must be provided by default (https://github.com/sigstore/architecture-docs/issues/15). (I also added https://github.com/sigstore/architecture-docs/issues/16 to document that a client should not fail on individual timestamp verification, that it's best effort to find as many timestamps as possible - if you disagree, feel free to chime in there)
2. This PR will be updated to have a `WithCurrentTime` policy flag, and fail if no signed timestamp is verified.
3. I will update conformance testing for detached tests to fetch a proof, like what sigstore-python does. We will remove detached tests entirely as a separate discussion.

> I'm leaning towards "fail", although it'll make sigstore-go harder to use with detached materials, which will exist for a while even as we move away from them. I think we can work around it by ensuring cosign does online transparency service queries when constructing a bundle to verify.

If we did want to keep detached conformance testing, what about if we update the conformance test to also support providing a proof as detached material? Cosign already persists this. If online queries are only for conformance testing, that's fine-ish (this is what sigstore-python does, as William told me), but a) we should strongly discourage online lookups and I'd love to just remove online lookups from clients entirely, and b) tile-based logs remove the ability to query by hash, so we'll need to move towards proofs as detached material regardless (or remove detached conformance tests, so this would be moot).

steiza:
> If we did want to keep detached conformance testing, what about if we update the conformance test to also support providing a proof as detached material?

I'd rather just remove detached materials from the conformance tests. If we add a flag to https://github.com/sigstore/sigstore-conformance/blob/main/docs/cli_protocol.md#verify to provide a proof separately, then all the clients have to update in order to support that new flag. Let's start moving away from detached materials!

Hayden-IO:
Ready for a review finally! Thank you @steiza for updating the conformance tests!

steiza (approved):
I think this is good to go!

There might be some places where people were using `WithoutAnyObserverTimestampsUnsafe()` as part of their unit testing and might have to update their tests (like you did for `signed_entity_test.go`) - should be fine though.

kommendorkapten (approved):

PR #283, opened 2024-08-28
Fix proof of key possession generation

### LINKED ISSUE #282: Proof of key possession is inconsistent with Fulcio
**Description**

sigstore-go [computes the proof of key possession signature over the token's subject](https://github.com/sigstore/sigstore-go/blob/79a4e821ad2deabb34a13b8ae6a271abdc725a6c/pkg/sign/certificate.go#L131-L135), while Fulcio verifies it using the token's email address. This results in the signing flow (using the public good instance) failing:

```
Error: Fulcio returned 400: {"code":3,"message":"The signature supplied in the request could not be verified","details":[]}
```

Related: https://github.com/sigstore/cosign/issues/3777

**Version**

v0.5.1

adityasaky:
#### Summary

This commit updates the proof of key possession signature to prioritize email over subject when the claim is present in the token. This matches the current behaviour of Fulcio, which verifies the proof signature using the token's email claim.

Closes #282 

#### Release Note

Updated proof of key possession signature to use email when it's present in the token.

#### Documentation

NONE


adityasaky:
I've copied `subjectFromClaims` from sigstore/sigstore for the time being. We can't currently use the exported `SubjectFromToken` function because it accepts an `oidc.IDToken` object which is only created after the token is verified by the client. This library doesn't currently verify the token, leaving that to Fulcio. FWIW, `SubjectFromToken` operates on the token's bytes (tok.Claims() is a wrapper around json.Unmarshal of the token bytes). However, changing it instead to just use the token's bytes would be a breaking change, and I also haven't checked whether the token bytes are still available when the function is invoked. I suppose another question is whether cosign, etc. should continue to verify the token at all or just leave it to Fulcio as this library does.

pkg/sign/certificate.go
```diff
@@ -220,3 +241,27 @@ func (f *Fulcio) GetCertificate(ctx context.Context, keypair Keypair, opts *Cert
...
+// signature.
+//
+// FIXME: this is taken from sigstore/sigstore. We don't currently use the
+// exported SubjectFromToken function because it requires an *oidc.IDToken
+// object, which is generated after the token is verified by the client.
+// In this library, the token is not verified as Fulcio also verifies the
+// token's signature anyway. Also, sigstore/sigstore's implementation of
+// SubjectFromToken operates on the token's bytes (tok.Claims() wraps around
+// json.Unmarshal), so possibly this change belongs there.
+func subjectFromClaims(c jsonWebToken) (string, error) {
```
Hayden-IO: I'd be OK with a second method in sigstore/sigstore, something like `SubjectFromUnverifiedToken`.

@woodruffw I feel like this came up before in sigstore-python?

pkg/sign/certificate.go
```diff
@@ -220,3 +241,27 @@ func (f *Fulcio) GetCertificate(ctx context.Context, keypair Keypair, opts *Cert
...
+	if c.Email != "" {
+		if !c.Verified {
+			return "", errors.New("not verified by identity provider")
```
steiza: Would it make sense to fall through to check `c.Sub` in this case? I could see an argument either way.

steiza (commented):
I think this approach makes sense. Could we add some unit tests?

adityasaky:
> Could we add some unit tests?

Happy to! I held off until there was consensus on whether `subjectFromClaims` should in fact be in this repo or sigstore/sigstore. Also, I think we can't test the proof of key possession signature itself, since fulcio is mocked?

pkg/sign/certificate.go
kommendorkapten: What are the scenarios where an ID Token contains an (unverified) email but we wish to use the subject instead? I feel like require email + verified for email, but fail if only email is provided would result in a more consistent behaviour with less surprises? I.e I have a slight preference to keep as is.

adityasaky:
(apologies for all the force pushes)

I think this is now ready for review, I've updated sigstore/sigstore to v1.8.9 that @haydentherapper just cut, which includes `SubjectFromUnverifiedToken` discussed in https://github.com/sigstore/sigstore-go/pull/283#discussion_r1735246780.

pkg/sign/certificate.go
adityasaky: Added in https://github.com/sigstore/sigstore/pull/1826

Hayden-IO (dismissed):
thanks!

adityasaky:
Didn't realize there was another go.mod, I've updated it as well now.

kommendorkapten (approved):
👍

steiza (approved):

PR #288, opened 2024-09-06
Allow parsing of certificates from Fulcio if ctlog is disabled

### LINKED ISSUE #287: Signing fails when Fulcio has ctlog disabled
**Description**

When signing with the ctlog disabled, sigstore-go returns the error message `Fulcio returned no certificates`.

**Version**

codysoyland:
#### Summary

Currently, signing will fail if Fulcio has the ctlog disabled. This patch fixes the parsing of the Fulcio response so it allows certificates without an embedded SCT.

Fixes https://github.com/sigstore/sigstore-go/issues/287


pkg/sign/certificate.go
```diff
@@ -72,13 +72,18 @@ type publicKey struct {
 }
 
 type fulcioResponse struct {
-	SctCertWithChain signedCertificateEmbeddedSct `json:"signedCertificateEmbeddedSct"`
+	SignedCertificateEmbeddedSct signedCertificateEmbeddedSct `json:"signedCertificateEmbeddedSct"`
+	SignedCertificateDetachedSct signedCertificateDetachedSct `json:"signedCertificateDetachedSct"`
```
Hayden-IO, line 76: We want to deprecate this in fulcio, no other client is using it so far. 

Can we avoid adding this?

Hayden-IO (changes requested):

pkg/sign/certificate.go
codysoyland: GitHub Artifact Attestations does not use certificate transparency for private CA, using essentially `fulcio serve --ct-log-url=""`, and when configured this way, Fulcio's response includes the certificate in `signedCertificateDetachedSct`. So this change is currently required to support GitHub's private CA.

Is the deprecation plan in Fulcio to disallow running Fulcio without a CT log?

kommendorkapten: I would parse this as detached SCT is going to be deprecated, which seems fair?

codysoyland: I located the [Fulcio issue](https://github.com/sigstore/fulcio/issues/1499) referencing this deprecation, and I noticed a [comment from @bdehamer](https://github.com/sigstore/fulcio/issues/1499#issuecomment-1863667790) (with my name cc'd no less 😆 ):

> One thing to note . . . we're running a private Fulcio instance w/ NO transparency log. In this mode, I noticed that the response from the create-signing-certificate API always uses the signedCertificateDetachedSct field in the response (despite the fact that there is no SCT). As the code is cleaned-up we'll want to make sure not to change this behavior so we don't break existing clients

This is the behavior I'm wanting to preserve/support in this PR -- not add support for detached SCTs in sigstore-go. I'm open to deprecating that JSON key if we want to improve naming the Fulcio response, but for now, this key is required for us to consume Fulcio certs in GitHub's Fulcio deployment.

@haydentherapper, btw, you [commented in that issue](https://github.com/sigstore/fulcio/issues/1499#issuecomment-1874652031) that the deprecation would preserve that field for this use case.

Hayden-IO: Sorry for taking so long to get back to you - yep, you're right, I was thinking we were writing the certificate in embedded still and that this field was only used when issuing a detached SCT. This LGTM, we'll make breaking changes as part of a 2.0 release that fixes this, but no changes now.

pkg/sign/certificate_test.go
```diff
@@ -56,7 +56,7 @@ func getFulcioResponse() (*http.Response, error) {
 	}))
 
 	responseStruct := fulcioResponse{
-		SctCertWithChain: signedCertificateEmbeddedSct{
+		SignedCertificateEmbeddedSct: signedCertificateEmbeddedSct{
 			Chain: chain{
 				Certificates: []string{certPEM},
 			},
```
Hayden-IO, line 65: Can we add a test?

Hayden-IO (dismissed):

steiza (approved):
Looks good to me! I think it's a good idea to support signing with private deployments where Fulcio might not have a CT log.

PR #296, opened 2024-09-26
feat: add unit test for online tlog verification

### LINKED ISSUE #53: Add unit tests for online log verification and inclusion proofs

vishal-chdhry:
#### Summary

Closes: #53
Adds test for online verification of artifact transparency log using mock rekor

### DIFF
--- a/pkg/testing/ca/ca.go
+++ b/pkg/testing/ca/ca.go
@@ -36,6 +36,7 @@ import (
 	"github.com/cyberphone/json-canonicalization/go/src/webpki.org/jsoncanonicalizer"
 	"github.com/digitorus/timestamp"
 	"github.com/go-openapi/runtime"
+	"github.com/go-openapi/swag"
 	"github.com/secure-systems-lab/go-securesystemslib/dsse"
 	"github.com/sigstore/rekor/pkg/generated/models"
 	"github.com/sigstore/rekor/pkg/pki"
@@ -43,6 +44,7 @@ import (
 	"github.com/sigstore/rekor/pkg/types/hashedrekord"
 	"github.com/sigstore/rekor/pkg/types/intoto"
 	"github.com/sigstore/rekor/pkg/types/rekord"
+	"github.com/sigstore/rekor/pkg/util"
 	"github.com/sigstore/sigstore-go/pkg/bundle"
 	"github.com/sigstore/sigstore-go/pkg/root"
 	"github.com/sigstore/sigstore-go/pkg/tlog"
@@ -193,7 +195,7 @@ func (ca *VirtualSigstore) AttestAtTime(identity, issuer string, envelopeBody []
 		return nil, err
 	}
 
-	entry, err := ca.GenerateTlogEntry(leafCert, envelope, sig, integratedTime.Unix())
+	entry, err := ca.GenerateTlogEntry(leafCert, envelope, signer, sig, integratedTime.Unix())
 	if err != nil {
 		return nil, err
 	}
@@ -245,7 +247,7 @@ func (ca *VirtualSigstore) SignAtTime(identity, issuer string, artifact []byte, 
 	}, nil
 }
 
-func (ca *VirtualSigstore) GenerateTlogEntry(leafCert *x509.Certificate, envelope *dsse.Envelope, sig []byte, integratedTime int64) (*tlog.Entry, error) {
+func (ca *VirtualSigstore) GenerateTlogEntry(leafCert *x509.Certificate, envelope *dsse.Envelope, signer *signature.ECDSASignerVerifier, sig []byte, integratedTime int64) (*tlog.Entry, error) {
 	leafCertPem, err := cryptoutils.MarshalCertificateToPEM(leafCert)
 	if err != nil {
 		return nil, err
@@ -284,7 +286,22 @@ func (ca *VirtualSigstore) GenerateTlogEntry(leafCert *x509.Certificate, envelop
 		return nil, err
 	}
 
-	return tlog.NewEntry(rekorBodyRaw, integratedTime, logIndex, rekorLogIDRaw, set, nil)
+	// TODO: what will be the root hash?
+	rootHash := ""
+	scBytes, err := util.CreateAndSignCheckpoint(context.TODO(), "rekor.localhost", int64(123), uint64(42), []byte(rootHash), signer)
+	if err != nil {
+		return nil, err
+	}
+	inclusionProof := &models.InclusionProof{
+		TreeSize: swag.Int64(int64(2)),
+		RootHash: swag.String(rootHash),
+		LogIndex: swag.Int64(1),
+		Hashes: []string{
+			"", // TODO: ???
+		},
+		Checkpoint: swag.String(string(scBytes)),
+	}
+	return tlog.NewEntry(rekorBodyRaw, integratedTime, logIndex, rekorLogIDRaw, set, inclusionProof)
 }
 
 func (ca *VirtualSigstore) generateTlogEntryHashedRekord(leafCert *x509.Certificate, artifact []byte, sig []byte, integratedTime int64) (*tlog.Entry, error) {
@@ -451,6 +468,7 @@ func (ca *VirtualSigstore) RekorLogs() map[string]*root.TransparencyLog {
 		ValidityPeriodEnd:   time.Now().Add(time.Hour),
 		HashFunc:            crypto.SHA256,
 		PublicKey:           ca.rekorKey.Public(),
+		SignatureHashFunc:   crypto.SHA256,
 	}
 	return verifiers
 }
--- a/pkg/testing/ca/ca.go
+++ b/pkg/testing/ca/ca.go
@@ -158,10 +158,16 @@ func (ca *VirtualSigstore) GenerateLeafCert(identity, issuer string) (*x509.Cert
 func (ca *VirtualSigstore) Attest(identity, issuer string, envelopeBody []byte) (*TestEntity, error) {
 	// The timing here is important. We need to attest at a time when the leaf
 	// certificate is valid, so we match what GenerateLeafCert() does, above
-	return ca.AttestAtTime(identity, issuer, envelopeBody, time.Now().Add(5*time.Minute))
+	return ca.AttestAtTime(identity, issuer, envelopeBody, time.Now().Add(5*time.Minute), false)
 }
 
-func (ca *VirtualSigstore) AttestAtTime(identity, issuer string, envelopeBody []byte, integratedTime time.Time) (*TestEntity, error) {
+func (ca *VirtualSigstore) AttestWithInclusionProof(identity, issuer string, envelopeBody []byte) (*TestEntity, error) {
+	// The timing here is important. We need to attest at a time when the leaf
+	// certificate is valid, so we match what GenerateLeafCert() does, above
+	return ca.AttestAtTime(identity, issuer, envelopeBody, time.Now().Add(5*time.Minute), true)
+}
+
+func (ca *VirtualSigstore) AttestAtTime(identity, issuer string, envelopeBody []byte, integratedTime time.Time, generateInclusionProof bool) (*TestEntity, error) {
 	leafCert, leafPrivKey, err := ca.GenerateLeafCert(identity, issuer)
 	if err != nil {
 		return nil, err
@@ -195,7 +201,7 @@ func (ca *VirtualSigstore) AttestAtTime(identity, issuer string, envelopeBody []
 		return nil, err
 	}
 
-	entry, err := ca.GenerateTlogEntry(leafCert, envelope, sig, integratedTime.Unix())
+	entry, err := ca.GenerateTlogEntry(leafCert, envelope, sig, integratedTime.Unix(), generateInclusionProof)
 	if err != nil {
 		return nil, err
 	}
@@ -247,7 +253,7 @@ func (ca *VirtualSigstore) SignAtTime(identity, issuer string, artifact []byte, 
 	}, nil
 }
 
-func (ca *VirtualSigstore) GenerateTlogEntry(leafCert *x509.Certificate, envelope *dsse.Envelope, sig []byte, integratedTime int64) (*tlog.Entry, error) {
+func (ca *VirtualSigstore) GenerateTlogEntry(leafCert *x509.Certificate, envelope *dsse.Envelope, sig []byte, integratedTime int64, generateInclusionProof bool) (*tlog.Entry, error) {
 	leafCertPem, err := cryptoutils.MarshalCertificateToPEM(leafCert)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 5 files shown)


vishal-chdhry:
@cmurphy @haydentherapper I tried to add mock test by creating mock Rekor Entry client and adding inclusion proof to GenerateTlogEntry but I could not figure out how to:
1. generate the correct inclusion proof checkpoint and,
2. generate hashes required to compute the inclusion proof

It kept on failing while verifying [rekorLogEntry](https://github.com/sigstore/sigstore-go/blob/5845298281875c153f40da081466ca716ff1abc6/pkg/verify/tlog.go#L148). I don't know enough about rekor internals to create them properly. Meanwhile I have created a test that verifies one of the examples using sigstore public rekor.

vishal-chdhry:
The test that uses public rekor is [passing](https://github.com/sigstore/sigstore-go/actions/runs/11051057941/job/30700146225?pr=296#step:5:27)

I have also added the unfinished test to the PR, What I don't understand is
1. What will be the root hash
2. How to calculate the "hashes"
3. Right way to calculate checkpoints

cmurphy:
> What will be the root hash

The CT RFC has an explainer on merkle tree hash calculations: https://www.rfc-editor.org/rfc/rfc6962.html#section-2.1

The simplest merkle tree will contain only a single leaf, and the root hash will be the SHA-256 hash of the leaf prepended by a null character. The inclusion proof won't need to include additional hashes because there is only one hash needed to verify the leaf's inclusion in the log: https://gist.github.com/cmurphy/1b1203f4fec5b2363c30901d4684c2c4#file-treesize1-go

I think simplifying `GenerateTlogEntry` to make the tree only have one node is probably fine. The unit test doesn't need to validate the transparency-dev merkle library, just the sigstore-go usage of it.

If you did need to generate a more complex tree, the RFC explains how the root hash is calculated by concatenating the root hashes of its subtrees, and it would look something like this: https://gist.github.com/cmurphy/1b1203f4fec5b2363c30901d4684c2c4#file-treesize2-go

> The test that uses public rekor is passing

It would be better if the unit tests did not make real network calls to the public infrastructure. Unit tests should be self-contained and never rely on external services, because:

1. If the external service is down or lagging, the CI running the unit tests will also slow down
2. Unit tests are run often and will significantly add to the network traffic hitting the public infrastructure, which could artificially degrade the service or increase infrastructure costs
3. Developers may want to run unit tests on a machine not connected to the internet

vishal-chdhry:
Thank you so much for the amazing explanation @cmurphy, I learned a lot about how tlogs actually work.

I fixed inclusion and checkpoint but I still had this problem where SET verification were failing. I noticed that while verifying SETs in offline mode we are converting the log IDs to hex:
https://github.com/sigstore/sigstore-go/blob/5845298281875c153f40da081466ca716ff1abc6/pkg/tlog/entry.go#L267-L272

But the rekor library does not do that https://github.com/sigstore/rekor/blob/8ca864f6b2bcece1b0b9b004938375bbe8fff03c/pkg/verify/verify.go#L190-L195

When I convert logID to hex in my mock log entry anon, SET verification using rekor library starts to pass, see: [`50f5a75` (#296)](https://github.com/sigstore/sigstore-go/pull/296/commits/50f5a75f20c387e2e150a4b27bd58eee7b627e26)

Is there a reason why we do that? I don't understand why I should convert it to hex in my mock to make it work

cmurphy:
> Is there a reason why we do that? I don't understand why I should convert it to hex in my mock to make it work

It looks like rekor stores the LogID of the log entry as hex encoded:

https://github.com/sigstore/rekor/blob/1e6728717d221a87792a8fc0c4cf588cc31dceb0/pkg/api/entries.go#L92
https://github.com/sigstore/rekor/blob/1e6728717d221a87792a8fc0c4cf588cc31dceb0/pkg/api/api.go#L177

It doesn't need to re-encoded it during verification. But the mock should ensure the log entry is encoded the way the verifier expects it to be.

vishal-chdhry:
okay, makes sense now. 

I have hex encoded the log entry and the PR is ready for review

pkg/verify/tlog_test.go
```diff
@@ -221,3 +228,68 @@ func TestMaxAllowedTlogEntries(t *testing.T) {
...
+	return nil, errors.New("not implemented")
+}
+
+func (m *mockEntriesClient) SearchLogQuery(_ *entries.SearchLogQueryParams, _ ...entries.ClientOption) (*entries.SearchLogQueryOK, error) {
+	return nil, errors.New("not implemented")
+}
+
+func (m *mockEntriesClient) SetTransport(_ runtime.ClientTransport) {}
+
+func TestTlogVerification(t *testing.T) {
```
cmurphy: Could this be renamed to make it clear it's testing online verification:
```suggestion
func TestOnlineVerification(t *testing.T) {
```

pkg/tlog/entry.go
```diff
@@ -248,6 +249,9 @@ func (entry *Entry) HasInclusionPromise() bool {
 func (entry *Entry) HasInclusionProof() bool {
 	return entry.logEntryAnon.Verification != nil
 }
+func (entry *Entry) Verification() *models.LogEntryAnonVerification {
```
cmurphy: Is it necessary to export this function just for the tests? Could we instead construct the expected inclusion proof within the test?

pkg/testing/ca/ca.go
```diff
@@ -284,7 +286,24 @@ func (ca *VirtualSigstore) GenerateTlogEntry(leafCert *x509.Certificate, envelop
...
+		return nil, err
+	}
+	inclusionProof := &models.InclusionProof{
+		TreeSize:   swag.Int64(int64(1)),
+		RootHash:   &encodedRootHash,
+		LogIndex:   swag.Int64(0),
+		Hashes:     nil,
+		Checkpoint: swag.String(string(scBytes)),
+	}
+	return tlog.NewEntry(rekorBodyRaw, integratedTime, logIndex, rekorLogIDRaw, set, inclusionProof)
```
cmurphy: This is now including an inclusion proof for all entries, which changes the verification path for all the tests that were using this helper. Could there be a flag to opt into generating the inclusion proof, and a separate test that opts into it, so that all the existing tests continue testing the path they were testing before?

pkg/verify/tlog.go
```diff
@@ -35,14 +35,39 @@ import (
...
+
 // VerifyArtifactTransparencyLog verifies that the given entity has been logged
 // in the transparency log and that the log entry is valid.
 //
 // The threshold parameter is the number of unique transparency log entries
 // that must be verified.
 //
 // If online is true, the log entry is verified against the Rekor server.
-func VerifyArtifactTransparencyLog(entity SignedEntity, trustedMaterial root.TrustedMaterial, logThreshold int, trustIntegratedTime, online bool) ([]Timestamp, error) { //nolint:revive
+func VerifyArtifactTransparencyLog(entity SignedEntity, trustedMaterial root.TrustedMaterial, logThreshold int, trustIntegratedTime, online bool, opts ...TlogOption) ([]Timestamp, error) { //nolint:revive
```
cmurphy: This doesn't feel like the optimal way to use the Options pattern to me, it adds a lot of lines of code just to allow the test to switch to a mock client and there are no other options. Could we instead add an argument to pass in the client directly, for instance like https://github.com/sigstore/cosign/blob/677a262c3205c7bf8612f30b7b44bdf51bd68bac/internal/pkg/cosign/tsa/signer.go#L38 ?

This library is pre-1.0 so it's okay to change the function signature if there is a good reason for it and if you mention it in the release notes.

cmurphy (commented):
Could you update the PR description to reflect the current purpose of this PR and add a reference to the issue it's addressing?

pkg/verify/tlog.go
vishal-chdhry: The reason why I did this is because: 
1. `VerifyArtifactTransparencyLog` generates a new rekor client for every entry in the entities tlog, therefore one rekor client might not work if base URLs are different. 
2. Having rekor client (or a func to create rekor client) as a required argument can be confusing for users of the library, as, in my opinion, they should not be expected to figure out what client to create

what do you think?

pkg/tlog/entry.go
vishal-chdhry: If I understand correctly, we will need the rekor private key from virtual sigstore to create the signedEntryTimestamp, so I will have to export that in virtual sigstore and copy the logic for creating SET and inclusion proof. 

Since `Verification()` exports the verification material for the tlog entry, I thought it would be okay to export it

pkg/verify/tlog.go
cmurphy: Here are two options:

https://gist.github.com/cmurphy/3a4cfef7c0ee72638f93305807d6e3fe - set a global variable for the client setup function which defaults to the current function. The user doesn't have to be aware of what kind of client to create and the client is recreated within the function the same way it is now.

https://gist.github.com/cmurphy/c07fa8c0310694c77428ebe3b0162ba9 - pass the client in as a pointer argument. If set to nil, the client is created with the current function the way it is now. The user doesn't have to think about how to generate the client as long as they pass in nil, and the client is recreated if it is nil.

pkg/tlog/entry.go
```diff
@@ -89,7 +89,8 @@ func NewEntry(body []byte, integratedTime int64, logIndex int64, logID []byte, s
 
 	if inclusionProof != nil {
 		entry.logEntryAnon.Verification = &models.LogEntryAnonVerification{
-			InclusionProof: inclusionProof,
+			InclusionProof:       inclusionProof,
+			SignedEntryTimestamp: signedEntryTimestamp,
```
cmurphy: This made me curious, because if this was really needed it would point to a real bug in online verification. But it's only needed for this change because we're using `NewEntry` to generate the mock entry returned from the rekor client, when in a real scenario `NewEntry` isn't used to parse the response from rekor. I think it's possible to avoid this functional change and keep the changes constrained to just the tests, more in tlog_test.go below.

pkg/tlog/entry.go
cmurphy: I think this can be avoided, more in tlog_test.go below.

pkg/verify/tlog_test.go
```diff
@@ -221,3 +228,68 @@ func TestMaxAllowedTlogEntries(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	tlogEntry := tlogEntries[0]
+	var logEntry models.LogEntry = make(models.LogEntry)
+	logEntry["foo"] = models.LogEntryAnon{
+		Body:           tlogEntry.Body(),
+		IntegratedTime: swag.Int64(tlogEntry.IntegratedTime().Unix()),
+		LogIndex:       swag.Int64(tlogEntry.LogIndex()),
+		LogID:          swag.String(hex.EncodeToString([]byte(tlogEntry.LogKeyID()))),
+		Verification:   tlogEntry.Verification(),
```
cmurphy: Looking closer at the functional changes in entry.go that were needed to make this work, I'm seeing that this test is tautological - it's using the entity returned by the attestation as the response to the mock rekor client, and effectively comparing it to itself. I think this can be avoided, and the changes in entry.go can be discarded, if we create the mocked response from scratch with some added helpers to VirtualSigstore: https://gist.github.com/cmurphy/44a742f9f08e5cb56bc79dccfbd3b327

vishal-chdhry: @cmurphy that makes sense, thank you so much for helping with this!

pkg/verify/tlog_test.go
```diff
@@ -221,3 +227,82 @@ func TestMaxAllowedTlogEntries(t *testing.T) {
...
+
+func (m *mockEntriesClient) SetTransport(_ runtime.ClientTransport) {}
+
+func TestOnlineVerification(t *testing.T) {
+	virtualSigstore, err := ca.NewVirtualSigstore()
+	assert.NoError(t, err)
+
+	statement := []byte(`{"_type":"https://in-toto.io/Statement/v0.1","predicateType":"customFoo","subject":[{"name":"subject","digest":{"sha256":"deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef"}}],"predicate":{}}`)
+	integratedTime := time.Now().Add(5 * time.Minute)
+	entity, err := virtualSigstore.AttestAtTime("foo@example.com", "issuer", statement, integratedTime, true)
```
cmurphy, line 310: One last thing - this change adds one test, `TestOnlineVerification`, but it's possible to have inclusion proofs without verifying them online and vice versa. Could there be an additional test that tests online verification without the inclusion proof, and maybe one that tests with the inclusion proof with offline verification?

vishal-chdhry: > an additional test that tests online verification without the inclusion proof

In online verification, we call [`rekorVerify.VerifyLogEntry`](https://github.com/sigstore/rekor/blob/8ca864f6b2bcece1b0b9b004938375bbe8fff03c/pkg/verify/verify.go#L217) for all valid tlog entires which requires you to have inclusion proof see [`rekorVerify.VerifyInclusion`](https://github.com/sigstore/rekor/blob/8ca864f6b2bcece1b0b9b004938375bbe8fff03c/pkg/verify/verify.go#L142-L144)

So I am assuming this is a failure test case right?

cmurphy: > In online verification, we call rekorVerify.VerifyLogEntry for all valid tlog entires which requires you to have inclusion proof see rekorVerify.VerifyInclusion

> So I am assuming this is a failure test case right?

You could test for that failure case too, but no that's not what I meant. If you call `Attest` and get a signed entity that doesn't have an inclusion proof, then making an online call to rekor should return the inclusion proof and the online verification should succeed.

pkg/verify/tlog_test.go
```diff
@@ -221,3 +227,178 @@ func TestMaxAllowedTlogEntries(t *testing.T) {
...
+func TestOfflineInclusionProofVerification(t *testing.T) {
+	virtualSigstore, err := ca.NewVirtualSigstore()
+	assert.NoError(t, err)
+
+	statement := []byte(`{"_type":"https://in-toto.io/Statement/v0.1","predicateType":"customFoo","subject":[{"name":"subject","digest":{"sha256":"deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef"}}],"predicate":{}}`)
+	integratedTime := time.Now().Add(5 * time.Minute)
+	entity, err := virtualSigstore.AttestAtTime("foo@example.com", "issuer", statement, integratedTime, true)
+	assert.NoError(t, err)
+
+	entriesClient, err := newMockRekorEntriesClient(*virtualSigstore, statement, integratedTime)
```
cmurphy: Since this is offline there shouldn't be a need to create a mock client, that code path will not be followed

cmurphy (approved):
lgtm :+1: 

@haydentherapper @steiza

Hayden-IO (approved):
Fantastic work, this looks great!

vishal-chdhry:
Thank you @cmurphy for all the help!

PR #300, opened 2024-09-28
Add interface types for TimestampingAuthority and CertificateAuthority

### LINKED ISSUE #293: Proposal: TrustedMaterial Interface Refinement
**Description**

The `TrustedMaterial` interface currently includes these methods:

```go
type TrustedMaterial interface {
	TimestampingAuthorities() []CertificateAuthority
	FulcioCertificateAuthorities() []CertificateAuthority
	RekorLogs() map[string]*TransparencyLog
	CTLogs() map[string]*TransparencyLog
	PublicKeyVerifier(string) (TimeConstrainedVerifier, error)
}
```

The `CertificateAuthority` interface is defined as:

```go
type CertificateAuthority struct {
	Root                *x509.Certificate
	Intermediates       []*x509.Certificate
	Leaf                *x509.Certificate
	ValidityPeriodStart time.Time
	ValidityPeriodEnd   time.Time
	URI                 string
}
```

There are a couple of issues with this interface:

- Timestamp Authorities are not a type of certificate authority, they just happen to share a lot of the same fields.
- Fulcio certificate authorities don't contain a leaf certificate, so this field is always nil.
- This approach doesn't allow for flexibility of the verification interface. Currently, the verification routines in [`verify/certificate.go`](https://github.com/sigstore/sigstore-go/blob/main/pkg/verify/certificate.go) and [`verify/tsa.go`](https://github.com/sigstore/sigstore-go/blob/main/pkg/verify/tsa.go) are tightly coupled to the `CertificateAuthority` struct. In some instances, we may want to allow a client to provide a `*x509.CertPool` (for example, to cleanly migrate cosign to the new `TrustedMaterial` interface), to provide a custom certificate verifier function, or to provide multiple root certificates.

I propose the following changes:

- Add a new `TimestampingAuthority` interface.
- Add a new `CertificateAuthority` interface.
- Add structs `FulcioCertificateAuthority` and `SigstoreTimestampingAuthority` that implement the new interfaces.
- Verification logic for these types would largely move out of the `pkg/verify` package and into the `pkg/root` package.

These would be defined as follows:

```go
type TimestampingAuthority interface {
    Verify(signedTimestamp []byte, signatureBytes []byte) error
}

type SigstoreTimestampingAuthority struct {
    Root                *x509.Certificate
    Intermediates       []*x509.Certificate
    Leaf                *x509.Certificate
    ValidityPeriodStart time.Time
    ValidityPeriodEnd   time.Time
    URI                 string
}

func (sta *SigstoreTimestampingAuthority) Verify(signedTimestamp []byte, signatureBytes []byte) error {
    // verification logic here
}
```

```go
type CertificateAuthority interface {
    Verify(cert *x509.Certificate, observerTimestamp time.Time) error
}

type FulcioCertificateAuthority struct {
    Root                *x509.Certificate
    Intermediates       []*x509.Certificate
    ValidityPeriodStart time.Time
    ValidityPeriodEnd   time.Time
    URI                 string
}

func (fca *FulcioCertificateAuthority) Verify(cert *x509.Certificate, observerTimestamp time.Time) error {
    // verification logic here
}
```

The TrustedMaterial interface would then be updated to look like this:

```go
type TrustedMaterial interface {
    TimestampingAuthorities() []TimestampingAuthority
    FulcioCertificateAuthorities() []CertificateAuthority
    RekorLogs() map[string]*TransparencyLog
    CTLogs() map[string]*TransparencyLog
    PublicKeyVerifier(string) (TimeConstrainedVerifier, error)
}
```

One of the benefits of this approach is a smoother transition for cosign. This would allow cosign's `CheckOpts` to implement the `TrustedMaterial` interface, as it currently is impossible due to the use of `x509.CertPool` in cosign, which is not compatible with the current `TrustedMaterial` interface.

codysoyland:
This PR modifies the `TrustedMaterial` interface with interface types for `CertificateAuthority` and `TimestampingAuthority`. This decouples the original `CertificateAuthority` struct from the verifier and allows clients to have more control over the verification of certificates and timestamps.

Additionally, this PR modifies the verifier to output the chains of certificates that were used to verify the certificate, which has the added benefit of improving performance of the SCT verifier by only attempting SCT verification on the chain that was used to verify the certificate.

This is a breaking change to the `TrustedMaterial` interface, so clients that directly modify `TrustedMaterial` must be updated to use the new interface types.

Fixes https://github.com/sigstore/sigstore-go/issues/293


kommendorkapten (dismissed):
Nice work!

pkg/root/trusted_root.go
```diff
@@ -287,6 +275,68 @@ func ParseCertificateAuthority(certAuthority *prototrustroot.CertificateAuthorit
...
+		}
+		end := validFor.GetEnd()
+		if end != nil {
+			timestampingAuthority.ValidityPeriodEnd = end.AsTime()
+		}
+	}
+
+	timestampingAuthority.URI = certAuthority.Uri
+
+	// TODO: Should we inspect/enforce ca.Subject?
```
Hayden-IO: Can we either file issues for these TODOs or decide if these are needed?

My two cents:

* Subject is more of a human-readable value and not something we need to check. It's also the subject of the root, so there's not really anything to verify, as we implicitly trust the root.
* What is there to handle with the validity period? It seems like the above is reading it.

pkg/root/certificate_authority.go
```diff
@@ -0,0 +1,64 @@
+// Copyright 2023 The Sigstore Authors.
```
Hayden-IO: Update copyright?

pkg/root/certificate_authority.go
Hayden-IO, line 1: Can we add tests for this file and the next? Given these are verification functions, having some coverage would be good.

pkg/verify/sct.go
```diff
@@ -29,16 +30,18 @@ import (
...
-func VerifySignedCertificateTimestamp(leafCert *x509.Certificate, threshold int, trustedMaterial root.TrustedMaterial) error { // nolint: revive
+func VerifySignedCertificateTimestamp(chains [][]*x509.Certificate, threshold int, trustedMaterial root.TrustedMaterial) error { // nolint: revive
+	if len(chains) == 0 && len(chains[0]) == 0 && chains[0][0] == nil {
+		return errors.New("no chains provided")
+	}
 	ctlogs := trustedMaterial.CTLogs()
-	fulcioCerts := trustedMaterial.FulcioCertificateAuthorities()
 
-	scts, err := x509util.ParseSCTsFromCertificate(leafCert.Raw)
+	scts, err := x509util.ParseSCTsFromCertificate(chains[0][0].Raw)
```
Hayden-IO: Can we add a comment noting that chains[<any>][0] is always the leaf certificate? This just stands out as looking problematic when it should be safe.

pkg/verify/tsa.go
```diff
@@ -98,39 +84,16 @@ func VerifyTimestampAuthorityWithThreshold(entity SignedEntity, trustedMaterial
 	return verifiedTimestamps, nil
 }
 
-func verifySignedTimestamp(signedTimestamp []byte, dsseSignatureBytes []byte, trustedMaterial root.TrustedMaterial, verificationContent VerificationContent) (Timestamp, error) {
-	certAuthorities := trustedMaterial.TimestampingAuthorities()
+func verifySignedTimestamp(signedTimestamp []byte, dsseSignatureBytes []byte, trustedMaterial root.TrustedMaterial) (*root.Timestamp, error) {
```
Hayden-IO: nit on naming, this isn't always a dsse signature

pkg/verify/tsa.go
```diff
@@ -98,39 +84,16 @@ func VerifyTimestampAuthorityWithThreshold(entity SignedEntity, trustedMaterial
...
-			continue
+	for _, tsa := range timestampAuthorities {
+		ts, err := tsa.Verify(signedTimestamp, dsseSignatureBytes)
+		if err == nil {
+			return ts, nil
 		}
-
-		// Check tlog entry time against bundle certificates
-		// TODO: technically no longer needed since we check the cert validity period in the main Verify loop
-		if !verificationContent.ValidAtTime(timestamp.Time, trustedMaterial) {
```
Hayden-IO, line 127: Just wanted to check, we removed this - this was intentional? I agree that I think it's fine to remove as we should decouple certificate and timestamp verification.

Hayden-IO (changes requested):
Thanks! Just a few comments about testing and some clean up.

pkg/root/trusted_root.go
codysoyland: I agree with you on both points! The validity period comment is out of date - I think that was added prior to moving the validity check into this function.

pkg/verify/tsa.go
codysoyland: Yes, this was removed because the validity period check occurs outside this function now. 👍🏻

pkg/verify/sct.go
```diff
@@ -29,16 +30,21 @@ import (
 // leaf certificate, will extract SCTs from the leaf certificate and verify the
 // timestamps using the TrustedMaterial's FulcioCertificateAuthorities() and
 // CTLogs()
-func VerifySignedCertificateTimestamp(leafCert *x509.Certificate, threshold int, trustedMaterial root.TrustedMaterial) error { // nolint: revive
+func VerifySignedCertificateTimestamp(chains [][]*x509.Certificate, threshold int, trustedMaterial root.TrustedMaterial) error { // nolint: revive
+	if len(chains) == 0 && len(chains[0]) == 0 && chains[0][0] == nil {
```
steiza:
```suggestion
	if len(chains) == 0 || len(chains[0]) == 0 || chains[0][0] == nil {
```

`len(chains)` has to be `>= 1` for us to check `chains[0]`?

pkg/verify/signed_entity.go
```diff
@@ -564,7 +566,7 @@ func (v *SignedEntityVerifier) Verify(entity SignedEntity, pb PolicyBuilder) (*V
 		// > Unless performing online verification (see §Alternative Workflows), the Verifier MUST extract the  SignedCertificateTimestamp embedded in the leaf certificate, and verify it as in RFC 9162 §8.1.3, using the verification key from the Certificate Transparency Log.
 
 		if v.config.weExpectSCTs {
-			err = VerifySignedCertificateTimestamp(leafCert, v.config.ctlogEntriesThreshold, v.trustedMaterial)
+			err = VerifySignedCertificateTimestamp(chains, v.config.ctlogEntriesThreshold, v.trustedMaterial)
```
steiza, line 569: I love this change - that we're actually verifying the SCT against the chain the leaf cert verified with!

steiza (commented):
One small suggested change, otherwise LGTM!

pkg/verify/sct.go
codysoyland: Nice catch!

steiza (dismissed):
LGTM!

steiza (approved):
Still LGTM!

Hayden-IO (approved):
fantastic work!

PR #310, opened 2024-10-10
Check if entry has inclusion proof rather than entity

adityasaky:
#### Summary

I stumbled upon an inconsistency when verifying an entry's inclusion proof (for offline verification). I think it should be checking if the specific entry has an inclusion proof instead of going by the entity (eg. a bundle). A bundle _could_ have entries with inclusion proofs and without, and the bundle returns true for this check if even one entry has an inclusion proof, so this could lead to an error when verifying an entry without the proof.

#### Release Note

Fixed check for whether a tlog entry has an inclusion proof

#### Documentation

NONE

### DIFF
--- a/pkg/verify/tlog.go
+++ b/pkg/verify/tlog.go
@@ -105,7 +105,7 @@ func VerifyArtifactTransparencyLog(entity SignedEntity, trustedMaterial root.Tru
 					verifiedTimestamps = append(verifiedTimestamps, Timestamp{Time: entry.IntegratedTime(), URI: tlogVerifier.BaseURL})
 				}
 			}
-			if entity.HasInclusionProof() {
+			if entry.HasInclusionProof() {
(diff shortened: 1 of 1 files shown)


adityasaky:
Amending my original message: right now, this should never be a problem because of how online vs offline verification is decided. If `WithOnlineVerification()` is not used, then all entries must have the inclusion proof (meaning no mix of entries with and without the proof) or fail indicating the entry doesn't have the proof (as it currently would). If `WithOnlineVerification()` is used, we don't hit this at all. Overall, it might be okay to leave it as is?

This can change if the online vs offline control becomes a per-log control, though I don't know of contexts where that makes sense.

steiza (approved):
Oops! Yes, it was definitely my intention to use `entry` here, not `entity`. With how bundles are constructed today I don't think this ends up being much of a functionality difference, but I do think we should make this change.

codysoyland (approved):
Nice find, thank you!

PR #314, opened 2024-10-14
Add workflow to synchronize go mod changes with submodules

codysoyland:
#### Summary

This PR attempt to improve the workflow of managing Dependabot updates to this repo. The problem is that when Dependabot updates a dependency in the main module, those updates are not reflected in the "submodule" (https://github.com/sigstore/sigstore-go/tree/main/examples/oci-image-verification) which includes a reference to the main module. That causes the build to fail, and one maintainer must manually run `go mod tidy` in the submodule and push those changes, and another maintainer must approve and merge them (as the last committer is disallowed from merging a PR). ([example](https://github.com/sigstore/sigstore-go/pull/311))

This PR automatically runs `go mod tidy` in both modules and commits/pushes them whenever Dependabot creates a new PR.

There is a caveat with this, which is that the actions do not run when the last committer used `GITHUB_TOKEN` to authenticate to the Git server, which is [intentional to avoid accidentally creating recursively invoked workflows](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow). This can be worked around by using a PAT in this workflow to authenticate to the Git server, or maintainers can push an empty commit to trigger the Actions workflows.

If this seems like a good approach to fellow maintainers, I'd recommend we create a PAT for use within this repo and modify this workflow to use it.

This workflow could also be added to `sigstore/sigstore` for use with the KMS providers ([example](https://github.com/sigstore/sigstore/pull/1852)).

### DIFF
--- a/.github/workflows/synchronize-go-mod.yml
+++ b/.github/workflows/synchronize-go-mod.yml
@@ -1,0 +1,48 @@
+name: Synchronize go.mod in submodules
+
+on:
+  pull_request:
+    branches:
+      - main
+    types:
+      - opened
+      - synchronize
+
+permissions:
+  contents: write
+
+jobs:
+  synchronize-go-mod:
+    if: github.actor == 'dependabot[bot]'
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+
+      - name: Install Go
+        uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32 # v5.0.2
+        with:
+          go-version-file: go.mod
+
+      - name: Synchronize go.mod in submodules
+        run: |
+          go mod tidy
+          cd examples/oci-image-verification; go mod tidy
+
+      - name: Commit and push changes
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          if [ -n "$(git status --porcelain)" ]; then
+            echo "Changes detected:"
+            git diff
+            git config --global user.name "github-actions[bot]"
+            git config --global user.email "github-actions[bot]@users.noreply.github.com"
+            git commit -a -s -m "Synchronize go.mod in submodules"
+            git push origin HEAD:${{ github.event.pull_request.head.ref }}
+          else
+            echo "No changes to commit"
+          fi
--- a/.github/workflows/synchronize-go-mod.yml
+++ b/.github/workflows/synchronize-go-mod.yml
@@ -17,24 +17,28 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - uses: octo-sts/action@6177b4481c00308b3839969c3eca88c96a91775f # v1.0.0
+        id: octo-sts
+        with:
+          scope: sigstore/sigstore-go
+          identity: sync
+
       - name: Checkout repository
         uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
         with:
           ref: ${{ github.event.pull_request.head.ref }}
+          token: ${{ steps.octo-sts.outputs.token }}
 
       - name: Install Go
         uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32 # v5.0.2
         with:
           go-version-file: go.mod
 
-      - name: Synchronize go.mod in submodules
+      - name: Synchronize go.mod in submodule
         run: |
-          go mod tidy
           cd examples/oci-image-verification; go mod tidy
 
       - name: Commit and push changes
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           if [ -n "$(git status --porcelain)" ]; then
             echo "Changes detected:"
--- a/.github/chainguard/sync.sts.yaml
+++ b/.github/chainguard/sync.sts.yaml
@@ -1,0 +1,7 @@
+issuer: https://token.actions.githubusercontent.com
+subject: repo:sigstore/sigstore-go:pull_request
+claim_pattern:
+  actor: "dependabot\[bot\]"
+
+permissions:
+  contents: write


kommendorkapten (dismissed):
Thanks Cody! This looks good.
I think it may even work to just re-run the failing workflows from the UI instead of pushing an empty commit?

codysoyland:
> I think it may even work to just re-run the failing workflows from the UI instead of pushing an empty commit?

I don't think there's a way to trigger the workflows from the UI if they are not run due to the last commit originating from Actions -- they are not actually "failed" workflows, they just aren't created at all. Maybe I'm missing something, but I think it would require an empty commit to trigger the build, unless we add a PAT to a secret and use it to commit the go.mod changes.

@haydentherapper (or maybe @cpanato?), what do you think of incorporating this workflow and using a PAT to authenticate with git? As explained in the description, we need to use a PAT or the actions-generated commit will not trigger actions to run again, therefore the automation would be incomplete. We can create an org-level PAT secret and also use it to keep the kms providers in sigstore/sigstore up to date using the same workflow. Somebody who is org admin would need to create the PAT and upload it as an org-level secret.

Hayden-IO:
I really like this workflow, this is great! This has definitely been a nuisance here and in Sigstore/sigstore.

PATs are a little annoying to keep up to date, but we’ve already got a number that we manage, one more is fine. In the future, we can look at creating a GitHub app to replace PATs.  

what are the permissions needed for the PAT?

.github/workflows/synchronize-go-mod.yml
```diff
@@ -0,0 +1,48 @@
...
+    types:
+      - opened
+      - synchronize
+
+permissions:
+  contents: write
+
+jobs:
+  synchronize-go-mod:
+    if: github.actor == 'dependabot[bot]'
```
Hayden-IO, line 16: Should this be an ID? If someone names their account this string, this would run. Not really a big deal though

.github/workflows/synchronize-go-mod.yml
```diff
@@ -0,0 +1,48 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+
+      - name: Install Go
+        uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32 # v5.0.2
+        with:
+          go-version-file: go.mod
```
Hayden-IO, line 35: What if the modules don’t use the same version of go? Should we search for all go.mod files, then install go and run go mod tidy for each?

.github/workflows/synchronize-go-mod.yml
codysoyland: This is the [documented way](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions#responding-to-events). The actor name is a user name of the entity that triggered the workflow.

.github/workflows/synchronize-go-mod.yml
codysoyland: I hadn't considered that, but I think we'd hit an error in that case, and we'd need to manually sync up the go versions.

codysoyland:
> PATs are a little annoying to keep up to date, but we’ve already got a number that we manage, one more is fine. In the future, we can look at creating a GitHub app to replace PATs.

Yeah an app is ideal, just requires a little more setup. PATs are indeed annoying.

> what are the permissions needed for the PAT?

It can be a fine-grained permission PAT with support to the specific repos we need (sigstore, sigstore-go), with "Contents" set to "Read and write".

Hayden-IO:
@cpanato  can you create the PAT?

.github/workflows/synchronize-go-mod.yml
kommendorkapten: If we create a `go.work` and workspace, we can define the go version at the top level and use `go-version-file: go.work`, then rely on `go work sync` to propagate the version to the sub modules. That could probably be part of this workflow too.

Hayden-IO:
Coming back to this, I can create the PAT...but with that said, keeping these PATs up to date is a bit annoying as they expire. Is there any way to do this without another PAT?

codysoyland:
> Coming back to this, I can create the PAT...but with that said, keeping these PATs up to date is a bit annoying as they expire. Is there any way to do this without another PAT?

I believe we can do this with a GitHub app.... I can test it on my own org first to make sure and help you set it up if that works!

Hayden-IO:
We had also looked into https://github.com/apps/octo-sts awhile ago which Chainguard made.

cpanato:
sorry for the delay here

we can create a PAT, but i would vote to use the octo-sts, because we dont need to manage the PAT and that will be valid only for the job execution

cpanato:
and this looks soo cool!!

codysoyland:
octo-sts is awesome, I just tried it on a test repo and it works great for this purpose. I'll update the PR

Hayden-IO:
We'll need to install it in the org as well. Should mention this to the TSC before we progress.

codysoyland:
I updated the workflow to use octo-sts, so if TSC approves, we can proceed to merging this.

Hayden-IO:
@codysoyland Can you file an issue on https://github.com/sigstore/tsc asking for approval?

codysoyland:
> @codysoyland Can you file an issue on https://github.com/sigstore/tsc asking for approval?

https://github.com/sigstore/TSC/issues/51

.github/workflows/synchronize-go-mod.yml
bobcallaway: I agree with Hayden's question here - having this pinned to an immutable ID feels marginally safer.

.github/workflows/synchronize-go-mod.yml
```diff
@@ -0,0 +1,52 @@
...
+          token: ${{ steps.octo-sts.outputs.token }}
+
+      - name: Install Go
+        uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32 # v5.0.2
+        with:
+          go-version-file: go.mod
+
+      - name: Synchronize go.mod in submodule
+        run: |
+          cd examples/oci-image-verification; go mod tidy
```
bobcallaway, line 39:
```suggestion
          cd examples/oci-image-verification && go mod tidy
```

Hayden-IO:
@codysoyland is this good to close?

codysoyland:
> @codysoyland is this good to close?

Yeah... the multi-directory config I added in #380 didn't help but this is resolved for grouped updates I added in #395. It's still an issue for major release updates, but I don't think it's worth dealing with octo-sts or PATs to make this workflow work.

PR #325, opened 2024-10-29
Improve TSA verification error handling

codysoyland:
#### Summary

This PR adds more detail to the error output of `VerifyTimestampAuthority` by returning all verification errors.

### DIFF
--- a/pkg/verify/tsa.go
+++ b/pkg/verify/tsa.go
@@ -67,7 +67,7 @@ func VerifyTimestampAuthority(entity SignedEntity, trustedMaterial root.TrustedM
 		verifiedTimestamps = append(verifiedTimestamps, verifiedSignedTimestamp)
 	}
 
-	if len(verifiedTimestamps) == 0 {
+	if len(verifiedTimestamps) == 0 && len(errs) > 0 {
 		return nil, fmt.Errorf("no verified signed timestamps: %w", errors.Join(errs...))
 	}
 	return verifiedTimestamps, nil
--- a/pkg/verify/tsa.go
+++ b/pkg/verify/tsa.go
@@ -67,10 +67,13 @@ func VerifyTimestampAuthority(entity SignedEntity, trustedMaterial root.TrustedM
 		verifiedTimestamps = append(verifiedTimestamps, verifiedSignedTimestamp)
 	}
 
-	if len(verifiedTimestamps) == 0 && len(errs) > 0 {
-		return nil, fmt.Errorf("no verified signed timestamps: %w", errors.Join(errs...))
+	// Return errors if there are any, even if there are verified timestamps.
+	// This allows the caller to handle errors if the threshold is not met.
+	err = nil
+	if len(errs) > 0 {
+		err = fmt.Errorf("verifying timestamps: %w", errors.Join(errs...))
 	}
-	return verifiedTimestamps, nil
+	return verifiedTimestamps, err
 }
 
 // VerifyTimestampAuthority verifies that the given entity has been timestamped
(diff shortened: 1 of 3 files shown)


codysoyland:
There is still a test failure here, as the test expects that there is no returned error even if there are no valid timestamps (as the count of timestamps is used to test against the threshold, which may be zero). Should we update that test? Or is there a valid reason to expect that behavior?

phillmv:
>as the test expects that there is no returned error even if there are no valid timestamps (as the count of timestamps is used to test against the threshold, which may be zero). Should we update that test? Or is there a valid reason to expect that behavior?

Thinking out loud, should `VerifyTimestampAuthority` error if there are no entity.Timestamps() to verify?

On the one hand, I'm inclined to say "yes". It feels dangerous to be have a function named "Verify" that can return successfully _when it has not actually verified anything_. I probably had a hand in writing this function, so I will admit to this being a principle or vibe I only recently adopted; this behaviour got us in trouble in `gh at verify`, where we returned successfully when supplied with an empty list of attestations.

On the other hand, as currently implemented there is some ambiguity around how `VerifyObserverTimestamps` calls this function; as currently implemented, returning a zero-len array is expected behaviour. **However** I'm not a huge fan of how `VerifyObserverTimestamps` ended up & would be willing to endorse a polite refactor - do we remember why the `TlogInclusion` is verified independently of this function?

Hayden-IO (changes requested):
This was set up intentionally to support verification with multiple timestamp sources. `VerifyTimestampAuthorityWithThreshold` requires a threshold, while `VerifyTimestampAuthority` does not. `VerifyTimestampAuthority` must be used with another function that will implement a threshold, which is done [here](https://github.com/sigstore/sigstore-go/blob/main/pkg/verify/signed_entity.go#L715-L720).

If we add a baseline threshold check to `VerifyTimestampAuthority`, then verification would fail if both TSA timestamps and SETs are provided where the trust root only specifies verification material for SETs.

pkg/verify/tsa.go
```diff
@@ -54,18 +54,26 @@ func VerifyTimestampAuthority(entity SignedEntity, trustedMaterial root.TrustedM
...
 	}
 
-	return verifiedTimestamps, nil
+	// Return error value if there are any errors, even if there are verified timestamps
+	// This allows the caller to handle errors if the threshold is not met
+	err = nil
+	if len(errs) > 0 {
+		err = fmt.Errorf("verifying timestamps: %w", errors.Join(errs...))
+	}
+	return verifiedTimestamps, err
```
codysoyland: This is a bit unconventional, as an error value may be returned even if there was a successful verification. Callers should check the returned `verifiedTimestamps` even if there were errors. The errors are only populated if there was at least one failed verification.

codysoyland: Thinking about this more, we can distinguish between verification errors and other types of errors by changing the return values of `func VerifyTimestampAuthority`:
```go
func VerifyTimestampAuthority(...) ([]*root.Timestamp, error)
```
to:
```go
func VerifyTimestampAuthority(...) ([]*root.Timestamp, []error, error)
```
where the added `[]error` would represent verification errors.

This would require a public API change but would read more like conventional Go code, where a non-nil error usually means a total failure.

steiza (approved):
LGTM!

steiza:
Sorry, I thought I was on #472

pkg/verify/tsa.go
```diff
@@ -54,18 +54,26 @@ func VerifyTimestampAuthority(entity SignedEntity, trustedMaterial root.TrustedM
...
 	}
 
-	return verifiedTimestamps, nil
+	// Return errors if there are any, even if there are verified timestamps.
+	// This allows the caller to handle errors if the threshold is not met.
+	err = nil
+	if len(errs) > 0 {
+		err = fmt.Errorf("verifying timestamps: %w", errors.Join(errs...))
+	}
+	return verifiedTimestamps, err
```
Hayden-IO, line 76: Like you mentioned, this is non conventional for Go. I'm partial to what you proposed in 473.

codysoyland:
Closing in favor of https://github.com/sigstore/sigstore-go/pull/473

PR #326, opened 2024-10-31
Fix intoto statement marshal/unmarshal

gillisandrew:
This pull request addresses a bug in the JSON marshaling of the in-toto statement in verification results. It currently outputs "predicate_type" instead of "predicateType" [required by the spec](https://github.com/in-toto/attestation/blob/main/spec/v1/predicate.md). 

The issue is related to in-toto/attestation#363, this applies the fix mentioned there, defining custom marshaler/unmarshaler on the VerificationResult struct and selectively marshaling/unmarshaling the statement using google.golang.org/protobuf/encoding/protojson

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,6 @@ require (
 	github.com/go-openapi/swag v0.23.0
 	github.com/google/certificate-transparency-go v1.2.1
 	github.com/in-toto/attestation v1.1.0
-	github.com/in-toto/in-toto-golang v0.9.0
 	github.com/secure-systems-lab/go-securesystemslib v0.8.0
 	github.com/sigstore/protobuf-specs v0.3.2
 	github.com/sigstore/rekor v1.3.6
@@ -51,6 +50,7 @@ require (
 	github.com/hashicorp/go-retryablehttp v0.7.7 // indirect
 	github.com/hashicorp/golang-lru/v2 v2.0.7 // indirect
 	github.com/hashicorp/hcl v1.0.0 // indirect
+	github.com/in-toto/in-toto-golang v0.9.0 // indirect
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
 	github.com/jedisct1/go-minisign v0.0.0-20211028175153-1c139d1cc84b // indirect
 	github.com/josharian/intern v1.0.0 // indirect
--- a/pkg/bundle/signature_content.go
+++ b/pkg/bundle/signature_content.go
@@ -17,7 +17,6 @@ package bundle
 import (
 	"encoding/base64"
 
-	in_toto "github.com/in-toto/attestation/go/v1"
 	"github.com/secure-systems-lab/go-securesystemslib/dsse"
 	"github.com/sigstore/sigstore-go/pkg/verify"
 	"google.golang.org/protobuf/encoding/protojson"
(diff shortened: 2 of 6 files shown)


gillisandrew:
For clarity, here is a diff of the output from the example command:
```sh
go run cmd/sigstore-go/main.go \
  -artifact-digest 76176ffa33808b54602c7c35de5c6e9a4deb96066dba6533f50ac234f4f1f4c6b3527515dc17c06fbe2860030f410eee69ea20079bd3a2c6f3dcf3b329b10751 \
  -artifact-digest-algorithm sha512 \
  -expectedIssuer https://token.actions.githubusercontent.com \
  -expectedSAN https://github.com/sigstore/sigstore-js/.github/workflows/release.yml@refs/heads/main \
  examples/bundle-provenance.json
```
```diff
diff before.json after.json
4c4
<       "type": "https://in-toto.io/Statement/v0.1",
---
>       "_type": "https://in-toto.io/Statement/v0.1",
13c13
<       "predicate_type": "https://slsa.dev/provenance/v0.2",
---
>       "predicateType": "https://slsa.dev/provenance/v0.2",
```

phillmv:
Hrm! Thanks for this. Can we get this fixed in intoto/attestation/go, I wonder? Or is this one of those "protobuf code gen knows no gods nor masters" kind of deals?

I'm tickled that we never noticed this, but also would prefer to avoid having custom serialization funcs if we can avoid it. As a stop gap this might be fine 🤔.

codysoyland:
Oof, protojson is a pain... I don't think it can be easily fixed upstream due to how the protobuf types are compiled. One thing we can do that could simplify this though: What if we wrap `VerificationResult.Statement` in a custom type that has Unmarshall/Marshall methods, e.g.:
```go
type Statement struct{
	in_toto.Statement
}

func (s *Statement) MarshalJSON() ([]byte, error) {
	return protojson.Marshal(s.Statement)
}

func (s *Statement) UnmarshalJSON(data []byte) error {
...
}

type VerificationResult struct {
	MediaType          string                        `json:"mediaType"`
	Statement          *Statement            `json:"statement,omitempty"`
...
```

That would simplify the code here a bit.

Could you also add tests?
Thanks!

gillisandrew:
@codysoyland I started off by wrapping it but that makes it a breaking change and involves updating types all over.

I'll add some tests tomorrow or over the weekend.

kommendorkapten:
What @codysoyland proppses makes a lot of sense, I like that idea.

codysoyland:
Thank you @gillisandrew! I like this better, but before merging, I'd like to see if the in_toto maintainers would be open to adding these marshal/unmarshal methods in the upstream. Otherwise this will continue to bite other folks...

It looks like https://github.com/in-toto/attestation/issues/363 was closed erroneously... perhaps one of us can add those methods in a PR to [this file](https://github.com/in-toto/attestation/blob/main/go/v1/statement.go)? I think that would be ideal and I bet they would accept it.

gillisandrew:
@codysoyland Sounds good. I'll open one there and see what they say.

codysoyland:
I created an issue to track this: https://github.com/sigstore/sigstore-go/issues/365

codysoyland:
Closed in favor of https://github.com/sigstore/sigstore-go/pull/366 (which includes parts of this). Thanks @gillisandrew!

PR #333, opened 2024-11-15
Update BaseSignedEntity interface implementation

cmurphy:
The SignedEntity interface has gone through a few iterations. The BaseSignedEntity struct was not keeping up with it, which wasn't caught because it is never used in sigstore-go. This change updates the struct to implement the interface it was meant to implement, and adds an assignment statement to make sure future errors in updates are caught at compile time.


codysoyland (approved):
Good find, thanks for adding the interface assertion on line 97 so we catch this in the future too!

PR #350, opened 2024-12-11
Adds a check to ensure SCT time is while a CT log key was valid

### LINKED ISSUE #178: SCT verification should compare timestamp against CT log validity window

steiza:
#### Summary

Fixes #178 - we want to make sure the SCT we're verifying with a CT log key happened while the CT log key was valid.

#### Release Note

NONE

#### Documentation

N/A

### DIFF
--- a/pkg/verify/sct.go
+++ b/pkg/verify/sct.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 
+	ct "github.com/google/certificate-transparency-go"
 	"github.com/google/certificate-transparency-go/ctutil"
 	ctx509 "github.com/google/certificate-transparency-go/x509"
 	"github.com/google/certificate-transparency-go/x509util"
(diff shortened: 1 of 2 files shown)


codysoyland (approved):
I forgot about this issue... thanks for fixing!

Hayden-IO (approved):

kommendorkapten (approved):
Nice!

Issue #298, opened 2024-09-27
Add interface for certificate validation


kommendorkapten:
**Description**

Originally discussed here: https://github.com/sigstore/sigstore-go/pull/253#issuecomment-2362927648

The original implementation was made against the operational mode of PGI using ephemeral certificates and the trust root delivered via TUF. This mode of operation allows us to make the following simplifications:
* End entity (leaf) certificates expires after only ten minutes, so no method for revocation is implemented
* The Sigstore bundle only contains the end entity certificate
* All intermediates are provided via the trust root
* Long lived certificates (root and intermediates) are distributed via TUF, which allows us to perform _instant_ revocation by removing the certificate from the trust root, so no method for verifying certificate revocation is needed

However, for private deployments these preconditions may not always be met, and certain users may want to rely on a more traditional PKI, where untrusted certificates are provided in the Sigstore bundle, and CLRs or OCSP is used to verify validity of the certificate. 

To achieve required security using Certificates in Sigstore-go the client has to perform certificate verification themselves, by parsing the Sigstore bundle and then verify the certificate, which is not great UX.

Proposed is to add the support for a callback function:

```go
func ValidateCertificate(c *x509.Certifiicate) (bool, error) {
        // validate certificate
}
```

That would then called _for each_ certificate found in the bundle, including the end entity certificate.

To not complicate the logic for the default policy (using PGI), if the callback function is not set, it would set it to the a function that **always** returns true. 

### Trust root

Should the callback method be called for the certificates found in the trust root? My take is no. The trust root is created and provide by the client, and so the client could perform the certificate verification prior to composing the trust root.

codysoyland:
For the API, are you thinking this would be provided by a special `verify.VerifierOption` passed to `verify.NewSignedEntityVerifier`?

Something like this?
```go
func WithCustomCertificateValidator(validateCert func(*x509.Certificate) (bool, error)) VerifierOption {
	return func(c *VerifierConfig) error {
		c.validateCert = validateCert
	}
}
```

This could definitely be done, but I wonder if this is flexible enough, as some clients would need more control over the actual verification process in https://github.com/sigstore/sigstore-go/blob/main/pkg/verify/certificate.go -- such as modification of the `VerifyOptions` (e.g. `KeyUsages`) or other custom logic that would require access to the timestamp and trusted material.

[...]

PR #351, opened 2024-12-11
Add custom-certificate-validator example

codysoyland:
This adds an example of how to use a custom TrustedMaterial to implement a Certificate Revocation List

### DIFF
--- a/examples/certificate-revocation/crl.go
+++ b/examples/certificate-revocation/crl.go
@@ -1,0 +1,73 @@
+// Copyright 2024 The Sigstore Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"crypto/x509"
+	"fmt"
+	"math/big"
+	"time"
+
+	"github.com/sigstore/sigstore-go/pkg/root"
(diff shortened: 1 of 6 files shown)


examples/certificate-revocation/crl.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func NewTrustedMaterialWithCertificateValidation(tm root.TrustedMaterial, validator CertificateValidator) *TrustedMaterialWithCertificateValidator {
+	return &TrustedMaterialWithCertificateValidator{
+		TrustedMaterial: tm,
+		validator:       validator,
+	}
+}
+
+func (tm *TrustedMaterialWithCertificateValidator) FulcioCertificateAuthorities() []root.CertificateAuthority {
```
steiza: Minor: what if there is a signed timestamp provider that has issued erroneous certificates with known serial numbers??

Maybe that can be left as an exercise to the reader.

steiza (dismissed):
LGTM!

Hayden-IO:
Can I ask why this example is needed? CRLs are a pain to maintain and distribute, and part of the motivation behind short-lived certs is so that CRLs are not needed. This example also isn't using a CRL format, so it's an exercise left to the implementer on how to get the revoked serial numbers.

If this is more to demonstrate how to use a custom validator, could we change the example to that?

kommendorkapten:
I think the code looks good, but as @haydentherapper mention, I think we can update the wording. Clarify that this is an example of how to hook in custom certificate verification, for people using their own PKI, where long lived certificates could be used, or intermediates are shipped in the bundle instead via the TUF trust root.

codysoyland:
> Can I ask why this example is needed?

This is in response to @kommendorkapten's proposal in https://github.com/sigstore/sigstore-go/issues/298.

> If this is more to demonstrate how to use a custom validator, could we change the example to that?

Exactly, this is not a full CRL implementation, just an example of how to utilize the hooks in `TrustedMaterial` to accomplish custom certificate validation.

I can update the wording to be more explicit about this.

Hayden-IO (approved):
Thanks for the wording change

Issue #360, opened 2024-12-18
Verification may fail for in-toto statements with multiple subjects using different hash algorithms


kommendorkapten:
**Description**

When verifying that an in-toto statement matches an artifact, we first start by choosing the "strongest" hash algorithm: 
https://github.com/sigstore/sigstore-go/blob/3cda2eabf4742567f9d43b4fd24799c562a89364/pkg/verify/signature.go#L160

Then we compute the digest using that algorithm and then compare the digest against what's in the in-toto statement. This would fail if a statement have multiple entries, referencing different artifacts using different digests. So as an example:

statement have two subjects (ordered list in the statement):
1. foo: sha512
2. bar: sha256

if we verify against the artifact bar, it would fail as we pick sha512, but that's not what's specified in the satement.

**Version**

codysoyland:
I remember writing this! I landed on this solution as the simplest way that should cover the vast majority of multi-subject attestations, however it does make the assumption that each subject in the attestation uses the same digest algorithm(s).

If we would like to support multi-subject attestations that use different digest algorithms per subject, we would need to hash the input `io.Reader` multiple times, once per algorithm. The simplest way to do that would be to copy the bytes from the `io.Reader` to a buffer and use that as the input to multiple hashers. That can be done in small blocks to avoid using too much memory for large artifacts. In any case, this would increase the complexity of this section of code.

I felt at the time that it was unlikely that somebody would produce a multiple-subject attestation with different algorithms per subject, but I suppose that could happen. Is this something that is important to support?

codysoyland:
I got a bit nerd-sniped and wrote a tool that can [compute multiple hashes at once](https://github.com/sigstore/sigstore-go/pull/361).

I'm still not sure if it's worth the added complexity, but this proves it can be done in a memory-efficient way: `io.Copy` defaults to a 32KB block size, so the multihasher does not need to buffer the whole file.

kommendorkapten:
I can see this may become an issue in the future, but still as an edge case. I would vote to implement the multi hasher, it's fast should doing one or two more hashes shouldn't be an issue. We can even start to iterate over the digests to build out a set of hashes listed, then have an allow list to filter against, then perform the matching. This means that in the most cases, we will still only be hashing with one algorithm.

Hayden-IO:
Fixed in #361.

PR #361, opened 2024-12-18
Support for multi-subject attestations using different hash algorithms

codysoyland:
#### Summary

Support for attestations with multiple subjects using different hash algorithms.

https://github.com/sigstore/sigstore-go/issues/360

### DIFF
--- a/pkg/verify/signature.go
+++ b/pkg/verify/signature.go
@@ -155,11 +155,14 @@ func verifyEnvelopeWithArtifact(verifier signature.Verifier, envelope EnvelopeCo
 	// determine which hash functions to use
 	hashFuncs, err := getHashFunctions(statement)
 	if err != nil {
-		return fmt.Errorf("could not verify artifact: unable to determine hash functions: %w", err)
+		return fmt.Errorf("unable to determine hash functions: %w", err)
 	}
 
 	// Compute digest of the artifact.
-	hasher := newMultihasher(hashFuncs)
+	hasher, err := newMultihasher(hashFuncs)
+	if err != nil {
+		return fmt.Errorf("could not verify artifact: unable to create hasher: %w", err)
+	}
 	_, err = io.Copy(hasher, artifact)
 	if err != nil {
 		return fmt.Errorf("could not verify artifact: unable to calculate digest: %w", err)
@@ -260,7 +263,10 @@ type multihasher struct {
 	hashes    []hash.Hash
 }
 
-func newMultihasher(hashfuncs []crypto.Hash) *multihasher {
+func newMultihasher(hashfuncs []crypto.Hash) (*multihasher, error) {
+	if len(hashfuncs) == 0 {
+		return nil, errors.New("no hash functions specified")
+	}
 	hashes := make([]hash.Hash, len(hashfuncs))
 	for i := range hashfuncs {
 		hashes[i] = hashfuncs[i].New()
@@ -268,7 +274,7 @@ func newMultihasher(hashfuncs []crypto.Hash) *multihasher {
 	return &multihasher{
 		hashfuncs: hashfuncs,
 		hashes:    hashes,
-	}
+	}, nil
 }
 
 func (m *multihasher) Write(p []byte) (n int, err error) {
--- a/pkg/verify/signature.go
+++ b/pkg/verify/signature.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"hash"
 	"io"
+	"slices"
 
 	in_toto "github.com/in-toto/attestation/go/v1"
 	"github.com/secure-systems-lab/go-securesystemslib/dsse"
@@ -146,56 +147,40 @@ func verifyEnvelopeWithArtifact(verifier signature.Verifier, envelope EnvelopeCo
 	if err = limitSubjects(statement); err != nil {
 		return err
 	}
-
-	var artifactDigestAlgorithm string
-	var artifactDigest []byte
-
-	// Determine artifact digest algorithm by looking at the first subject's
-	// digests. This assumes that if a statement contains multiple subjects,
-	// they all use the same digest algorithm(s).
+	// Sanity check (no subjects)
 	if len(statement.Subject) == 0 {
 		return errors.New("no subjects found in statement")
 	}
-	if len(statement.Subject[0].Digest) == 0 {
-		return errors.New("no digests found in statement")
-	}
 
-	// Select the strongest digest algorithm available.
-	for _, alg := range []string{"sha512", "sha384", "sha256"} {
-		if _, ok := statement.Subject[0].Digest[alg]; ok {
-			artifactDigestAlgorithm = alg
-			continue
-		}
-	}
-	if artifactDigestAlgorithm == "" {
-		return errors.New("could not verify artifact: unsupported digest algorithm")
+	// determine which hash functions to use
+	hashFuncs, err := determineHashFunctions(statement)
+	if err != nil {
+		return fmt.Errorf("could not verify artifact: unable to determine hash functions: %w", err)
 	}
 
 	// Compute digest of the artifact.
-	var hasher hash.Hash
-	switch artifactDigestAlgorithm {
-	case "sha512":
-		hasher = crypto.SHA512.New()
-	case "sha384":
-		hasher = crypto.SHA384.New()
-	case "sha256":
-		hasher = crypto.SHA256.New()
-	}
+	hasher := newMultihasher(hashFuncs)
 	_, err = io.Copy(hasher, artifact)
 	if err != nil {
 		return fmt.Errorf("could not verify artifact: unable to calculate digest: %w", err)
 	}
-	artifactDigest = hasher.Sum(nil)
+	artifactDigests := hasher.Sum(nil)
 
 	// Look for artifact digest in statement
 	for _, subject := range statement.Subject {
 		for alg, digest := range subject.Digest {
-			hexdigest, err := hex.DecodeString(digest)
+			hf, err := algStringToHashFunc(alg)
 			if err != nil {
-				return fmt.Errorf("could not verify artifact: unable to decode subject digest: %w", err)
+				continue
 			}
-			if alg == artifactDigestAlgorithm && bytes.Equal(artifactDigest, hexdigest) {
-				return nil
+			if artifactDigest, ok := artifactDigests[hf]; ok {
+				hexdigest, err := hex.DecodeString(digest)
+				if err != nil {
+					continue
+				}
+				if bytes.Equal(artifactDigest, hexdigest) {
+					return nil
+				}
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


codysoyland:
Currently, this PR is a draft containing only a multi-digest hashing tool.

pkg/verify/signature.go
```diff
@@ -269,3 +254,122 @@ func limitSubjects(statement *in_toto.Statement) error {
...
+	case "sha384":
+		return crypto.SHA384, nil
+	case "sha512":
+		return crypto.SHA512, nil
+	default:
+		return 0, errors.New("unsupported digest algorithm")
+	}
+}
+
+func determineHashFunctions(statement *in_toto.Statement) ([]crypto.Hash, error) {
```
codysoyland: This func needs tests.

pkg/verify/signature.go
```diff
@@ -146,56 +147,40 @@ func verifyEnvelopeWithArtifact(verifier signature.Verifier, envelope EnvelopeCo
...
-			hexdigest, err := hex.DecodeString(digest)
+			hf, err := algStringToHashFunc(alg)
 			if err != nil {
-				return fmt.Errorf("could not verify artifact: unable to decode subject digest: %w", err)
+				continue
 			}
-			if alg == artifactDigestAlgorithm && bytes.Equal(artifactDigest, hexdigest) {
-				return nil
+			if artifactDigest, ok := artifactDigests[hf]; ok {
+				hexdigest, err := hex.DecodeString(digest)
```
kommendorkapten: Minor but `hexdigest` is misleading IMHO, we are decoding, so call it the `digest` on line 171 `hexdigest` and the value above digest, as more accurately represents the data.

pkg/verify/signature.go
kommendorkapten: nit but maybe call this `getHashFunctions`?

pkg/verify/signature.go
```diff
@@ -269,3 +254,122 @@ func limitSubjects(statement *in_toto.Statement) error {
...
+	}
+
+	algorithmCounts := supportedHashFuncsCount()
+	for _, subject := range statement.Subject {
+		for alg := range subject.Digest {
+			hf, err := algStringToHashFunc(alg)
+			if err != nil {
+				continue
+			}
+			algorithmCounts[hf]++
```
kommendorkapten: This would add whatever algorithms that's listed in the subject, `algorithmCounts["foo"]++` would set the value for `foo` to one. You would first need to test with `_, ok := algorithmCounts[hf]; ok` first.

kommendorkapten: See https://go.dev/play/p/lA0HVW2a8jx for an example

pkg/verify/signature.go
kommendorkapten: I'm also thinking if we need this complexity? My take is that we can simplify this create a set (i.e using a map) of all seen algorithms that are listed in the subjects which also satisfies `supportedHashFuncs()`, then just return the content of the set. As its not expensive to compute the digests I would favour simpler code. Thoughs?

pkg/verify/signature.go
codysoyland: That's the intent... I don't think there's an issue here? algorithmCounts is set to a map with the keys already set.

pkg/verify/signature.go
codysoyland: Yeah, as said in our private chat, it would probably be good to simplify. This is an attempt to find the smallest number of digests that satisfy the most number of subject candidates, but it's probably over-engineered.

pkg/verify/signature.go
kommendorkapten: Ok, I think I misunderstood the code. If the subject lists `md5`, the function would return that? I don't see any filtering against `supportedHashFuncsCount()`?

codysoyland: The filtering is done implicitly in altStringToHashFunc, so md5 would not be returned. The code is unclear so I'll rewrite and simplify.

pkg/verify/signature.go
```diff
@@ -269,3 +254,110 @@ func limitSubjects(statement *in_toto.Statement) error {
...
+	}
+
+	supportedHashFuncs := []crypto.Hash{crypto.SHA512, crypto.SHA384, crypto.SHA256}
+	chosenHashFuncs := make([]crypto.Hash, 0, len(supportedHashFuncs))
+	subjectHashFuncs := make([][]crypto.Hash, len(statement.Subject))
+
+	// go through the statement and make a simple data structure to hold the
+	// list of hash funcs for each subject (subjectHashFuncs)
+	for i, subject := range statement.Subject {
+		for alg := range subject.Digest {
```
Hayden-IO, line 325: Should we err out if digest is empty? This is the current behavior.

pkg/verify/signature.go
```diff
@@ -269,3 +254,110 @@ func limitSubjects(statement *in_toto.Statement) error {
...
+}
+
+// getHashFunctions returns the smallest subset of supported hash functions
+// that are needed to verify all subjects in a statement.
+func getHashFunctions(statement *in_toto.Statement) ([]crypto.Hash, error) {
+	if len(statement.Subject) == 0 {
+		return nil, errors.New("no subjects found in statement")
+	}
+
+	supportedHashFuncs := []crypto.Hash{crypto.SHA512, crypto.SHA384, crypto.SHA256}
```
Hayden-IO, line 318: Can this be a constant?

pkg/verify/signature.go
```diff
@@ -269,3 +254,110 @@ func limitSubjects(statement *in_toto.Statement) error {
...
+
+	// for each subject, see if we have chosen a compatible hash func, and if
+	// not, add the first one that is supported
+	for _, hfs := range subjectHashFuncs {
+		// if any of the hash funcs are already in chosenHashFuncs, skip
+		if len(intersection(hfs, chosenHashFuncs)) > 0 {
+			continue
+		}
+
+		// check each supported hash func and add chose it if the subject
```
Hayden-IO:
```suggestion
		// check each supported hash func and add it if the subject
```

pkg/verify/signature.go
```diff
@@ -146,56 +147,40 @@ func verifyEnvelopeWithArtifact(verifier signature.Verifier, envelope EnvelopeCo
...
-	for _, alg := range []string{"sha512", "sha384", "sha256"} {
-		if _, ok := statement.Subject[0].Digest[alg]; ok {
-			artifactDigestAlgorithm = alg
-			continue
-		}
-	}
-	if artifactDigestAlgorithm == "" {
-		return errors.New("could not verify artifact: unsupported digest algorithm")
+	// determine which hash functions to use
+	hashFuncs, err := getHashFunctions(statement)
```
Hayden-IO, line 156: What do you think about maintaining a multi hasher per subject? With the current approach, we have to hash every subject with all algorithms, which is potentially inefficient. For example, if we have subject A with digests sha512 and sha384, and subject B with digest sha256, then we only need to compute 3 total digests, though with this approach, we'd compute 6.

pkg/verify/signature.go
```diff
@@ -146,56 +147,40 @@ func verifyEnvelopeWithArtifact(verifier signature.Verifier, envelope EnvelopeCo
...
 		return fmt.Errorf("could not verify artifact: unable to calculate digest: %w", err)
 	}
-	artifactDigest = hasher.Sum(nil)
+	artifactDigests := hasher.Sum(nil)
 
 	// Look for artifact digest in statement
 	for _, subject := range statement.Subject {
-		for alg, digest := range subject.Digest {
-			hexdigest, err := hex.DecodeString(digest)
+		for alg, hexdigest := range subject.Digest {
```
Hayden-IO, line 174: Another idea for an optimization is to create the mapping between subject and the algorithms and expected digests in one pass. With this proposed solution, we make two passes, one to build the digest algorithms and one to verify the digests, which would be a performance regression from the previous solution.

Hayden-IO (dismissed):
Had some suggestions for performance improvements to minimize iterations over the statement and compute fewer digests.

pkg/verify/signature.go
codysoyland: This code still only supports calculating the digests for a single artifact. In your example, it would compute two digests (sha512 and sha256) for the same artifact and compare against both subjects in the attestation.

If we add support for #363, we would need to calculate four total digests (both sha512 and sha256) for each input file, and compare against the subject digests.

pkg/verify/signature.go
phillmv: Interpreting [the spec](https://github.com/in-toto/attestation/blob/main/spec/v1/statement.md#fields) to say the subject is invalid if the digest is the empty string seems pretty defensible to me / 👍 to erroring out on empty digests?

pkg/verify/signature.go
Hayden-IO: You're right, I was thinking we specify the subject name we're verifying, rather than searching for a given digest. The optimization I specified would require knowing the subject name, which isn't something checked currently.

pkg/verify/signature.go
codysoyland: I'm a little iffy on this, because verification may still pass if there exists a subject with the given digest, even if a subject exists that does not contain a digest. I'm not sure it's the responsibility of the verifier to make sure every subject contains a digest.

pkg/verify/signature.go
codysoyland: Slices are mutable and therefore cannot be declared as constants.

pkg/verify/signature.go
codysoyland: Given that would require shifting a lot of logic into the already tested `getHashFunctions` (if I understand correctly), I'm not too concerned about the performance penalty of iterating through the digests here again. I do think it's worth revisiting when/if we implement #363.

pkg/verify/signature.go
kommendorkapten:
```suggestion
	supportedHashFuncs := map[crypto.Hash]struct{}{
	        crypto.SHA512: {}, 
	        crypto.SHA384:  {},
	        crypto.SHA256: {},
	}
```
And of course for the other slices too.
this way we can just use _, found := supportedHashFunc[someHashFunc] and remove some custom code that manages all the checking. WDYT?

pkg/verify/signature.go
```diff
@@ -269,3 +254,110 @@ func limitSubjects(statement *in_toto.Statement) error {
 	}
 	return nil
 }
+
+type multihasher struct {
+	hashfuncs []crypto.Hash
+	hashes    []hash.Hash
+}
+
+func newMultihasher(hashfuncs []crypto.Hash) *multihasher {
```
kommendorkapten:
```suggestion
func newMultihasher(hashfuncs []crypto.Hash) (*multihasher, error) {
```

pkg/verify/signature.go
```diff
@@ -269,3 +254,110 @@ func limitSubjects(statement *in_toto.Statement) error {
...
+	hashes := make([]hash.Hash, len(hashfuncs))
```
kommendorkapten, line 270:
```suggestion
        if len(hashfuncs) == 0 {
                return nil, errors.New("no hash functions specified")
        }
	hashes := make([]hash.Hash, len(hashfuncs))
```

pkg/verify/signature.go
kommendorkapten: I rather be extra cautious here so we don't can fall into a trap where nothing is selected and that somehow can trigger a valid comparison. There is a check in `getHashFunctions`, but I rather have extra safety nets for this code, as subject matching is very important to get right. So I would prefer that each function verifies the input first.

pkg/verify/signature.go
codysoyland: I was just looking and can't really find what custom code we could remove if we use a map here. I'm using the generic [`slices.Contains`](https://pkg.go.dev/slices#Contains) in a few places so I think that's equivalent to your suggestion?

pkg/verify/signature.go
codysoyland: Added the return value and also wrote a table-based test for it.

kommendorkapten (approved):
💯

PR #364, opened 2024-12-19
Refactor DoS limits to separate func

codysoyland:

#### Summary

Just a small cleanup of the DoS protection logic.

### DIFF
--- a/pkg/verify/signature.go
+++ b/pkg/verify/signature.go
@@ -24,6 +24,7 @@ import (
 	"hash"
 	"io"
 
+	in_toto "github.com/in-toto/attestation/go/v1"
 	"github.com/secure-systems-lab/go-securesystemslib/dsse"
 	"github.com/sigstore/sigstore-go/pkg/root"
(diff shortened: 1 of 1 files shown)


pkg/verify/signature.go
```diff
@@ -142,6 +143,11 @@ func verifyEnvelopeWithArtifact(verifier signature.Verifier, envelope EnvelopeCo
 	if err != nil {
 		return fmt.Errorf("could not verify artifact: unable to extract statement from envelope: %w", err)
 	}
+	err = limitSubjects(statement)
```
Hayden-IO: style nit, can make this one clause with:

```
if err := limitSubjects(statement); err != nil {
  return err
}
```

Hayden-IO (dismissed):

pkg/verify/signature.go
codysoyland: I'm surprised we don't have a linter complaining about that.

Hayden-IO (approved):

kommendorkapten (approved):

PR #366, opened 2024-12-20
Fix intoto unmarshal

### LINKED ISSUE #365: JSON marshal of VerificationResult yields incorrectly-formatted in-toto attestation
**Description**

Due to in-toto moving to protobuf, there is an issue where marshalling `VerificationResult` (with `json.Marshal`) will produce a JSON object with an incorrectly formatted in-toto statement.

Specifically, the `predicateType` field is named `predicate_type`, and other fields are made into snake_case instead of the expected `camelCase` as [required by spec.](https://github.com/in-toto/attestation/blob/main/spec/v1/predicate.md)

codysoyland:
#### Summary

This PR is a modified copy of @gillisandrew's https://github.com/sigstore/sigstore-go/pull/326 with some slight simplifications. It is API-compatible with the existing codebase.

The [attempt to upstream](https://github.com/in-toto/attestation/pull/403) the fix for this is taking longer than expected, so this fix will serve as a temporary solution.

Fixes #365

### DIFF
--- a/pkg/verify/signed_entity.go
+++ b/pkg/verify/signed_entity.go
@@ -225,7 +225,6 @@ func NewVerificationResult() *VerificationResult {
 
 // MarshalJSON deals with protojson needed for the Statement.
 // Can be removed when https://github.com/in-toto/attestation/pull/403 is merged.
-// TODO: Should we also add UnmarshalJSON?
 func (b *VerificationResult) MarshalJSON() ([]byte, error) {
 	statement, err := protojson.Marshal(b.Statement)
 	if err != nil {
@@ -241,6 +240,21 @@ func (b *VerificationResult) MarshalJSON() ([]byte, error) {
 		Alias:     Alias(*b),
 		Statement: statement,
 	})
+}
+
+func (b *VerificationResult) UnmarshalJSON(data []byte) error {
+	b.Statement = &in_toto.Statement{}
+	type Alias VerificationResult
+	aux := &struct {
+		Alias
+		Statement json.RawMessage `json:"statement,omitempty"`
+	}{
+		Alias: Alias(*b),
+	}
+	if err := json.Unmarshal(data, aux); err != nil {
+		return err
+	}
+	return protojson.Unmarshal(aux.Statement, b.Statement)
 }
 
 type PolicyOption func(*PolicyConfig) error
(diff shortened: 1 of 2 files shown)


pkg/verify/signed_entity_test.go
```diff
@@ -422,3 +424,47 @@ func TestSigstoreBundle2Sig(t *testing.T) {
...
+
+	result := verify.NewVerificationResult()
+	result.Statement = &statement
+
+	// marshal the statement to JSON
+	statementJSON, err := json.Marshal(result)
+	assert.NoError(t, err)
+	want := `
+	{
+		"mediaType": "application/vnd.dev.sigstore.verificationresult+json;version=0.1",
```
phillmv, line 452: This is technically not `verificationResult`'s fault at all but since we _are_ modifying the schema as a result of this change, I feel like it'd be reasonable to increment the version 🤔.

pkg/verify/signed_entity.go
```diff
@@ -221,6 +223,26 @@ func NewVerificationResult() *VerificationResult {
 	}
 }
 
+// MarshalJSON deals with protojson needed for the Statement.
+// Can be removed when https://github.com/in-toto/attestation/pull/403 is merged.
+// TODO: Should we also add UnmarshalJSON?
```
phillmv: If someone were to try to parse the json ouput using Go,

- would they use this library's types to do so?
- would the unmarshal step work with the output sans errors?

if it would error out otherwise, seems wise to provide a working output/input loop by default (i.e. yes to adding an UnmarshalJSON)?

pkg/verify/signed_entity_test.go
codysoyland: I don't really care either way, but this is a recently-introduced bug, and older versions of sigstore-go produced the correct schema and called it `0.1`. But we can call it whatever because I am unaware of anything that actually parses this mediatype.

pkg/verify/signed_entity.go
codysoyland: I'm just being lazy. It's easy enough to add, so I'll do it.

pkg/verify/signed_entity_test.go
Hayden-IO: Given it's a bug fix, keeping it the same seems reasonable.

Hayden-IO (approved):
Creative solution with the alias!

PR #379, opened 2025-01-14
Support additional SigningConfig configurations

Hayden-IO:
This adds option functions to replace and append signing configuration URLs. The signing example has also been updated to pull the SigningConfig from staging, as it is already present in the repo.


kommendorkapten (approved):
Nice.

PR #384, opened 2025-01-17
chore: relax go directive to permit 1.22.x

dnwe:
#### Summary

Now that sigstore/sigstore#1878 and sigstore/rekor#2323 have both landed in tagged release we can pull those updates into sigstore-go and similarly relax the go.mod go directive here to 1.22.0 as well.

The go directive sets the minimum version of Go required to use this module. There's no need to set this to the latest semver patch version of a given Go release unless the semantics of that version of Go are required to build/test/use the module.

Ref: https://go.dev/ref/mod#go-mod-file-go

#### Release Note

NONE

#### Documentation

NONE


examples/oci-image-verification/go.mod
```diff
@@ -61,8 +63,8 @@ require (
 	github.com/sassoftware/relic v7.2.1+incompatible // indirect
 	github.com/secure-systems-lab/go-securesystemslib v0.9.0 // indirect
 	github.com/shibumi/go-pathspec v1.3.0 // indirect
-	github.com/sigstore/rekor v1.3.7 // indirect
-	github.com/sigstore/timestamp-authority v1.2.3 // indirect
+	github.com/sigstore/rekor v1.3.8 // indirect
+	github.com/sigstore/timestamp-authority v1.2.2 // indirect
```
codysoyland: Was `timestamp-authority` downgraded by accident??

dnwe: No, it’s an indirect dep and has been downgraded elsewhere, presumably because the current version has a higher go directive. It’ll need similar changes to this repo to bring it in-line

dnwe: No, it’s an indirect dep and has been downgraded elsewhere, presumably because the current version has a higher go directive. It’ll need similar changes to this repo to bring it in-line

dnwe: Raised https://github.com/sigstore/timestamp-authority/pull/927 for that one

dnwe: Looking at the version tag comparison https://github.com/sigstore/timestamp-authority/compare/v1.2.2...v1.2.3 it seems to be 95% dependabot dependency bumps and only https://github.com/sigstore/timestamp-authority/pull/644 is a functional change, which is in CLI tooling that doesn't effect the import usage here, so I think it is fine for it to be downgraded in the meantime anyway

Hayden-IO: We can cut a new release of timestamp-authority with the downgrade.

Hayden-IO: Done - https://github.com/sigstore/timestamp-authority/releases/tag/v1.2.4

dnwe: Now that v1.2.4 is available with a similar go1.22.0 we can bump it back up again. Added a commit to bump timestamp-authority

Hayden-IO: Should this go mod file be updated as well?

dnwe: Yep fixed with new commit to do that and go mod tidy

Also bumped toolchain to latest go1.23.5 as that has been released in the interim too

Hayden-IO (dismissed):

dnwe:
@haydentherapper please can you re-review, I had to add another commit to rollback #388 which was just merged to master and required 1.23

codysoyland (approved):

codysoyland:
Thanks @dnwe!

PR #404, opened 2025-02-07
Simplify test data getters

codysoyland:
#### Summary

I find the existing test data functions a bit confusing and require a lot of boilerplate, so I decided it would be cleaner to modify the helpers to be more general and fetch by filename.

I also renamed the bundles to be more consistent and use the `.sigstore.json` extension.


steiza (approved):
LGTM!

kommendorkapten (approved):
This is a good improvement!

PR #408, opened 2025-02-12
Improve URLToPath

### LINKED ISSUE #407: `URLToPath` is not Windows-safe
**Description**

`tuf.URLToPath` trims the `https://` prefix of a URL and replaces other `/`s with `-` to be POSIX-safe when the output is used as a directory name. This does not work on Windows if the URL starts with `http://` or uses a nonstandard port, both of which can occur in tests.

See this test run on Windows as an example: https://github.com/cmurphy/sigstore-go/actions/runs/13270875530/job/37049712149

This affects the cosign unit tests as we try to implement https://github.com/sigstore/cosign/issues/3548.

**Version**

5008d5c88f8bfe9d3f9f4bf55150cf118849478b

codysoyland:
#### Summary

Improvements to URLToPath func for wider compatibility.

- handle `http://` scheme
- cut out colons
- make filename lowercase
- added tests

Fixes https://github.com/sigstore/sigstore-go/issues/407

### DIFF
--- a/pkg/tuf/client.go
+++ b/pkg/tuf/client.go
@@ -199,9 +199,10 @@ func URLToPath(url string) string {
 	// Strip scheme, replace slashes with dashes
 	// e.g. https://github.github.com/prod-tuf-root -> github.github.com-prod-tuf-root
 	fn := url
-	if len(fn) > 8 && fn[:8] == "https://" {
-		fn = fn[8:]
-	}
+	fn, _ = strings.CutPrefix(fn, "https://")
+	fn, _ = strings.CutPrefix(fn, "http://")
 	fn = strings.ReplaceAll(fn, "/", "-")
-	return fn
+	fn = strings.ReplaceAll(fn, ":", "-")
+
+	return strings.ToLower(fn)
 }
--- a/pkg/tuf/client_test.go
+++ b/pkg/tuf/client_test.go
@@ -17,16 +17,15 @@ package tuf
 import (
 	"crypto"
 	"crypto/sha256"
+	"fmt"
 	"net/url"
+	"os"
+	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
 	"testing"
 	"time"
-
-	"fmt"
-	"os"
-	"path/filepath"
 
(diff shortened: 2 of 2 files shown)


pkg/tuf/client.go
```diff
@@ -199,9 +199,10 @@ func URLToPath(url string) string {
 	// Strip scheme, replace slashes with dashes
 	// e.g. https://github.github.com/prod-tuf-root -> github.github.com-prod-tuf-root
 	fn := url
-	if len(fn) > 8 && fn[:8] == "https://" {
-		fn = fn[8:]
-	}
+	fn, _ = strings.CutPrefix(fn, "https://")
+	fn, _ = strings.CutPrefix(fn, "http://")
```
Hayden-IO, line 203: Does the TUF client require a TLS connection? Is that why stripping off the scheme is OK?

I ask because accessing TUF metadata over HTTP shouldn't be an issue as the metadata is signed (same reason why CA certs/CRLs are served over HTTP sometimes).

The root of trust must be the bundled or provided TUF root, not WebPKI.

codysoyland: I think `http://` is used in unit tests (see https://github.com/sigstore/sigstore-go/issues/407).

Hayden-IO: I'm wondering if this needs more changes, that we can't just drop the scheme and need to plumb through whether or not the connection requires TLS to the TUF client.

codysoyland: This is just used for the cache directory name, but I can see how we might want to put the scheme in the directory name. The full URL is passed to go-tuf so I think non-TLS is supported there.

Hayden-IO: It’s a niche case to support, maybe this is fine and we just document it? I wouldn’t really expect access to a TUF repo over both http and https and needing multiple distinct caches.

Hayden-IO (approved):

pkg/tuf/client.go
kommendorkapten: TUF itself does not require TLS, the integrity control is managed via the TUF metadata. Only reason for TLS would be to add confidentiality.

PR #420, opened 2025-02-25
Update golangci-lint

codysoyland:

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -14,7 +14,6 @@
 
 run:
   timeout: 10m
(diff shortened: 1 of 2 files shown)


.golangci.yaml
```diff
@@ -14,7 +14,6 @@
 
 run:
   timeout: 10m
-  go: '1.22'
```
codysoyland, line 17: By default, this uses the Go version from the go.mod file.

Hayden-IO (approved):

PR #424, opened 2025-03-04
Use default Verifier for the public key contained in a certificate (closes #74)

### LINKED ISSUE #74: Allow configurable signing algorithms

ret2libc:
#### Summary
Closes #74 

#### Release Note
* use default verifier based on public key type, according to the algorithm registry in sigstore/sigstore (https://github.com/sigstore/sigstore/blob/main/pkg/signature/algorithm_registry.go)

#### Documentation
* ed25519 public keys will use by default the ed25519ph algorithm


go.mod
```diff
@@ -108,3 +107,5 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	k8s.io/klog/v2 v2.130.1 // indirect
 )
+
+replace github.com/sigstore/sigstore => github.com/trail-of-forks/sigstore v0.0.0-20250305100229-6bb0c7c5e889
```
kommendorkapten: This will be removed once that branch is merged into `sigstore/sigstore` right?

pkg/verify/signature.go
```diff
@@ -97,8 +97,7 @@ func VerifySignatureWithArtifactDigest(sigContent SignatureContent, verification
 
 func getSignatureVerifier(verificationContent VerificationContent, tm root.TrustedMaterial) (signature.Verifier, error) {
 	if leafCert := verificationContent.Certificate(); leafCert != nil {
-		// TODO: Inspect certificate's SignatureAlgorithm to determine hash function
-		return signature.LoadVerifier(leafCert.PublicKey, crypto.SHA256)
+		return signature.LoadVerifierFromPublicKey(leafCert.PublicKey, options.WithED25519ph())
```
kommendorkapten: the option to default to use Ed25519 with pre-hash, how would that affect scenarios where say a NIST Curve is used for ECC (e.g. P-256).

ret2libc: when the key is ECDSA the ed25519ph option is ignored, because of ECDSASignerVerifier is used. That option is only useful when the key is ED25519. I hope that answer your question.

go.mod
ret2libc: Yes! I have split those in separate commits marked with `REMOVE ME` :)

pkg/verify/signature.go
kommendorkapten: Ok. For RSA the current choice of PKCS#1v1.5 would be used still? So all works as before, except the digest would be chosen based on the key type?

This is good IMHO, but I think a comment explaining this would be helpful, as it may appear confusing otherwise if there is always a PH option?

ret2libc: > Ok. For RSA the current choice of PKCS#1v1.5 would be used still? So all works as before, except the digest would be chosen based on the key type?

Correct!

> This is good IMHO, but I think a comment explaining this would be helpful, as it may appear confusing otherwise if there is always a PH option?

Sure, I'll add a comment.

kommendorkapten:
Thanks @ret2libc, could you run `go mod tidy` then push again?

go.mod
```diff
@@ -16,7 +16,7 @@ require (
 	github.com/secure-systems-lab/go-securesystemslib v0.9.0
 	github.com/sigstore/protobuf-specs v0.4.0
 	github.com/sigstore/rekor v1.3.9
-	github.com/sigstore/sigstore v1.8.15
+	github.com/sigstore/sigstore v1.9.1-0.20250310170251-f577ba06d54c
```
Hayden-IO: can bump to 1.9.1

pkg/verify/signature.go
```diff
@@ -97,8 +97,11 @@ func VerifySignatureWithArtifactDigest(sigContent SignatureContent, verification
 
 func getSignatureVerifier(verificationContent VerificationContent, tm root.TrustedMaterial) (signature.Verifier, error) {
 	if leafCert := verificationContent.Certificate(); leafCert != nil {
-		// TODO: Inspect certificate's SignatureAlgorithm to determine hash function
-		return signature.LoadVerifier(leafCert.PublicKey, crypto.SHA256)
+		// LoadDefaultSigner/Verifier functions accept a few options to select
+		// the default signer/verifier when there are ambiguities, like for
+		// ED25519 keys, which could be used with PureEd25519 or Ed25519ph.
+		// Pass `WithED25519ph()` to select Ed25519ph by default, when ED25519 key is found.
```
Hayden-IO: Can we mention the reason for defaulting to Ed25519ph, that for hashedrekord, this is the only option?

ret2libc:
> Thanks @ret2libc, could you run `go mod tidy` then push again?

@kommendorkapten fixed!

kommendorkapten (dismissed):
Thanks!

ret2libc:
@kommendorkapten can you approve the the workflows?

Hayden-IO (dismissed):

kommendorkapten:
Can you also update the title to refer to the issue: https://github.com/sigstore/sigstore-go/issues/74 (I believe that's the correct one).

kommendorkapten:
The conformance test is failing, and the version was updated in `main` a few days ago, would you mind rebasing your changes on `main` so we can rule out that as the cause for the error?

ret2libc:
> The conformance test is failing, and the version was updated in `main` a few days ago, would you mind rebasing your changes on `main` so we can rule out that as the cause for the error?

I think it was already updated, but I re-did anyway and it's still failing :(

What's weird is that this change shouldn't really affect most common scenarios as nothing changes for ecdsa-p256. Moreover, I tested sigstore-conformance locally and I don't see any error here.

ret2libc:
I think I understand now. All errors are in `test_verify_cpython_release_bundles` and that test is disabled locally. I'll dig into that.

kommendorkapten:
Hm, I ran this (before the rebase)

```
kommendorkapten@m1m14-msft:~/git/sigstore-go % git merge-base --fork-point main
4c441c3cff4cbfb3e96d9a7703670d8ddec1628c
```

And that commit was before the merge in main, but anyway. It's still failing :( 

The test is successful for me locally too, but there are some extra work that has to be done locally to run the tests that fails (verify cpython) that requires a clone of https://github.com/woodruffw/cpython-release-tracker IIRC. I can spend some time after lunch to see if I can get conformance test running properly locally. You can some references here https://github.com/sigstore/sigstore-conformance/blob/main/action.yml#L37

ret2libc:
> Hm, I ran this (before the rebase)
> 
> ```
> kommendorkapten@m1m14-msft:~/git/sigstore-go % git merge-base --fork-point main
> 4c441c3cff4cbfb3e96d9a7703670d8ddec1628c
> ```
> 
> And that commit was before the merge in main, but anyway. It's still failing :(
> 
> The test is successful for me locally too, but there are some extra work that has to be done locally to run the tests that fails (verify cpython) that requires a clone of https://github.com/woodruffw/cpython-release-tracker IIRC. I can spend some time after lunch to see if I can get conformance test running properly locally. You can some references here https://github.com/sigstore/sigstore-conformance/blob/main/action.yml#L37

Yes indeed! I just found out about it and I was able to reproduce locally. There are some non-p256 keys there :) Thanks

ret2libc:
The problem now is that this is essentially a breaking change: by using a given hash for a given key type (instead of using sha256 everywhere), we cannot verify anymore blobs signed by older cosign with a key like ecdsa-384.

ECDSA-P384 now expects a SHA384, but until now we just used SHA256. Thus the errors.

AFAIS there are a bunch of tests in sigstore-conformance that uses ECDSA-P384 (no other non-standard keys). This was discussed previously and I was under the impression we were fine with the breakage. It would require re-signing all blobs with the new cosign (https://github.com/sigstore/cosign/pull/4050).

Just to re-iterate, the options are:
1) fix the sigstore-conformance tests so that ECDSA-P384 uses SHA384. We communicate this properly to users as a breaking change. This happens only for the BYOK case, because by default cosign uses ECDSA-P256 and we are not aware of other sigstore clients that allow BYOK.
2) have a compatibility layer in sigstore-go. We return a SignerVerifier that first performs verification with the new default algorithm, if that fails it fallbacks to use the older algorithm. This can also be controlled with a policy. At some point later we disable compatibility.

cc @haydentherapper @woodruffw

kommendorkapten:
Thanks for the summary @ret2libc 

My vote would be to fix conformance test so that P-384 uses SHA-384 to avoid confusion, and look at P-384/SHA-256 as a legacy construct.

If needed, we could bump the conformance version to indicate that his is a potentially breaking change, to let other clients update in their pace.

ret2libc:
Apparently the assumption that ECDSA-384 was not too much used might be wrong. Parts of the conformance tests use python executables and sigstore bundles (e.g. https://www.python.org/ftp/python/3.10.1/) and those .sigstore files have a ECDSA-P384 public key inside, but they use SHA256.

Not supporting ECDSA-P384 + SHA256 means people won't be able to verify those python binaries anymore or they need to be re-generated (I'm still not sure how those are generated). This affects all python releases after ~3.10.

For reference:
- https://peps.python.org/pep-0761/
- https://www.python.org/downloads/metadata/sigstore/

Hayden-IO:
@ret2libc thanks for digging into this. Can we go with the previous proposal to support p384/521 paired with sha256 as a deprecated key type?

If we can support this on the verification path, but use the correct pairing on the sign path, this won’t be breaking. The tricky part is how to do verification - we can try multiple key types, albeit that’s a bit sketchy. 

The bundle can contain a key hint, could we also add which algorithm is used? Then we can update all cpython bundles. 

Worst case, we can just support this old mapping and deprecate it with major releases across all clients later on.

Hayden-IO:
Another thought - maybe we can treat keys vs certs differently. I doubt there are many self provided keys that also requested certificates.

woodruffw:
> If we can support this on the verification path, but use the correct pairing on the sign path, this won’t be breaking. The tricky part is how to do verification - we can try multiple key types, albeit that’s a bit sketchy.

I was thinking about this a bit, and I think we can do this in a compatible and non-sketchy/brittle way:

1. Perform the "normal" pairing verification flow: P256+SHA256, P384+SHA384, etc.
2. If the key/cert conveys a P384 key but the SHA384 image doesn't verify, then _additionally_ compute the SHA256 image and attempt verification with that.

I _think_ this would retain compatibility at the cost of a small amount of additional client complexity during verification flows (namely, clients would need to additionally compute a new digest over the input in the worst case). The security of this fallback depends on the assumption that SHA256 and SHA384 don't have overlapping image domains, which should be true both as a property of their cryptographic strength and their difference in digest size.

CC @ret2libc @steiza for thoughts 🙁

kommendorkapten:
> I was thinking about this a bit, and I think we can do this in a compatible and non-sketchy/brittle way:

That seems reasonable I think. However, as we are standardizing on a set of supported algorithms in the bundle spec, can we at least make sure that this fallback scheme is only valid for bundles up v5? That would make me feel a bit better.

woodruffw:
> That seems reasonable I think. However, as we are standardizing on a set of supported algorithms in the bundle spec, can we at least make sure that this fallback scheme is only valid for bundles up v5? That would make me feel a bit better.

Same 🙂 -- I suppose that would prevent infallible automatic upgrades of older bundles to newer versions, but IMO that's not a huge issue.

ret2libc:
> > If we can support this on the verification path, but use the correct pairing on the sign path, this won’t be breaking. The tricky part is how to do verification - we can try multiple key types, albeit that’s a bit sketchy.
> 
> I was thinking about this a bit, and I think we can do this in a compatible and non-sketchy/brittle way:
> 
> 1. Perform the "normal" pairing verification flow: P256+SHA256, P384+SHA384, etc.
> 2. If the key/cert conveys a P384 key but the SHA384 image doesn't verify, then _additionally_ compute the SHA256 image and attempt verification with that.

This should be doable.

> That seems reasonable I think. However, as we are standardizing on a set of supported algorithms in the bundle spec, can we at least make sure that this fallback scheme is only valid for bundles up v5? That would make me feel a bit better.

I'll investigate what's needed for that!

Hayden-IO:
LGTM to all of the above.

Can we review if there's any service changes we need to make or revert with this? I just checked [Rekor](https://github.com/sigstore/rekor/blob/26bfbf550e1c3b086851d0ccbf2d8ebf024c7f46/pkg/pki/x509/x509.go#L93), and it looks like since we aren't loading in any default hashes, we'll continue to use sha256, so clients can continue to upload key-signed artifacts to Rekor. (I was thinking about this in the context of the Cosign change - currently you can upload an artifact to Rekor whose digest is sha512, but I think the artifact must be signed using sha256)

ret2libc:
Ok this seems to work. I've implemented fallbacks for P384/P521, but right now it's always enabled. If we want to do this compatibility check only for some artifacts, we need a way to propogate the bundle version somehow (not sure there is one already).

Moreover, shall we update the [client spec](https://docs.google.com/document/d/1kbhK2qyPPk8SLavHzYSDM8-Ueul9_oxIMVFuWMWKz0E/edit?tab=t.0#heading=h.g11ovq2s1jxh) to mention the fallback mechanisms?

ret2libc:
In order to enable/disable the fallback logic based on the bundle version, I had to expose a `Version` method in `SignedEntity`. This way the logic is not policy-controlled, but just controlled based on the entity being verified.

examples/oci-image-verification/go.mod
```diff
@@ -99,3 +99,7 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	k8s.io/klog/v2 v2.130.1 // indirect
 )
+
+replace github.com/sigstore/protobuf-specs => github.com/trail-of-forks/protobuf-specs v0.3.3-0.20250331162432-e5e775cafa56
```
Hayden-IO: Can bump to v0.4.1 now

pkg/testing/ca/ca.go
```diff
@@ -56,16 +58,17 @@ import (
...
+	tsaCA                   *root.SigstoreTimestampingAuthority
+	tsaLeafKey              *ecdsa.PrivateKey
+	rekorKey                *ecdsa.PrivateKey
+	ctlogKey                *ecdsa.PrivateKey
+	publicKeyVerifier       map[string]root.TimeConstrainedVerifier
+	signingAlgorithmDetails signature.AlgorithmDetails
 }
 
-func NewVirtualSigstore() (*VirtualSigstore, error) {
+func NewVirtualSigstoreCustom(signingKeyDetails v1.PublicKeyDetails) (*VirtualSigstore, error) {
```
Hayden-IO: nit on the name to be more descriptive, `NewVirtualSigstoreWithSigningAlg` or something?

pkg/verify/signature.go
```diff
@@ -104,10 +116,96 @@ func VerifySignatureWithArtifactDigests(sigContent SignatureContent, verificatio
...
+}
+
+func compatSignatureVerifier(leafCert *x509.Certificate, enableCompat bool) (signature.Verifier, error) {
+	// LoadDefaultSigner/Verifier functions accept a few options to select
+	// the default signer/verifier when there are ambiguities, like for
+	// ED25519 keys, which could be used with PureEd25519 or Ed25519ph.
+	//
+	// Pass `WithED25519ph()` to select Ed25519ph by default, when ED25519
+	// key is found, because for hashedrekord entries this is the only option.
+	defaultOpts := []signature.LoadOption{options.WithED25519ph()}
```
Hayden-IO: This would mean this fallback verifier is only valid for hashedrekord and not DSSE, yes? We should document if so

Hayden-IO (commented):
Overall this LGTM but I'd want another maintainer to chime in as well

pkg/verify/signature.go
ret2libc: to be honest there's no runtime check or anything in place to enforce that this is done only for hashedrekord. It will just use pre-hashed every time there is a ed25519 key no matter what. Shall we restrict the behaviour to just hashedrekord?

pkg/verify/signed_entity.go
```diff
@@ -665,20 +665,45 @@ func (v *SignedEntityVerifier) Verify(entity SignedEntity, pb PolicyBuilder) (*V
...
+	case "0.1":
+		fallthrough
+	case "v0.2":
+		fallthrough
+	case "0.2":
+		fallthrough
+	case "v0.3":
+		fallthrough
+	case "0.3":
+		enableCompat = true
```
kommendorkapten, line 686: For the future, this should be exposed as a policy option I think. But that's for the future when we have `v0.4` bundles 🙂

kommendorkapten (dismissed):
Nice job, thank you!

pkg/verify/signed_entity.go
ret2libc: I initially exposed it as an option, but then every sigstore-go user (well, maybe there's only cosign, I'm not sure) would have to implement the bundle version check+policy enable/disable by themselves. In this way, instead, v0.4 bundles will never use the fallback logic, while v0.3 will. This is required if clients still want to verify older signatures, I think.

kommendorkapten: That's fair. And to prevent downgrade attacks, the `Bundle.MinVersion` check can be exposed as a verification option `WithMinBundleVersion` to safeguard that no older bundles can be verified if a user want to ensure only non deprecated algorithms are used.

pkg/verify/signed_entity.go
```diff
@@ -665,20 +665,45 @@ func (v *SignedEntityVerifier) Verify(entity SignedEntity, pb PolicyBuilder) (*V
 		return nil, fmt.Errorf("failed to fetch signature content: %w", err)
 	}
 
+	entityVersion, err := entity.Version()
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch entity version: %w", err)
+	}
+
+	var enableCompat bool
```
steiza, line 673: TIL that Go booleans default to `false`

examples/oci-image-verification/go.mod
```diff
@@ -9,7 +9,7 @@ replace github.com/sigstore/sigstore-go => ../../
 require (
 	github.com/google/go-containerregistry v0.20.3
 	github.com/sigstore/protobuf-specs v0.4.1
-	github.com/sigstore/sigstore v1.9.1
+	github.com/sigstore/sigstore v1.9.2-0.20250408060058-404e5b5549bc
```
Hayden-IO: tagged https://github.com/sigstore/sigstore/releases/tag/v1.9.2

pkg/verify/signature.go
Hayden-IO: Can we pass the type in and set whether it’s the PH variant or not based on the type? There could also be DSSE bundles with the older mapping that we should support.

Hayden-IO (commented):
Looks great! Just one thought on supporting dsse

pkg/verify/signature.go
ret2libc: Waiting for tests, but this should be addressed with the last commit.

Hayden-IO: This looks great!

Hayden-IO (approved):

kommendorkapten (approved):
Thank you!

PR #427, opened 2025-03-08
Update `Keypair.SignData` with context param

### LINKED ISSUE #426: `Keypair.SignData` should accept a `Context` parameter
**Description**

When building a [`Keypair`](https://github.com/sigstore/sigstore-go/blob/693bbb25c3404d0089ce6de3fe7bc7089a57ce14/pkg/sign/keys.go#L32-L38) implementation backed by a cloud HSM, it would be helpful if the `SignData` function accepted a `Context` for the current request.

This would allow for better retry/timeout support when interacting with the cloud HSM API.

Something like:
```golang
SignData(ctx context.Context, data []byte) ([]byte, []byte, error)
```

instead of the existing:
```golang
SignData(data []byte) ([]byte, []byte, error)
```

bdehamer:
#### Summary
Updates the signature for the `SignData` function in the `Keypair` interface to include a `Context` parameter.

There's maybe a case to be made that ALL of the functions in the `Keypair` interface should include a `Context` parameter . . . however, in most cases there's probably a reasonable way to pre-calculate the values returned by those functions at the time that the `Keypair` is created. Open to suggestions here.

Fixes: https://github.com/sigstore/sigstore-go/issues/426

#### Release Note

Breaking change to the public `Keypair` interface

### DIFF
--- a/pkg/sign/keys_test.go
+++ b/pkg/sign/keys_test.go
@@ -15,6 +15,7 @@
 package sign
 
 import (
+	"context"
 	"testing"
 
 	protocommon "github.com/sigstore/protobuf-specs/gen/pb-go/common/v1"
@@ -26,6 +27,7 @@ func Test_EphemeralKeypair(t *testing.T) {
 		Hint: []byte("asdf"),
 	}
 
+	ctx := context.TODO()
 	ephemeralKeypair, err := NewEphemeralKeypair(opts)
 	assert.NotNil(t, ephemeralKeypair)
 	assert.Nil(t, err)
@@ -43,7 +45,7 @@ func Test_EphemeralKeypair(t *testing.T) {
 	assert.NotEqual(t, pem, "")
 	assert.Nil(t, err)
 
-	signature, digest, err := ephemeralKeypair.SignData([]byte("hello world"))
+	signature, digest, err := ephemeralKeypair.SignData(ctx, []byte("hello world"))
 	assert.NotEqual(t, signature, "")
 	assert.NotEqual(t, digest, "")
 	assert.Nil(t, err)
--- a/pkg/sign/transparency_test.go
+++ b/pkg/sign/transparency_test.go
@@ -103,6 +103,8 @@ func (m *mockRekor) CreateLogEntry(_ *entries.CreateLogEntryParams, _ ...entries
 }
 
 func Test_GetTransparencyLogEntry(t *testing.T) {
+	ctx := context.TODO()
+
 	// First create a bundle with DSSE content
 	keypair, err := NewEphemeralKeypair(nil)
 	assert.Nil(t, err)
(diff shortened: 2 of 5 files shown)


steiza (approved):
LGTM!

> There's maybe a case to be made that ALL of the functions in the Keypair interface should include a Context parameter . . . however, in most cases there's probably a reasonable way to pre-calculate the values returned by those functions at the time that the Keypair is created.

🤷 Yeah, there's always a balance here between simplicity and a maximally flexible interface. `SignData` can definitely generate a request, like if you're using a HSM, so it definitely makes sense to use `Context` there.

pkg/sign/keys_test.go
```diff
@@ -26,6 +27,7 @@ func Test_EphemeralKeypair(t *testing.T) {
 		Hint: []byte("asdf"),
 	}
 
+	ctx := context.TODO()
```
codysoyland, line 30: Nit: for tests, `context.Background()` is more appropriate.

```suggestion
	ctx := context.Background()
```

pkg/sign/transparency_test.go
```diff
@@ -103,14 +103,16 @@ func (m *mockRekor) CreateLogEntry(_ *entries.CreateLogEntryParams, _ ...entries
 }
 
 func Test_GetTransparencyLogEntry(t *testing.T) {
+	ctx := context.TODO()
```
codysoyland, line 106:
```suggestion
	ctx := context.Background()
```

codysoyland (changes requested):

pkg/sign/keys_test.go
bdehamer: @codysoyland I opted for `TODO` cause that's what I saw used in other tests:

https://github.com/sigstore/sigstore-go/blob/3280a6017c880f6a4288ee43759b828977a1a810/pkg/sign/transparency_test.go#L106

https://github.com/sigstore/sigstore-go/blob/797ca59e9032c12c7e4a39092eb91c9d7c1b3b63/pkg/sign/timestamping_test.go#L82

Would you still like me to change it?

Hayden-IO: Functionally they're the same I believe. For tests, I think TODO is fine.

Hayden-IO (approved):

pkg/sign/keys_test.go
codysoyland: Yes they are functionally the same, but TODO is [intended to be updated later](https://pkg.go.dev/context#TODO):

> Code should use context.TODO when it's unclear which Context to use or it is not yet available (because the surrounding function has not yet been extended to accept a Context parameter).

codysoyland (commented):
Removing block

codysoyland (approved):

PR #428, opened 2025-03-09
added public key check for SCTs

### LINKED ISSUE #403: Verifier does not enforce `WithSignedCertificateTimestamps` when the bundle is signed with a public key
**Description**

[`func WithSignedCertificateTimestamps(threshold int) VerifierOption`](https://github.com/sigstore/sigstore-go/blob/5008d5c88f8bfe9d3f9f4bf55150cf118849478b/pkg/verify/signed_entity.go#L169) is used to specify a required threshold of SCTs present in the leaf certificate. This threshold is not enforced when the bundle is signed with a plain public key instead of a certificate. The [check for SCTs](https://github.com/sigstore/sigstore-go/blob/5008d5c88f8bfe9d3f9f4bf55150cf118849478b/pkg/verify/signed_entity.go#L586) is conditioned on the presence of a certificate. It should be an error to verify a bundle with a public key when the verifier is configured to require SCTs.

**Version**

Horiodino:
#### Summary

return err when the bundle is signed with a public key with Scts
fixes: https://github.com/sigstore/sigstore-go/issues/403

### DIFF
--- a/pkg/verify/signed_entity.go
+++ b/pkg/verify/signed_entity.go
@@ -591,6 +591,13 @@ func (v *SignedEntityVerifier) Verify(entity SignedEntity, pb PolicyBuilder) (*V
 		}
 	}
 
+	// If SCTs are required, ensure the bundle is certificate-signed not public key-signed
+	if v.config.requireSCTs {
+		if verificationContent.PublicKey() != nil {
+			return nil, errors.New("SCTs required but bundle is signed with a public key, which cannot contain SCTs")
+		}
+	}
+
 	// From spec:
 	// > ## Signature Verification
 	// > The Verifier MUST verify the provided signature for the constructed payload against the key in the leaf of the certificate chain.
--- a/pkg/verify/signed_entity_test.go
+++ b/pkg/verify/signed_entity_test.go
@@ -15,6 +15,7 @@
 package verify_test
 
 import (
+	"crypto/x509"
 	"errors"
 	"strings"
 	"testing"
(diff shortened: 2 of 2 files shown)


pkg/verify/signed_entity.go
```diff
@@ -515,6 +515,18 @@ func (v *SignedEntityVerifier) Verify(entity SignedEntity, pb PolicyBuilder) (*V
 		return nil, fmt.Errorf("failed to build policy: %w", err)
 	}
 
+	// Early check: If SCTs are required, ensure the bundle is certificate-signed not public key-signed
+	if v.config.requireSCTs {
```
Hayden-IO: Could we do this in the construction of [`SignedEntityVerifier`](https://github.com/sigstore/sigstore-go/blob/693bbb25c3404d0089ce6de3fe7bc7089a57ce14/pkg/verify/signed_entity.go#L86)?

pkg/verify/signed_entity.go
```diff
@@ -515,6 +515,18 @@ func (v *SignedEntityVerifier) Verify(entity SignedEntity, pb PolicyBuilder) (*V
...
+		verificationContent, err := entity.VerificationContent()
+		if err != nil {
+			return nil, fmt.Errorf("failed to fetch verification content: %w", err)
+		}
+
+		if leafCert := verificationContent.Certificate(); leafCert == nil {
```
Hayden-IO: Can we check for `.PublicKey()` rather than not `Certificate()`, since we also may support certificate chains at some point?

steiza (changes requested):
We want to have the concept of a `SignedEntityVerifier` be separate from a `SignedEntity`.

Imagine you're running `sigstore-go` on a server. You have your policy for what you consider valid or not, which is what you use to construct your `SignedEntityVerifier`. But at that time, you don't yet have a bundle you're verifying. Then you (re)use that verifier many, many times to verify various user-submitted `SignedEntity`s. This isn't hypothetical - this is exactly how [npm provenance](https://docs.npmjs.com/generating-provenance-statements) works on the server side.

That said, it does seem like `requireSCTs` is ignored if your bundle refers to a public key instead of having a signing certificate. Instead of this approach, I think we could have an [if statement in Verify](https://github.com/sigstore/sigstore-go/blob/693bbb25c3404d0089ce6de3fe7bc7089a57ce14/pkg/verify/signed_entity.go#L593) that checks if you have a public key but said you require SCTs?

pkg/verify/signed_entity.go
steiza: I don't think we want this check in `SignedEntityVerifier` - see https://github.com/sigstore/sigstore-go/pull/428#pullrequestreview-2671946308

Hayden-IO: I agree that this implementation of binding SignedEntity and SignedEntityVerifier together isn't what we want to do. I was thinking we could implement this check by checking for a public key in the trust root material, specifically if there is a `TrustedPublicKeyMaterial` verifier. Though I recognize now this might not be straightforward - if there are multiple verifiers for both keys and certificates, it's only at verification time for the SignedEntity we will know if it is incompatible with the verifier policy. 

so tldr - let's move it to `Verify()` like you suggested

steiza (commented):
I think this makes sense. Should we add a test to `pkg/verify/signed_entity_test.go`?

steiza (approved):
Looks good to me - thanks!

PR #430, opened 2025-03-11
go1.24.1, use 1024-bits key size in tests

dmitris:
#### Summary
Update go to `1.24.1`, modify `TestVerificationContent` in pkg/bundle/bundle_test.go to use 1024-bit keys to avoid the errors due to the current [minimum key size])(https://pkg.go.dev/crypto/rsa#hdr-Minimum_key_size).

#### Release Note
NONE

#### Documentation
n/a

### DIFF
--- a/examples/oci-image-verification/go.mod
+++ b/examples/oci-image-verification/go.mod
@@ -1,8 +1,6 @@
 module github.com/sigstore/sigstore-go/examples/oci-image-verification
 
-go 1.23.0
-
-toolchain go1.24.0
+go 1.24.1
 
 replace github.com/sigstore/sigstore-go => ../../
 
--- a/pkg/bundle/bundle_test.go
+++ b/pkg/bundle/bundle_test.go
@@ -575,9 +575,9 @@ func TestVerificationContent(t *testing.T) {
 	leafCert := &x509.Certificate{
 		SerialNumber: big.NewInt(2),
 	}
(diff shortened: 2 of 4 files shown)


codysoyland:
Is there a reason to update the go version directive? I would like to preserve compatibility with go1.23 since it is still a supported Go version.

pkg/bundle/bundle_test.go
```diff
@@ -575,9 +575,9 @@ func TestVerificationContent(t *testing.T) {
 	leafCert := &x509.Certificate{
 		SerialNumber: big.NewInt(2),
 	}
-	caKey, err := rsa.GenerateKey(rand.Reader, 512) //nolint:gosec
+	caKey, err := rsa.GenerateKey(rand.Reader, 1024) //nolint:gosec
```
Hayden-IO, line 578: We can do 2048 - 512 or 1024 is not representative of a real-world key.

examples/oci-image-verification/go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/sigstore/sigstore-go/examples/oci-image-verification
 
-go 1.23.0
-
-toolchain go1.24.0
+go 1.24.1
```
Hayden-IO, line 3: Unless we need a feature from 1.24, let's keep it at 1.23. When you build the example, the version of the Go toolchain that you build with is what is used.

dmitris:
go1.24 upgrade not necessary, let's keep on 1.23

PR #431, opened 2025-03-11
Add support for verifying multiple artifacts

### LINKED ISSUE #363: Support verification of multiple subjects
**Description**

Using [ArtifactPolicyOptions](https://pkg.go.dev/github.com/sigstore/sigstore-go@v0.6.2/pkg/verify#ArtifactPolicyOption), the verifier supports checking in-toto statement subjects for a specific artifact digest. The artifact is provided using `verify.WithArtifact` or `verify.WithArtifactDigest`. I propose adding `verify.WithArtifacts(artifacts []io.Reader)`/`verify.WithArtifactDigests([]ArtifactDigest)` to specify multiple artifacts that must all pass verification. Multi-subject verification will fail unless the bundle content is DSSE and contains an in-toto statement with a subject referencing all provided artifacts.

malancas:
#### Summary

Closes https://github.com/sigstore/sigstore-go/issues/363

This adds support for verifying multiple artifacts against in-toto statement subjects. I added `verify.WithArtifacts(artifacts []io.Reader)` and `verify.WithArtifactDigests([]ArtifactDigest)`.

To avoid creating nested for loops when checking for the presence of artifacts in the intoto statement, I utilized maps to store subjects stored in the statement by hash algorithm.

### DIFF
--- a/pkg/verify/signature.go
+++ b/pkg/verify/signature.go
@@ -240,14 +240,13 @@ func verifyEnvelopeWithArtifacts(verifier signature.Verifier, envelope EnvelopeC
 		return fmt.Errorf("unable to determine hash functions: %w", err)
 	}
 
-	// Compute digest of the artifact.
-	hasher, err := newMultihasher(hashFuncs)
-	if err != nil {
-		return fmt.Errorf("could not verify artifact: unable to create hasher: %w", err)
-	}
-
 	hashedArtifacts := make([]map[crypto.Hash][]byte, len(artifacts))
 	for i, artifact := range artifacts {
+		// Compute digest of the artifact.
+		hasher, err := newMultihasher(hashFuncs)
+		if err != nil {
+			return fmt.Errorf("could not verify artifact: unable to create hasher: %w", err)
+		}
 		_, err = io.Copy(hasher, artifact)
 		if err != nil {
 			return fmt.Errorf("could not verify artifact: unable to calculate digest: %w", err)
@@ -280,12 +279,12 @@ func verifyEnvelopeWithArtifacts(verifier signature.Verifier, envelope EnvelopeC
 	// to the mapped subject digests
 	// if we cannot find a match, exit with an error
 	for _, ha := range hashedArtifacts {
+		matchFound := false
 		for key, value := range ha {
 			statementDigests, ok := subjectDigests[key]
 			if !ok {
-				return fmt.Errorf("provided artifact digests do not match digests in statement")
+				return fmt.Errorf("no matching artifact hash algorithm found in subject digests")
 			}
-			matchFound := false
 			for _, sd := range statementDigests {
 				// if we have found a match, exit the for loop early
 				if bytes.Equal(value, sd) {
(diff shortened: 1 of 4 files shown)


codysoyland:
This looks great so far, nice work! I'll give it a more thorough pass when out of draft.

pkg/verify/signature.go
```diff
@@ -136,8 +164,12 @@ func verifyEnvelope(verifier signature.Verifier, envelope EnvelopeContent) error
 }
 
 func verifyEnvelopeWithArtifact(verifier signature.Verifier, envelope EnvelopeContent, artifact io.Reader) error {
-	err := verifyEnvelope(verifier, envelope)
-	if err != nil {
+	artifacts := []io.Reader{artifact}
```
malancas: We could just remove `verifyEnvelopeWithArtifact ` and `verifyEnvelopeWithArtifactDigest ` altogether and call `verifyEnvelopeWithArtifacts` and `verifyEnvelopeWithArtifactDigests` directly in the `VerifySignatureWithArtifact` and `VerifySignatureWithArtifactDigest` functions in signed_entity.go.

steiza (commented):
This looks great!

It would expand the scope of the pull request a bit, but I'm wondering if we should get rid of `VerifySignatureWithArtifact` and `VerifySignatureWithArtifactDigest` completely and just use `VerifySignatureWithArtifacts` and `VerifySignatureWithArtifactDigests`.

Technically those methods are part of a public API, but from some quick searching of various projects nobody is calling them directly (they're using things like `WithArtifact` and `WithArtifactDigest` to construct the policy).

Then we could also get rid of `PolicyConfig` options like `verifyArtifact`, `artifact`, `verifyArtifactDigest`, and `artifactDigest`.

What do you think? We'd have to update more files (like `pkg/verify/fuzz_test.go`) but I think it would actually simplify changes in other places. Just to be clear, we would retain `WithArtifact` and `WithArtifactDigest` as convenience functions.

malancas:
> This looks great!
> 
> It would expand the scope of the pull request a bit, but I'm wondering if we should get rid of `VerifySignatureWithArtifact` and `VerifySignatureWithArtifactDigest` completely and just use `VerifySignatureWithArtifacts` and `VerifySignatureWithArtifactDigests`.
> 
> Technically those methods are part of a public API, but from some quick searching of various projects nobody is calling them directly (they're using things like `WithArtifact` and `WithArtifactDigest` to construct the policy).
> 
> Then we could also get rid of `PolicyConfig` options like `verifyArtifact`, `artifact`, `verifyArtifactDigest`, and `artifactDigest`.
> 
> What do you think? We'd have to update more files (like `pkg/verify/fuzz_test.go`) but I think it would actually simplify changes in other places. Just to be clear, we would retain `WithArtifact` and `WithArtifactDigest` as convenience functions.

That sounds good for me. Should be easy to update tests. I'll keep `WithArtifact` and `WithArtifactDigest` intact.

steiza (approved):
LGTM!

PR #436, opened 2025-03-14
Fix lint errors, standardize policy language

Hayden-IO:
#### Summary

Small cleanup of linting errors, and standardizing policy language around Require/Ignore to be in sync with VerifierConfig.

#### Release Note

* Renamed PolicyConfig methods and unexported PolicyConfig validation. Clients are not expected to directly use these methods.

### DIFF
--- a/pkg/fulcio/certificate/summarize.go
+++ b/pkg/fulcio/certificate/summarize.go
@@ -58,7 +58,7 @@ func SummarizeCertificate(cert *x509.Certificate) (Summary, error) {
 		san, _ = cryptoutils.UnmarshalOtherNameSAN(cert.Extensions)
 	}
 	if san == "" {
-		return Summary{}, errors.New("No Subject Alternative Name found")
+		return Summary{}, errors.New("no Subject Alternative Name found")
 	}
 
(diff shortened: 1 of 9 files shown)


codysoyland (dismissed):
I agree with the naming changes. I'm wondering if we should privatize these funcs on `PolicyConfig` since they are only used in the verify package? Otherwise, I would probably add some notes in the `Release Notes` section of the PR description to make a note about these name changes.

Hayden-IO:
I also wondered about making them private, given it looked like these values are only set based on when an artifact/attestation/identity/etc is provided, rather than requiring a client to explicitly opt in to verification.

codysoyland:
I can't find the funcs `WeExpectSigningKey` and `WeExpectAnArtifact` in use anywhere on GitHub code search, so I would either make them private or just delete them and use the struct bools directly.

Hayden-IO:
Because the tests are in a different package, it is not trivial to unexport these methods. I'll dig more into this in a follow up. For now, I unexported two methods only used within the file.

codysoyland (approved):

Issue #439, opened 2025-03-19
Provide a common package for opentelemetry Instr.


miyunari:
**Description**

Initializing the OpenTelemetry tracing SDK results in relatively boilerplate code to determine which propagator (which must be identical across all components) is used, etc.

Following [#4101](https://github.com/sigstore/cosign/issues/4101), I think it would be a good idea to provide a common package for the initialization, which can be used in e.g. Fulcio and Cosign.

PR #440, opened 2025-03-20
util: provide common otel sdk initalization

miyunari:
#### Summary

Provide a common package for opentelemetry Instr.

- https://github.com/sigstore/sigstore-go/issues/439


loosebazooka:
Does it make more sense to have this in `sigstore/sigstore`? `sigstore-go` isn't necessarily imported into the infrastructure libraries, `sigstore/sigstore` is kind of where common shared code lives.

codysoyland:
@miyunari Thanks for the contribution! I do have some questions. I love more observability but I have a few concerns about adding this to the library.

- If this is intended to be a common func for use across multiple Sigstore projects (rekor, fulcio, tsa), this will need to live in a separate repo, as it will be a cyclic dependency otherwise (sigstore-go depends on the aforementioned repos). sigstore/sigstore is probably the most obvious place.
- It it necessary for this to live in a client library? The dependency bloat is something we've been trying to avoid. It makes more sense to me in the context of the Sigstore services like Fulcio, but I am not sure that typical users will benefit from this.
- Would this conflict with existing users' otel setups? It seems to set global otel settings, but users already using otel may have different needs. For example, at GitHub we use otel to trace timings and forward to our internal otel endpoints.
- If I'm not mistaken, the primary beneficiary would be the [operators of the public good instance](https://github.com/sigstore/sig-public-good-operations/tree/main). Have you spoken to anyone about the current needs of observability tooling in the public instance (@pwelch or @bobcallaway for example)?

miyunari:
@codysoyland Yes, you have to initialise the SDK itself. Something we have to do in all projects like rekor, fulcio and so on. Moving it to `sigstore/sigstore` sounds like a good idea.

From my point of view it would make sense to instrument the clients. This would allow users to track a whole transaction triggered by a CI/CD pipeline, for example. But I am happy to start with instrumenting the backend first and discuss this at a later point. wdyt?

> Would this conflict with existing users' otel setups? It seems to set global otel settings, but users already using otel may have different needs. For example, at GitHub we use otel to trace timings and forward to our internal otel endpoints.

The settings themselves should only affect one process that calls the initialisation method and is provided with valid settings for the SDK. Once other projects such as fulcio are instrumented, you should be able to configure the instance to send traces to your internal otel endpoints and use the internal traces there too.

> If I'm not mistaken, the primary beneficiary would be the [operators of the public good instance](https://github.com/sigstore/sig-public-good-operations/tree/main). Have you spoken to anyone about the current needs of observability tooling in the public instance (@pwelch or @bobcallaway for example)?

Not yet, this changes are mainly driven by our needs to be able to observe deployed Sigstore instances. But thanks for pointing me to them, I will reach out to :)

miyunari:
Closing this PR on the behalf of:
- https://github.com/sigstore/sigstore/pull/2063

PR #448, opened 2025-04-07
root: Fix trusted root creation with ed25519 keys

jku:
This is golang being weird: unlike all other keys ed25519.PublicKey is not a struct, just bytes.

Without this fix `cosign trusted-root create` fails when used with a rekor public key of type ED25519:

```
Error: failed constructing protobuf TrustRoot representation: 
failed converting rekor log b409400995a452b89ff59f6c388ae99356a4d48f25a1136e7e4efd37a97548d8 to protobuf:
failed converting public key to protobuf:
unknown public key type: ed25519.PublicKey
```

### DIFF
--- a/pkg/root/trusted_root_create.go
+++ b/pkg/root/trusted_root_create.go
@@ -260,7 +260,7 @@ func publicKeyToProtobufPublicKey(publicKey crypto.PublicKey, start time.Time, e
 		default:
 			return nil, fmt.Errorf("unsupported public modulus for RSA key: %d", p.Size())
 		}
-	case *ed25519.PublicKey:
+	case ed25519.PublicKey:
 		pkd.KeyDetails = protocommon.PublicKeyDetails_PKIX_ED25519
 	default:
 		return nil, fmt.Errorf("unknown public key type: %T", p)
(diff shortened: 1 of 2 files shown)


kommendorkapten (dismissed):

jku:
I suppose I could add a test: the existing tests could just test that a JSON round trip with the created trustroot works

kommendorkapten:
Yes, that would be good. I'm not sure how well-tested this functionality is? It would be good to have a trusted root with all the supported key types.

jku:
Modified tests so that one fails without the added fix. The test is a bit shallow as it does not actually compare the contents, only that json round trip (with different keytypes) succeeds... But better than nothing.

kommendorkapten:
To prevent regressions, can we add a Ed25519 key to the test suite? Either in this file https://github.com/sigstore/sigstore-go/blob/main/examples/trusted-root-public-good.json or add a new file?

jku:
> To prevent regressions, can we add a Ed25519 key to the test suite? Either in this file https://github.com/sigstore/sigstore-go/blob/main/examples/trusted-root-public-good.json or add a new file?

There is an ed25519 key in the root tests and  TestTrustedMaterialCollectionED25519 will fail if this regresses.

kommendorkapten (approved):

PR #452, opened 2025-04-11
don't return error if logIndex is 0

bobcallaway:
the first entry in a log is indexed at `0` so this is not an error; right now in proto we have this specified as an `int64` (even though a negative value is not possible and nonsensical) so changing this check to look for a negative number.

### DIFF
--- a/pkg/tlog/entry.go
+++ b/pkg/tlog/entry.go
@@ -101,7 +101,7 @@ func ParseEntry(protoEntry *v1.TransparencyLogEntry) (entry *Entry, err error) {
 	if protoEntry == nil ||
 		protoEntry.CanonicalizedBody == nil ||
 		protoEntry.IntegratedTime == 0 ||
-		protoEntry.LogIndex == 0 ||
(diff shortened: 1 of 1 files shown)


jku (approved):
Change lgtm.

The previous line got me thinking: I think integratedtime should be considered optional here (IOW zero should be ok) with the reasoning that this way the same proto is usable with rekor-tiles. @haydentherapper is this correct?

bobcallaway:
> Change lgtm.
> 
> The previous line got me thinking: I think integratedtime should be considered optional here (IOW zero should be ok) with the reasoning that this way the same proto is usable with rekor-tiles. @haydentherapper is this correct?

there have been discussions about removing that in a new version of the `TransparencyLogEntry` type (as well as changing `log_index` to `uint64` which also is more correct)

Hayden-IO (approved):
Agreed with Bob that we shouldn't make any piecemeal changes yet for Rekor v2 support in this library.

Issue #455, opened 2025-04-14
verify.Verify() should accept Context paramater


steiza:
**Description**

We are trying to make any breaking API changes before we cut a v1.0 release.

Similar to `sign.SignData()`, implementations of verifiers may make network calls, and so it would be helpful to have a context parameter.

For now this won't be wired up to anything, but this means we won't have to change the function signature of `verify.Verify()` in the future.

steiza:
In discussing #456 we determined this was not needed because we don't want verification making any network calls.

PR #456, opened 2025-04-14
Add context parameter to verify.Verify()

steiza:
#### Summary

See https://github.com/sigstore/sigstore-go/issues/455

Essentially we are trying to finish making major breaking API changes ahead of the v1.0 release, and we think in the future it will be useful to have a context parameter for verifying to either help with cases where verifying makes a network call, or as part of observability telemetry.

#### Release Note

* Adds context parameter to `verify.Verify()`

#### Documentation

N/A

### DIFF
--- a/docs/verification.md
+++ b/docs/verification.md
@@ -104,7 +104,7 @@ Then, we load the bundle and perform the verification:
 		panic(err)
 	}
 
-	result, err := sev.Verify(b, verify.NewPolicy(verify.WithArtifactDigest("sha512", digest), verify.WithCertificateIdentity(certID)))
+	result, err := sev.Verify(context.TODO(), b, verify.NewPolicy(verify.WithArtifactDigest("sha512", digest), verify.WithCertificateIdentity(certID)))
 	if err != nil {
 		panic(err)
 	}
@@ -180,6 +180,7 @@ Putting it together, the following script will verify the example bundle and pri
 package main
 
 import (
+	"context"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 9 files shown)


Hayden-IO:
When is Verify making a network call? Besides TUF, which happens in a separate thread, is it just telemetry that would trigger it? And wouldn't the context be passed to the telemetry library in another way?

codysoyland:
There are a few places we're using context.TODO in the verification code paths:
- [tlog.VerifyInclusion](https://github.com/sigstore/sigstore-go/blob/8617fb929f11e80343daf2bed73e12e13afd207c/pkg/tlog/entry.go#L253)
- [verify.verifyEnvelope](https://github.com/sigstore/sigstore-go/blob/8617fb929f11e80343daf2bed73e12e13afd207c/pkg/verify/signature.go#L139)

If we are not doing online verification, I guess tlog doesn't need it? I'm not sure why the dsse verifier needs it.

codysoyland:
I know this was my suggestion (adding context to Verify), but I concede there may not be a good reason to add it now that we are not doing online tlog verification. We can use functional options in the future to provide context if we find we need?

@miyunari might have an opinion on the telemetry need for it?

Hayden-IO:
I don't know why DSSE verification needs a context, though I haven't dug into it. Though I guess context might be needed if verifying with a KMS key, to pull down the pub key.

miyunari:
> I know this was my suggestion (adding context to Verify), but I concede there may not be a good reason to add it now that we are not doing online tlog verification. We can use functional options in the future to provide context if we find we need?
> 
> @miyunari might have an opinion on the telemetry need for it?

The OpenTelemetry SDK propagates in golang the entire trace context via the go context object. Looking into it, I think it would be a good idea if we want to trace some specific logic there or we do some network calls. Even when they are async, we may want to link the operation to the routine that does the call(s).

(Please keep in mind, I am not super familiar with the sigstore-go code yet).

steiza:
I guess thinking about this more, we don't want `verify.Verify()` to make any network calls, so we probably don't need `context.Context`.

> Though I guess context might be needed if verifying with a KMS key, to pull down the pub key.

I think the way we'd handle this is by having you fetch the KMS public keys and [add that into your trustedMaterial](https://github.com/sigstore/sigstore-go/blob/db889a109b0deb1494e03d6da27c8348a2d815da/examples/sigstore-go-verification/main.go#L169-L183), well before you call `verify.Verify()`.

Never mind then!

PR #458, opened 2025-04-16
Signing example improvements

### LINKED ISSUE #457: SigningConfig improvements
* Signing example does not support providing a signing config file (in the same way verify example supports trusted root). It should
* signing config should also support providing a trusted root (since it does verification after signing and we can't use the default trusted root with some other signing config)

jku:
Fixes #457 and improves singing example:
* Support `--signing-config <FILE>`  (similarly to verify supporting trustedroot)
* Also support `--trusted-root <FILE>` (so that the bundle can be verified after signing)

This PR requires #465.

### DIFF
--- a/examples/sigstore-go-signing/main.go
+++ b/examples/sigstore-go-signing/main.go
@@ -34,12 +34,21 @@ var idToken *string
 var intoto *bool
 var tsa *bool
 var rekor *bool
+var signingconfigPath string
+var trustedrootPath string
 
 func init() {
 	idToken = flag.String("id-token", "", "OIDC token to send to Fulcio")
 	intoto = flag.Bool("in-toto", false, "Content to sign is in-toto document")
 	tsa = flag.Bool("tsa", false, "Include signed timestamp from timestamp authority")
 	rekor = flag.Bool("rekor", false, "Including transparency log entry from Rekor")
+
+	flag.StringVar(&signingconfigPath, "signing-config", "", "Path to signingconfig JSON file")
+	flag.StringVar(&signingconfigPath, "s", "", "Path to signingconfig JSON file")
+
+	flag.StringVar(&trustedrootPath, "trusted-root", "", "Path to trusted root JSON file")
+	flag.StringVar(&trustedrootPath, "t", "", "Path to trusted root JSON file")
+
 	flag.Parse()
 	if flag.NArg() == 0 {
 		usage()
@@ -83,78 +92,95 @@ func main() {
 
 	opts := sign.BundleOptions{}
 
-	// Get trusted_root.json
-	fetcher := fetcher.DefaultFetcher{}
-	fetcher.SetHTTPUserAgent(util.ConstructUserAgent())
+	var signingConfig *root.SigningConfig
 
-	tufOptions := &tuf.Options{
-		Root:              tuf.StagingRoot(),
-		RepositoryBaseURL: tuf.StagingMirror,
-		Fetcher:           &fetcher,
-	}
-	tufClient, err := tuf.New(tufOptions)
-	if err != nil {
-		log.Fatal(err)
-	}
+	// A trusted root is not required but we will load one if
+	// * it is given as argument or
+	// * we are using default signing config (as in that case we know which trusted root to use)
+	if trustedrootPath != "" {
+		opts.TrustedRoot, err = root.NewTrustedRootFromPath(trustedrootPath)
+		if err != nil {
+			log.Fatal(err)
+		}
+	} else if signingconfigPath == "" {
+		// Get staging trusted_root.json by default
+		fetcher := fetcher.DefaultFetcher{}
+		fetcher.SetHTTPUserAgent(util.ConstructUserAgent())
 
-	trustedRoot, err := root.GetTrustedRoot(tufClient)
-	if err != nil {
-		log.Fatal(err)
-	}
-
-	// TODO: Uncomment once an updated v0.2 SigningConfig is distributed
-	// via TUF
-	// signingConfigPGI, err := root.GetSigningConfig(tufClient)
-
-	signingConfig, err := root.NewSigningConfig(
-		root.SigningConfigMediaType02,
-		// Fulcio URLs
-		[]root.Service{
-			{
-				URL:                 "https://fulcio.sigstage.dev",
-				MajorAPIVersion:     1,
-				ValidityPeriodStart: time.Now().Add(-time.Hour),
-				ValidityPeriodEnd:   time.Now().Add(time.Hour),
-			},
-		},
-		// OIDC Provider URLs
-		[]root.Service{
-			{
-				URL:                 "https://oauth2.sigstage.dev/auth",
-				MajorAPIVersion:     1,
-				ValidityPeriodStart: time.Now().Add(-time.Hour),
-				ValidityPeriodEnd:   time.Now().Add(time.Hour),
-			},
-		},
-		// Rekor URLs
-		[]root.Service{
-			{
-				URL:                 "https://rekor.sigstage.dev",
-				MajorAPIVersion:     1,
-				ValidityPeriodStart: time.Now().Add(-time.Hour),
-				ValidityPeriodEnd:   time.Now().Add(time.Hour),
-			},
-		},
-		root.ServiceConfiguration{
-			Selector: v1.ServiceSelector_ANY,
-		},
-		[]root.Service{
-			{
-				URL:                 "https://timestamp.githubapp.com/api/v1/timestamp",
-				MajorAPIVersion:     1,
-				ValidityPeriodStart: time.Now().Add(-time.Hour),
-				ValidityPeriodEnd:   time.Now().Add(time.Hour),
-			},
-		},
-		root.ServiceConfiguration{
-			Selector: v1.ServiceSelector_ANY,
-		},
-	)
-	if err != nil {
-		log.Fatal(err)
+		tufOptions := &tuf.Options{
+			Root:              tuf.StagingRoot(),
+			RepositoryBaseURL: tuf.StagingMirror,
+			Fetcher:           &fetcher,
+		}
+		tufClient, err := tuf.New(tufOptions)
+		if err != nil {
+			log.Fatal(err)
+		}
+		opts.TrustedRoot, err = root.GetTrustedRoot(tufClient)
+		if err != nil {
+			log.Fatal(err)
+		}
 	}
 
-	opts.TrustedRoot = trustedRoot
+	if signingconfigPath != "" {
+		signingConfig, err = root.NewSigningConfigFromPath(signingconfigPath)
+		if err != nil {
+			log.Fatal(err)
+		}
+	} else {
+		// TODO: Uncomment once an updated v0.2 SigningConfig is distributed
+		// via TUF
+		// signingConfigPGI, err := root.GetSigningConfig(tufClient)
+
+		// for now we hard code the staging services here
+		signingConfig, err = root.NewSigningConfig(
+			root.SigningConfigMediaType02,
+			// Fulcio URLs
+			[]root.Service{
+				{
+					URL:                 "https://fulcio.sigstage.dev",
+					MajorAPIVersion:     1,
+					ValidityPeriodStart: time.Now().Add(-time.Hour),
+					ValidityPeriodEnd:   time.Now().Add(time.Hour),
+				},
+			},
+			// OIDC Provider URLs
+			[]root.Service{
+				{
+					URL:                 "https://oauth2.sigstage.dev/auth",
+					MajorAPIVersion:     1,
+					ValidityPeriodStart: time.Now().Add(-time.Hour),
+					ValidityPeriodEnd:   time.Now().Add(time.Hour),
+				},
+			},
+			// Rekor URLs
+			[]root.Service{
+				{
+					URL:                 "https://rekor.sigstage.dev",
+					MajorAPIVersion:     1,
+					ValidityPeriodStart: time.Now().Add(-time.Hour),
+					ValidityPeriodEnd:   time.Now().Add(time.Hour),
+				},
+			},
+			root.ServiceConfiguration{
+				Selector: v1.ServiceSelector_ANY,
+			},
+			[]root.Service{
+				{
+					URL:                 "https://timestamp.sigstage.dev/api/v1/timestamp",
+					MajorAPIVersion:     1,
+					ValidityPeriodStart: time.Now().Add(-time.Hour),
+					ValidityPeriodEnd:   time.Now().Add(time.Hour),
+				},
+			},
+			root.ServiceConfiguration{
+				Selector: v1.ServiceSelector_ANY,
+			},
+		)
+		if err != nil {
+			log.Fatal(err)
+		}
+	}
 
 	if *idToken != "" {
 		fulcioURL, err := root.SelectService(signingConfig.FulcioCertificateAuthorityURLs(), []uint32{1}, time.Now())


examples/sigstore-go-signing/main.go
```diff
@@ -180,7 +188,7 @@ func main() {
 		}
 		for _, tsaURL := range tsaURLs {
 			tsaOpts := &sign.TimestampAuthorityOptions{
-				URL:     tsaURL,
+				URL:     fmt.Sprintf("%s/api/v1/timestamp", tsaURL),
```
jku: maybe this would make more sense as an implementation detail of TimestampAuthority itself but the API docstring seemed clear: it's the api endpoint, not base url

jku:
> I did not add any tests so far

seems like a broke some existing ones: will take a look tomorrow

pkg/root/signing_config_test.go
```diff
@@ -79,7 +79,7 @@ func TestSelectService(t *testing.T) {
 			URL:                 "url_no_end",
 			MajorAPIVersion:     2,
 			ValidityPeriodStart: farFuture,
-			ValidityPeriodEnd:   time.Time{},
+			ValidityPeriodEnd:   time.Unix(0, 0),
```
jku: this is actually quite surprising and we'll have to think this through: clients currently have to consider 1970-01-01 as a magic number that means the value was not set. 

That's really quite annoying, I wish we didn't use a wire format as a file format :(

jku: the end time is actually marked as "optional" (which in protobuf means _"you can check if the value was set or not"_)... I will look into this

jku: as expected I really hate how this makes the code look (suddenly we're dealing with pointers everywhere). But I guess this is what properly dealing with protobufs means... I'll include the fix as an independent commit, not sure if it makes sense

jku:
Last commit now handles `valid_for.end` as an _optional_ value (to make sure we know when the value has not been set). It's not pretty but I think it's a little more correct. 

I'm also happy to just drop the last commit: the code does work without it, it just then considers 1970-01-01 as a magic end date value that means "no end date specified"

examples/sigstore-go-signing/main.go
```diff
@@ -83,79 +86,87 @@ func main() {
...
-		},
-		root.ServiceConfiguration{
-			Selector: v1.ServiceSelector_ANY,
-		},
-	)
-	if err != nil {
-		log.Fatal(err)
+			[]root.Service{
+				{
+					URL:                 "https://timestamp.sigstage.dev/api/v1/timestamp",
```
jku, line 170: this should really be "https://timestamp.sigstage.dev" only I think

examples/sigstore-go-signing/main.go
```diff
@@ -34,12 +34,15 @@ var idToken *string
...
 var tsa *bool
 var rekor *bool
+var signingconfigJSONpath *string
 
 func init() {
 	idToken = flag.String("id-token", "", "OIDC token to send to Fulcio")
 	intoto = flag.Bool("in-toto", false, "Content to sign is in-toto document")
 	tsa = flag.Bool("tsa", false, "Include signed timestamp from timestamp authority")
 	rekor = flag.Bool("rekor", false, "Including transparency log entry from Rekor")
+	signingconfigJSONpath = flag.String("signingconfigJSONpath", "", "Path to signingconfig JSON file")
```
Hayden-IO: I'd have a shorter flag name and use `-` rather than camel case, maybe `--signing-config-path`?

Hayden-IO:
```suggestion
	signingconfigJSONPath = flag.String("signingconfigJSONpath", "", "Path to signingconfig JSON file")
```

examples/sigstore-go-signing/main.go
```diff
@@ -83,79 +86,87 @@ func main() {
...
-		log.Fatal(err)
-	}
-
-	trustedRoot, err := root.GetTrustedRoot(tufClient)
-	if err != nil {
-		log.Fatal(err)
-	}
+	if *signingconfigJSONpath != "" {
+		signingConfig, err = root.NewSigningConfigFromPath(*signingconfigJSONpath)
+		// NOTE: trusted root is not set here (could be taken with another flag)
```
Hayden-IO: In this case, we would no longer pull in a trusted root, which IIRC is now required for signing to verify the results. Could we use this same pattern of providing either a flag or pulling from TUF, something like:

```
if signingconfigJSONPath != "" {
   // read from file
} else {
  // read from TUF
}

if trustedRootJSONPath != "" {
  // read from file
} else {
 // read from TUF
}


```

examples/sigstore-go-signing/main.go
Hayden-IO: We require the [full URL](https://github.com/sigstore/sigstore-go/blob/db889a109b0deb1494e03d6da27c8348a2d815da/pkg/sign/timestamping.go#L81). This is because RFC3161 (unfortunately) does not define a standard API path for requesting timestamps, and so if we want to support other TSAs besides our sigstore/timestamp-authority implementation, we need the full path.

examples/sigstore-go-signing/main.go
Hayden-IO: As mentioned above, we need to take in the full URL path to support other TSA implementations, so this should still be `tsaURL`.

pkg/root/signing_config.go
```diff
@@ -38,7 +38,7 @@ type Service struct {
 	URL                 string
 	MajorAPIVersion     uint32
 	ValidityPeriodStart time.Time
-	ValidityPeriodEnd   time.Time
+	ValidityPeriodEnd   *time.Time
```
Hayden-IO: It's recommend to pass `time.Time` as a struct, and check `time.IsZero` to see if it's initialized - https://pkg.go.dev/time#Time

We follow this same pattern in trusted root, checking if it's [zero](https://github.com/sigstore/sigstore-go/blob/db889a109b0deb1494e03d6da27c8348a2d815da/pkg/tlog/entry.go#L282) on use.

jku: `isZero()` does not work: that's the bug I'm fixing:
 * protobuf default value `Timestamp{0, 0}` means "1970-01-01", golang zero time `time.Time{0, 0}` means "01-01-01"
 * We can keep using the magic value approach, but must then use "1970-01-01" as the magic value which seems a bit icky. I can drop the last commit to achieve that.

Hayden-IO: Oh I see. Do we have the same bug in TrustedRoot then?

jku: I'll check

examples/sigstore-go-signing/main.go
jku: > In this case, we would no longer pull in a trusted root, which IIRC is now required for signing to verify the results.

The signing works, it just does not verify in this case.

> Could we use this same pattern of providing either a flag or pulling from TUF,

I think this TUF suggestion does not work: if user gives us a random signingconfig we should not expect staging trustedroot to verify the resulting signature bundle.  Trust root and signing config should always come from same source.


I can add another flag so user could provide a trust root as well.

examples/sigstore-go-signing/main.go
jku: sure. I think I copied what verify example does with trust root (`-trustedrootJSONpath`), but I can change both of them

pkg/root/signing_config.go
jku: I think the end timestamp typically does not matter in  verification use case: even if we are past the `ValidityPeriodEnd` of a specific tlog, we still want to use the log to verify entries from that log.

I'm not sure there is any situation where a normal client is actually interested in `ValidityPeriodEnd` from trustedroot.

jku: oh! once we add timestamp into trustedroot, clients are interested in ValidityPeriodEnd for that. So yes, we do have the same bug in verification.  I've confirmed  verification currently fails if 
* trust root contains a TSA and
* the entry contains a timestamp

I've not looked at details yet though...

examples/sigstore-go-signing/main.go
jku: Added a flag` --trusted-root`

jku:
I'm working on this tomorrow but I think I'll split this into two: bug fix and the feature work...

jku:
The bug in ValidFor handling is handled in the linked PR: I'm leaving this PR for the feature work that follows the bug fix.

I'm still seeing a strange failure in the verification step in `examples/sigstore-go-signing` when a TSA is included in the signingconfig: I will document it here later today.

jku:
Here is the remaining issue I see:

**background**
* I'm using the signing example with signingconfig and trustedroot files from https://github.com/sigstore/root-signing-staging/pull/244 -- I'm trying to test the next *staging* signing config and trustedroot before we add them into the TUF repository
* the signingconfig and trustedroot include a TSA

**results**
 
 Running the signing-example with just the signingconfig seems to work:
 
```bash
 $ ./sigstore-go-signing --tsa --rekor -id-token $TOKEN --signing-config ~/src/root-signing-staging/targets/signing_config_0_2.json  hello_world.txt 
Using public key:

-----BEGIN PUBLIC KEY-----
MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAE6CLoAIFCWshnK9+qdzFcWi3+FsE6
/TIdzstUtpA/CTBFkgd3KFW6/8cCKx0QrI1I1NZSz3WIH3xIcuHhBspf2Q==
-----END PUBLIC KEY-----


{"mediaType":"application/vnd.dev.sigstore.bundle.v0.3+json", "verificationMaterial":{"certificate":{"rawBytes":"MIICzDCCAlGgAwIBAgIUD0nVfEiMHNzJKcw7Ms4Yu9Wc9IUwCgYIKoZIzj0EAwMwNzEVMBMGA1UEChMMc2lnc3RvcmUuZGV2MR4wHAYDVQQDExVzaWdzdG9yZS1pbnRlcm1lZGlhdGUwHhcNMjUwNDI0MTQwOTQ0WhcNMjUwNDI0MTQxOTQ0WjAAMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAE6CLoAIFCWshnK9+qdzFcWi3+FsE6/TIdzstUtpA/CTBFkgd3KFW6/8cCKx0QrI1I1NZSz3WIH3xIcuHhBspf2aOCAXAwggFsMA4GA1UdDwEB/wQEAwIHgDATBgNVHSUEDDAKBggrBgEFBQcDAzAdBgNVHQ4EFgQUYJ/dxDrreOaYYfp6A6CIcf/dYX0wHwYDVR0jBBgwFoAUcYYwphR8Ym/599b0BRp/X//rb6wwGQYDVR0RAQH/BA8wDYELamt1QGdvdG8uZmkwLAYKKwYBBAGDvzABAQQeaHR0cHM6Ly9naXRodWIuY29tL2xvZ2luL29hdXRoMC4GCisGAQQBg78wAQgEIAweaHR0cHM6Ly9naXRodWIuY29tL2xvZ2luL29hdXRoMIGLBgorBgEEAdZ5AgQCBH0EewB5AHcAKzC83GiIyeLh2CYpXnQfSDkxlgLynDPLXkNA/rKshnoAAAGWaCINeAAABAMASDBGAiEA0y8FZzhVvFOn8IAbUprJsmtFt7RSLGqI5Hd+csFQBiACIQDL0zuCyO8mny6l/sRWlAfo3eHt331oB9AmXTuqzwJiIjAKBggqhkjOPQQDAwNpADBmAjEA53S44PXGDsFERYpn+b0i++L/c3GaLAgSCDUNCCa5Y82ZJ16U3PylYU9DVIVMtjrPAjEAzM5yO4wPNXTUkT7P+Refy7NcrbIytNBHtSELZZQCWAX+iOUZM43xiAN+kYi0YZx0"}, "tlogEntries":[{"logIndex":"41223928", "logId":{"keyId":"0y8wo8MtY5wrdiIFohx7sHeI5oKDpK5vQhGHI6G+pJY="}, "kindVersion":{"kind":"hashedrekord", "version":"0.0.1"}, "integratedTime":"1745503784", "inclusionPromise":{"signedEntryTimestamp":"MEQCIGXw+c1qm/wZ+r0PHci4zXaCKXZJK6X0jW7ngXThTSeLAiAIzWQLPaAg19ocfjfTsBJgvdssD+DyLNtehHcBrTGVbQ=="}, "inclusionProof":{"logIndex":"9541516", "rootHash":"nBsLUmG61wfyG/epi/j2Eg7zE/YuseKi3bnUIbIQfSE=", "treeSize":"9541517", "hashes":["5VtXgxznIhXYCl9KeOfcv20ihITVBd4/WLXA+i5xWaE=", "B6UqFLsE9IJ//b7/JS3GARryY3xjBiWr5xU8vs0CmOI=", "Llj6SdgRE5sRer95msj+rzDjiRI4PFJndcpDqkG+IFk=", "O2XaZfUGIee6bMZWvo8haWG3QV0OuQNA5z/Z1q5Lisc=", "DMHH0f5alupejso2qvNvVRJ/XvlIBGBOf3YGltmaQMI=", "ZC8CFCObTb0ofLpLSxhg8erNiSjMItnCVQY0sXPBg0E=", "FySOP2K+cnf8w7rq2oX/MlXshQka2y+MJ0Bffe5EHv4=", "/R3TBYaRYRzbN32wMW3eh/hoXTU8qkZcYMRf2xj5w3E=", "z4/Qhzab/IOndrkc6duUjoZbQq/ELELHRdEvlG7MTJU=", "GoHRwlhYuJIYJdmRnHX5HWLr2ngxzHnAIIqBewovBi0=", "OdoqbUqBYHhj2W1RLM8APkQOnM2K9gzGm1KPFmwIIeQ="], "checkpoint":{"envelope":"rekor.sigstage.dev - 8202293616175992157\n9541517\nnBsLUmG61wfyG/epi/j2Eg7zE/YuseKi3bnUIbIQfSE=\n\n— rekor.sigstage.dev 0y8wozBGAiEAq4++5Yc/3/Aw27ZnEaHyYZiyzBZXnGmuF5Wegk2xOZECIQC1d16gBsugA/o4nlgyiT222cu/ZXfeArLmcplrebZhYg==\n"}}, "canonicalizedBody":"eyJhcGlWZXJzaW9uIjoiMC4wLjEiLCJraW5kIjoiaGFzaGVkcmVrb3JkIiwic3BlYyI6eyJkYXRhIjp7Imhhc2giOnsiYWxnb3JpdGhtIjoic2hhMjU2IiwidmFsdWUiOiJiOTRkMjdiOTkzNGQzZTA4YTUyZTUyZDdkYTdkYWJmYWM0ODRlZmUzN2E1MzgwZWU5MDg4ZjdhY2UyZWZjZGU5In19LCJzaWduYXR1cmUiOnsiY29udGVudCI6Ik1FUUNJRkF1YW9KaFhzZm53aC9uNzJRTGlCRGVhWGZQR3hSVlNCbXMzZFp4T0M5VkFpQUEyT0NjSFAwdWlvK21ZYUZlZGhnbkNkbmtwY3E2Q1YvWHBXL2pKZnhUa2c9PSIsInB1YmxpY0tleSI6eyJjb250ZW50IjoiTFMwdExTMUNSVWRKVGlCRFJWSlVTVVpKUTBGVVJTMHRMUzB0Q2sxSlNVTjZSRU5EUVd4SFowRjNTVUpCWjBsVlJEQnVWbVpGYVUxSVRucEtTMk4zTjAxek5GbDFPVmRqT1VsVmQwTm5XVWxMYjFwSmVtb3dSVUYzVFhjS1RucEZWazFDVFVkQk1WVkZRMmhOVFdNeWJHNWpNMUoyWTIxVmRWcEhWakpOVWpSM1NFRlpSRlpSVVVSRmVGWjZZVmRrZW1SSE9YbGFVekZ3WW01U2JBcGpiVEZzV2tkc2FHUkhWWGRJYUdOT1RXcFZkMDVFU1RCTlZGRjNUMVJSTUZkb1kwNU5hbFYzVGtSSk1FMVVVWGhQVkZFd1YycEJRVTFHYTNkRmQxbElDa3R2V2tsNmFqQkRRVkZaU1V0dldrbDZhakJFUVZGalJGRm5RVVUyUTB4dlFVbEdRMWR6YUc1TE9TdHhaSHBHWTFkcE15dEdjMFUyTDFSSlpIcHpkRlVLZEhCQkwwTlVRa1pyWjJRelMwWlhOaTg0WTBOTGVEQlJja2t4U1RGT1dsTjZNMWRKU0RONFNXTjFTR2hDYzNCbU1tRlBRMEZZUVhkblowWnpUVUUwUndwQk1WVmtSSGRGUWk5M1VVVkJkMGxJWjBSQlZFSm5UbFpJVTFWRlJFUkJTMEpuWjNKQ1owVkdRbEZqUkVGNlFXUkNaMDVXU0ZFMFJVWm5VVlZaU2k5a0NuaEVjbkpsVDJGWldXWndOa0UyUTBsalppOWtXVmd3ZDBoM1dVUldVakJxUWtKbmQwWnZRVlZqV1ZsM2NHaFNPRmx0THpVNU9XSXdRbEp3TDFndkwzSUtZalozZDBkUldVUldVakJTUVZGSUwwSkJPSGRFV1VWTVlXMTBNVkZIWkhaa1J6aDFXbTFyZDB4QldVdExkMWxDUWtGSFJIWjZRVUpCVVZGbFlVaFNNQXBqU0UwMlRIazVibUZZVW05a1YwbDFXVEk1ZEV3eWVIWmFNbXgxVERJNWFHUllVbTlOUXpSSFEybHpSMEZSVVVKbk56aDNRVkZuUlVsQmQyVmhTRkl3Q21OSVRUWk1lVGx1WVZoU2IyUlhTWFZaTWpsMFRESjRkbG95YkhWTU1qbG9aRmhTYjAxSlIweENaMjl5UW1kRlJVRmtXalZCWjFGRFFrZ3dSV1YzUWpVS1FVaGpRVXQ2UXpnelIybEplV1ZNYURKRFdYQllibEZtVTBScmVHeG5USGx1UkZCTVdHdE9RUzl5UzNOb2JtOUJRVUZIVjJGRFNVNWxRVUZCUWtGTlFRcFRSRUpIUVdsRlFUQjVPRVphZW1oV2RrWlBiamhKUVdKVmNISktjMjEwUm5RM1VsTk1SM0ZKTlVoa0syTnpSbEZDYVVGRFNWRkVUREI2ZFVONVR6aHRDbTU1Tm13dmMxSlhiRUZtYnpObFNIUXpNekZ2UWpsQmJWaFVkWEY2ZDBwcFNXcEJTMEpuWjNGb2EycFBVRkZSUkVGM1RuQkJSRUp0UVdwRlFUVXpVelFLTkZCWVIwUnpSa1ZTV1hCdUsySXdhU3NyVEM5ak0wZGhURUZuVTBORVZVNURRMkUxV1RneVdrb3hObFV6VUhsc1dWVTVSRlpKVmsxMGFuSlFRV3BGUVFwNlRUVjVUelIzVUU1WVZGVnJWRGRRSzFKbFpuazNUbU55WWtsNWRFNUNTSFJUUlV4YVdsRkRWMEZZSzJsUFZWcE5ORE40YVVGT0sydFphVEJaV25nd0NpMHRMUzB0UlU1RUlFTkZVbFJKUmtsRFFWUkZMUzB0TFMwSyJ9fX19"}], "timestampVerificationData":{"rfc3161Timestamps":[{"signedTimestamp":"MIICyjADAgEAMIICwQYJKoZIhvcNAQcCoIICsjCCAq4CAQMxDTALBglghkgBZQMEAgEwgbgGCyqGSIb3DQEJEAEEoIGoBIGlMIGiAgEBBgkrBgEEAYO/MAIwMTANBglghkgBZQMEAgEFAAQgrCVlhEfTP6ohyMTlPUXQ3qUcVh8cdj9lU1Qk/x+EpPoCFQDGwni5BYaccCC+sfJLHJ3Xm3a3QxgPMjAyNTA0MjQxNDA5NDRaMAMCAQGgMqQwMC4xFTATBgNVBAoTDHNpZ3N0b3JlLmRldjEVMBMGA1UEAxMMc2lnc3RvcmUtdHNhoAAxggHbMIIB1wIBATBRMDkxFTATBgNVBAoTDHNpZ3N0b3JlLmRldjEgMB4GA1UEAxMXc2lnc3RvcmUtdHNhLXNlbGZzaWduZWQCFAo1oQZh1eJBc8aJlqfyffJ+A3ynMAsGCWCGSAFlAwQCAaCB/DAaBgkqhkiG9w0BCQMxDQYLKoZIhvcNAQkQAQQwHAYJKoZIhvcNAQkFMQ8XDTI1MDQyNDE0MDk0NFowLwYJKoZIhvcNAQkEMSIEICPo7q3CvNGVkc+Rz2QF0gCzXGggFXKyWjzwRCMrO26VMIGOBgsqhkiG9w0BCRACLzF/MH0wezB5BCAG9P/gR/6zWZm3M7DXoyNQHPwY5MAzZqhF13U250snRDBVMD2kOzA5MRUwEwYDVQQKEwxzaWdzdG9yZS5kZXYxIDAeBgNVBAMTF3NpZ3N0b3JlLXRzYS1zZWxmc2lnbmVkAhQKNaEGYdXiQXPGiZan8n3yfgN8pzAKBggqhkjOPQQDAgRnMGUCMA2GdvoiqRo0axmjHhkBjHdz+JLi5l/+aX5yrUA9Jl6qm49p462M0+Fc48w8E4V3wAIxAJluBUYMk1CVWtNG8ZpJsp7zFnaGSQu9m70JU4bWwvGGQ4s2gcUpyn2HMzYLzXYM1A=="}]}}, "messageSignature":{"messageDigest":{"algorithm":"SHA2_256", "digest":"uU0nuZNNPgilLlLX2n2r+sSE7+N6U4DukIj3rOLvzek="}, "signature":"MEQCIFAuaoJhXsfnwh/n72QLiBDeaXfPGxRVSBms3dZxOC9VAiAA2OCcHP0uio+mYaFedhgnCdnkpcq6CV/XpW/jJfxTkg=="}}
```

The resulting bundle passes the verify example:

```bash
$ ./sigstore-go-verification -artifact ../sigstore-go-signing/hello_world.txt -requireTimestamp -expectedSAN jku@goto.fi -expectedIssuer https://github.com/login/oauth -trustedrootJSONpath ~/src/root-signing-staging/targets/trusted_root.json ../sigstore-go-signing/bundle.json 
Verification successful!
{
   "mediaType": "application/vnd.dev.sigstore.verificationresult+json;version=0.1",
   "signature": {
      "certificate": {
         "certificateIssuer": "CN=sigstore-intermediate,O=sigstore.dev",
         "subjectAlternativeName": "jku@goto.fi",
         "issuer": "https://github.com/login/oauth"
      }
   },
   "verifiedTimestamps": [
      {
         "type": "Tlog",
         "uri": "https://rekor.sigstage.dev",
         "timestamp": "2025-04-24T17:09:44+03:00"
      }
   ],
   "verifiedIdentity": {
      "subjectAlternativeName": {
         "subjectAlternativeName": "jku@goto.fi"
      },
      "issuer": {
         "issuer": "https://github.com/login/oauth"
      }
   },
   "statement": {}
}
```

the issue is that running the signing example with a trusted root (enabling some verification after the signing) fails:

```
$ ./sigstore-go-signing --tsa --rekor -id-token $TOKEN --signing-config ~/src/root-signing-staging/targets/signing_config_0_2.json --trusted-root ~/src/root-signing-staging/targets/trusted_root.json hello_world.txt 
Using public key:

-----BEGIN PUBLIC KEY-----
MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAExtzu0NNdtnFEmLEGSChSXYo5AcJ7
tq282LMifHmdhnBSFCtqrgCjxbXzH8QAzAbki81mowFvD4U6LuQha7dQJg==
-----END PUBLIC KEY-----


2025/04/24 17:05:57 failed to verify timestamps: threshold not met for verified signed timestamps: 0 < 1
```

* In both cases (sigstore-go-verify and sigstore-go-signing with --trusted-root) the code calls verifySignedTimestamp (https://github.com/sigstore/sigstore-go/blob/main/pkg/verify/tsa.go#L58) and the result is an error
* sigstore-go-verify does not seem to think the error is a problem -- it even lists the timestamp as verified in the output

I am not sure where the issue is -- sigstore-go, my trusted root or my signing config.

jku:
> it even lists the timestamp as verified in the output

Actually that "timestamp" might be the integrated time since it says "type: TLog"?

jku:
oh wait, my timestamp URL in signingconfig is "https://timestamp.sigstage.dev" and I think that URL was supposed to be the actual API endpoint... Why does the signing seemingly work then?

jku:
Documenting the debugging so far:
the timestamp verification fails because PKCS7.VerifyWithOpts() (https://github.com/sigstore/timestamp-authority/blob/main/pkg/verification/verify.go#L290 ) fails with `pkcs7: No certificate for signer`. The pkcs7 code expects to get a certificate with a specific issuer name and serial number: issuer matches but the serial number does not

loosebazooka:
Just an update to keep people from investigating this further. I think the issue here is that the wrong TSA cert is embedded in https://github.com/sigstore/root-signing-staging/pull/244/files#diff-ca7959d2b4378c4b4d637c1ab78718347f9c6381b239c38f10f4d9f31a78843eR97-R101

Changing this to the one from the staging TSA: http://timestamp.sigstage.dev/api/v1/timestamp/certchain appears to work for me.

jku:
> Changing this to the one from the staging TSA: http://timestamp.sigstage.dev/api/v1/timestamp/certchain appears to work for me.

Thank you :1st_place_medal: 

The whole thing seems to work fine now: and the output in verification example looks correct:

```
Verification successful!
{
   "mediaType": "application/vnd.dev.sigstore.verificationresult+json;version=0.1",
   "signature": {
      "certificate": {
         "certificateIssuer": "CN=sigstore-intermediate,O=sigstore.dev",
         "subjectAlternativeName": "jku@goto.fi",
         "issuer": "https://github.com/login/oauth"
      }
   },
   "verifiedTimestamps": [
      {
         "type": "Tlog",
         "uri": "https://rekor.sigstage.dev",
         "timestamp": "2025-04-25T09:37:19+03:00"
      },
      {
         "type": "TimestampAuthority",
         "uri": "https://timestamp.sigstage.dev/api/v1/timestamp",
         "timestamp": "2025-04-25T06:37:19Z"
      }
   ],
   "verifiedIdentity": {
      "subjectAlternativeName": {
         "subjectAlternativeName": "jku@goto.fi"
      },
      "issuer": {
         "issuer": "https://github.com/login/oauth"
      }
   },
   "statement": {}
}
```

I think even the fact that verify and sign examples worked differently (one succeeds and one fails loudly) is understandable: 
* verify found a rekorv1 entry with a SET: that's enough "timestamps"
* signing specifically checks for the timestamp authority timestamp because it knows one was added

jku:
rebased on main since #465 was merged: ready for review

codysoyland (approved):
LGTM!

Hayden-IO (approved):
Sorry for not merging, thank you for cleaning this up!

PR #459, opened 2025-04-18
Select highest API version with multiple SigningConfig services

### LINKED ISSUE #453: SigningConfig should select the highest API version when possible
**Description**

**Version**

Right now, when selecting a service or set of services, the SigningConfig will select the first match given a validity period and API version. If the client supports two API versions concurrently (e.g. Rekor v1 and Rekor v2) and the distributed signing config has entries for multiple API versions (which it will for Rekor), then the client may select an older API version depending on the sorting of the services.

I will also update the language in protobuf-specs to be more prescriptive on how clients should select a service given support for multiple API versions.

Hayden-IO:
Right now, when selecting a service or set of services, the SigningConfig will select the first match given a validity period and API version. If the client supports two API versions concurrently (e.g. Rekor v1 and Rekor v2) and the distributed signing config has entries for multiple API versions (which it will for Rekor), then the client may select an older API version depending on the sorting of the services.

This change sorts the supported API versions to select the highest API version first.

Fixes #453

### DIFF
--- a/pkg/root/signing_config.go
+++ b/pkg/root/signing_config.go
@@ -47,15 +47,27 @@ type ServiceConfiguration struct {
 }
 
 // SelectService returns which service endpoint should be used based on supported API versions
-// and current time. It will select the first service that matches the criteria. Services should
-// be sorted from newest to oldest validity period start time, to minimize how far clients
-// need to search to find a matching service.
+// and current time. It will select the first service with the highest API version that matches
+// the criteria. Services should be sorted from newest to oldest validity period start time, to
+// minimize how far clients need to search to find a matching service.
 func SelectService(services []Service, supportedAPIVersions []uint32, currentTime time.Time) (string, error) {
-	for _, s := range services {
-		if slices.Contains(supportedAPIVersions, s.MajorAPIVersion) && s.ValidAtTime(currentTime) {
-			return s.URL, nil
+	if len(supportedAPIVersions) == 0 {
+		return "", fmt.Errorf("no supported API versions")
+	}
+
+	sortedVersions := make([]uint32, len(supportedAPIVersions))
+	copy(sortedVersions, supportedAPIVersions)
+	slices.Sort(sortedVersions)
+	slices.Reverse(sortedVersions)
+
+	for _, version := range sortedVersions {
+		for _, s := range services {
+			if version == s.MajorAPIVersion && s.ValidAtTime(currentTime) {
+				return s.URL, nil
+			}
 		}
 	}
+
 	return "", fmt.Errorf("no matching service found for API versions %v and current time %v", supportedAPIVersions, currentTime)
 }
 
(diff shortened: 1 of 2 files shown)


Hayden-IO:
FYI @jku

pkg/root/signing_config.go
```diff
@@ -46,48 +47,89 @@ type ServiceConfiguration struct {
...
-	for _, s := range services {
-		if slices.Contains(supportedAPIVersions, s.MajorAPIVersion) && s.ValidAtTime(currentTime) {
-			return s.URL, nil
+	if len(supportedAPIVersions) == 0 {
+		return "", fmt.Errorf("no supported API versions")
+	}
+
+	sortedVersions := make([]uint32, len(supportedAPIVersions))
+	copy(sortedVersions, supportedAPIVersions)
+	sort.Sort(sort.Reverse(uint32Slice(sortedVersions)))
```
codysoyland: You can use the generic slices package to remove `uint32Slice`:

```go
	slices.Sort(sortedVersions)
	slices.Reverse(sortedVersions)
```

Hayden-IO: Thanks, updated!

pkg/root/signing_config.go
```diff
@@ -47,47 +47,83 @@ type ServiceConfiguration struct {
...
-	var urls []string
+	if len(supportedAPIVersions) == 0 {
+		return nil, fmt.Errorf("no supported API versions")
+	}
+
+	urlsByVersion := make(map[uint32][]string)
 	for _, s := range services {
 		if slices.Contains(supportedAPIVersions, s.MajorAPIVersion) && s.ValidAtTime(currentTime) {
-			urls = append(urls, s.URL)
+			urls, ok := urlsByVersion[s.MajorAPIVersion]
```
kommendorkapten: You can do: 
```go
urlsByVersion[s.MajorAPIVersion] = append(urlsByVersion[s.MajorAPIVersion], s.URL)
```

Which is a bit simpler, see https://go.dev/play/p/tiUTV34saTE

pkg/root/signing_config.go
```diff
@@ -47,47 +47,83 @@ type ServiceConfiguration struct {
...
+			if !ok {
+				urlsByVersion[s.MajorAPIVersion] = []string{s.URL}
+			} else {
+				urls = append(urls, s.URL)
+				urlsByVersion[s.MajorAPIVersion] = urls
+			}
```
kommendorkapten:
```suggestion
			urlsByVersion[s.MajorAPIVersion] = append(urlsByVersion[s.MajorAPIVersion], s.URL)
```

pkg/root/signing_config.go
Hayden-IO: TIL, thanks!

pkg/root/signing_config.go
codysoyland: Seems not to be updated, maybe you didn't push?

Hayden-IO: I think a git reset and git push -f wiped out a commit, oops! Added back.

kommendorkapten (approved):
Thanks!

PR #465, opened 2025-04-24
Fix SigningConfig ValidFor when dates are missing

### LINKED ISSUE #464: Validity periods are handled incorrectly
This was found during #457: The handling for ValidFor time ranges is incorrect in multiple places

* The code uses `Time.IsZero()` as a proxy for _"was the time stamp value unset in the protobuf?"_
* This is incorrect because IsZero compares to date 01-01-01 whereas the the protobuf Timestamp default value is 1970-01-01
* This issue impacts all ValidFor fields in SigningConfig
* In verification side all values are already changed in serialization: missing dates become 01-01-01

**Potential fixes:** 

* The minimal fix for this would be to consider 19070-01-01 a magic date value and replace all checks for `IsZero()` with a comparison to the magic date
* We can do the same as verify does: set date to 01-01-01 if date is missing in serialization **<- this was selected**
* A more reasonable but larger fix would be to create a TimeRange type that can handle unset start and end dates and manages the comparison

 I will have a look at how big of a job the latter one is

jku:
Fixes #464: Make sure we handle missing dates in ValidFor protobuf as a special case. The time that we deserialize in that case needs to be 01-01-01 and not the protobuf default 1970-01-01 because we want to be able to check for it with `Time.IsZero()`

### DIFF
--- a/pkg/root/signing_config.go
+++ b/pkg/root/signing_config.go
@@ -46,6 +46,16 @@ type ServiceConfiguration struct {
 	Count    uint32
 }
 
+func NewService(s *prototrustroot.Service) Service {
+	validFor := s.GetValidFor()
+	return Service{
+		URL:                 s.GetUrl(),
+		MajorAPIVersion:     s.GetMajorApiVersion(),
+		ValidityPeriodStart: validFor.GetStart().AsTime(),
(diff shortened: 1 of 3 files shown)


Hayden-IO (approved):
Thanks for finding and fixing this!

PR #470, opened 2025-05-07
Disable TUF timestamping when TUF cache disabled

codysoyland:

#### Summary

I noticed that the TUF client will write timestamp files into CWD when the TUF cache is disabled. The intent of these files is to track the last TUF update time, but when the cache is disabled, these files should not be persisted. This PR modifies the logic so the last updated time is not persisted when the cache is disabled.

### DIFF
--- a/pkg/tuf/client.go
+++ b/pkg/tuf/client.go
@@ -150,7 +150,10 @@ func (c *Client) Refresh() error {
 	if err != nil {
 		return fmt.Errorf("tuf refresh failed: %w", err)
 	}
-
+	// If cache is disabled, we don't need to persist the last timestamp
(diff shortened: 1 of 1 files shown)


Hayden-IO:
Did you have more information about this?

kommendorkapten (approved):

codysoyland:
> Did you have more information about this?

I updated the description. The last updated time is always persisted when doing TUF updates currently, so this PR disables writing the last updated time if the cache is disabled.

rdimitrov:
I just hit this today but with an older version of sigstore-go, thanks for fixing it in v1.0.0 😃

PR #472, opened 2025-05-07
Prevent duplicate timestamps from same TSA

### LINKED ISSUE #471: Multiple timestamps from same TSA are allowed
**Description**

The [TSA duplication detection](https://github.com/sigstore/sigstore-go/blob/8dff965/pkg/verify/tsa.go#L40-L47) compares the bytes of each timestamp response. This would allow a compromised TSA to mint multiple timestamp responses to meet the threshold.

I think it's reasonable to assume that multiple timestamps MUST come from different authorities, so I recommend we do not allow duplicate timestamps from the same authority.

**Version**

codysoyland:
#### Summary

This modifies the duplicate detection so it does not compare the bytes of the timestamp, but instead the authority.

This prevents a single compromised TSA from being used to meet a threshold of greater than 1 by duplicating timestamps.

Fixes https://github.com/sigstore/sigstore-go/issues/471

### DIFF
--- a/pkg/testing/ca/ca.go
+++ b/pkg/testing/ca/ca.go
@@ -69,7 +69,10 @@ type VirtualSigstore struct {
 }
 
 func NewVirtualSigstoreWithSigningAlg(signingKeyDetails v1.PublicKeyDetails) (*VirtualSigstore, error) {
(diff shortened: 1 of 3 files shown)


kommendorkapten (approved):
Nice.

steiza (approved):
LGTM!

Hayden-IO (commented):
Looks great!

PR #473, opened 2025-05-08
Add verification errors to output of VerifyTimestampAuthority

codysoyland:
#### Summary

This is an alternative to https://github.com/sigstore/sigstore-go/pull/325 which modifies the return value of `VerifyTimestampAuthority` to include verification errors separately from non-verification errors.


steiza (approved):
I like this! It'll definitely help us surface non-critical failures to users so they can do debugging or better understand what is going on, like for #48.

Hayden-IO (approved):
Thanks for working on two approaches. I like this one more, since it fits well into how we are currently verifying bundles. We can also update Rekor to do the same.

PR #476, opened 2025-05-08
Rename and deprecate SignedEntityVerifier in favor of Verifier

codysoyland:

#### Summary

On the road to 1.0, now is our chance to clean up any naming mistakes we made along the way. `SignedEntityVerifier` was intially used to distinguish from other "types" of verifier, but it feels awkward as the main entrypoint for verification. This PR renames `NewSignedEntityVerifier` and `SignedEntityVerifier` to `NewVerifier` and `Verifier`, respectively, for a more friendly developer interface.

Aliases for the old types are added to allow a deprecation period, and these aliases should be removed before sigstore-go 2.0.

### DIFF
--- a/docs/verification.md
+++ b/docs/verification.md
@@ -36,11 +36,11 @@ This library includes a few abstractions to support different use cases, testing
 
 ## Verifier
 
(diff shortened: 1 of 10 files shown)


Hayden-IO (approved):

kommendorkapten (approved):
Thank you, this is a big improvement for the API design.

PR #477, opened 2025-05-09
Deprecate and rename VerifyTimestampAuthority/VerifyArtifactTransparencyLog

codysoyland:
- Rename VerifyTimestampAuthority to VerifySignedTimestamp
- Rename VerifyTimestampAuthorityWithThreshold to VerifySignedTimestampWithThreshold
- Rename VerifyArtifactTransparencyLog to VerifyTlogEntry
- Add aliases for deprecated functions

Similarly to https://github.com/sigstore/sigstore-go/pull/476, this PR is intended to improve some naming prior to sigstore-go 1.0.

### DIFF
--- a/pkg/verify/tlog.go
+++ b/pkg/verify/tlog.go
@@ -29,12 +29,12 @@ import (
 
 const maxAllowedTlogEntries = 32
 
-// VerifyArtifactTransparencyLog verifies that the given entity has been logged
+// VerifyTlogEntry verifies that the given entity has been logged
 // in the transparency log and that the log entry is valid.
 //
 // The threshold parameter is the number of unique transparency log entries
 // that must be verified.
-func VerifyArtifactTransparencyLog(entity SignedEntity, trustedMaterial root.TrustedMaterial, logThreshold int, trustIntegratedTime bool) ([]root.Timestamp, error) { //nolint:revive
+func VerifyTlogEntry(entity SignedEntity, trustedMaterial root.TrustedMaterial, logThreshold int, trustIntegratedTime bool) ([]root.Timestamp, error) { //nolint:revive
 	entries, err := entity.TlogEntries()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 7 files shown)


pkg/verify/tlog.go
```diff
@@ -29,12 +29,12 @@ import (
...
 const maxAllowedTlogEntries = 32
 
-// VerifyArtifactTransparencyLog verifies that the given entity has been logged
+// VerifyTlogEntry verifies that the given entity has been logged
 // in the transparency log and that the log entry is valid.
 //
 // The threshold parameter is the number of unique transparency log entries
 // that must be verified.
-func VerifyArtifactTransparencyLog(entity SignedEntity, trustedMaterial root.TrustedMaterial, logThreshold int, trustIntegratedTime bool) ([]root.Timestamp, error) { //nolint:revive
+func VerifyTlogEntry(entity SignedEntity, trustedMaterial root.TrustedMaterial, logThreshold int, trustIntegratedTime bool) ([]root.Timestamp, error) { //nolint:revive
```
codysoyland, line 37: We can also go with `VerifyRekorEntry` -- I'm open to either.

kommendorkapten: VerifyTlogEntry is what I would prefer.

kommendorkapten (dismissed):
Very good!

kommendorkapten (approved):

Hayden-IO (approved):
For what it's worth, I don't mind dropping the old method names and making breaking changes, we should just push out 1.0 a little longer if we do.

codysoyland:
> For what it's worth, I don't mind dropping the old method names and making breaking changes, we should just push out 1.0 a little longer if we do.

I forgot to reply to this one... I went ahead and pushed out 1.0 containing the deprecated funcs. I don't think it's worth delaying 1.0 over, but I'm also not really committed to keeping the old deprecated names past 1.1 (I think golangci-lint will warn about the deprecations immediately for most folks).

Hayden-IO:
1.1 next week? 😜

PR #481, opened 2025-05-16
Add support for Rekor V2 for signing and verification

cmurphy:
Add support for handling and uploading a Rekor v2 entry. Replace
reliance on rekor v1 with the TransparencyLogEntry from protobuf-specs
as much as possible to reduce divergence in how different entries are
handled. Add new functions where needed to avoid breaking backwards
compatibility with existing functions.

Partial https://github.com/sigstore/rekor-tiles/issues/349


pkg/bundle/bundle.go
```diff
@@ -320,6 +381,30 @@ func (b *Bundle) TlogEntries() ([]*tlog.Entry, error) {
 	return tlogEntries, nil
 }
 
+func (b *BundleForRekorV2) TlogV2Entries() ([]tlog.LogEntry, error) {
```
Hayden-IO: What's the reason for a second TlogEntries function?

Can we reuse the existing and when constructing the bundle, have a validation step that the expected fields are unset?

pkg/verify/tlog.go
```diff
@@ -138,6 +141,78 @@ func VerifyTlogEntry(entity SignedEntity, trustedMaterial root.TrustedMaterial,
...
+
+	verificationContent, err := entity.VerificationContent()
+	if err != nil {
+		return fmt.Errorf("getting verification content from signed entity: %w", err)
+	}
+
+	logEntriesVerified := 0
+
+	for _, entry := range entries {
+		// TODO: validate hashedrekord/dsse obj?
```
Hayden-IO: If we have the artifact, we should verify that the artifact hash matches the hashedrekord hash/dsse payload hash. This was something missing from the current validation function since we can't verify dsse envelope hashes, but this is now resolved since we've dropped the envelope hash.

pkg/tlog/entry.go
```diff
@@ -148,10 +205,18 @@ func ParseEntry(protoEntry *v1.TransparencyLogEntry) (entry *Entry, err error) {
 	if entry.kind != protoEntry.KindVersion.Kind || entry.version != protoEntry.KindVersion.Version {
 		return nil, fmt.Errorf("kind and version mismatch: %s/%s != %s/%s", entry.kind, entry.version, protoEntry.KindVersion.Kind, protoEntry.KindVersion.Version)
 	}
+	entry.tle = protoEntry
 
 	return entry, nil
 }
 
+func ValidateLogEntryBody(entry LogEntry) error {
+	if e, ok := entry.(*rekorV1Entry); ok {
```
Hayden-IO: Ideally we don't have to depend on Rekor v1. What do you think about creating a small Rekor entry verification library that doesn't depend on Rekor v1's generated types, unmarshals canonicalized bodies, and handles validation? We could replace calls to [`UnmarshalEntry`](https://github.com/sigstore/sigstore-go/blob/e7f98e2724435cc9d4798146bb547f46102c45da/pkg/tlog/entry.go#L69).

I'm thinking something like:

```
type RekorEntry interface {
  Kind() string
  Version() string
  Validate() error // validates expected fields are set
  Digest() []byte // return payload hash for DSSE/intoto
  Verifiers() []Verifier // from rekor-tiles
  Signature() []byte
}
```

And then have implementations of the interface for a minimal set of Rekor entries, `HashedRekordV001`, `HashedRekordV002`, `DSSEV001`, `DSSEV002` and `IntotoV002`.

Then have one function like `UnmarshalRekorEntry` that parses a canonicalized body into an `RekorEntry` given its kind and version, and gracefully handles unknown kinds/versions. (Or we can recreate the Rekor v1 types and typeimpls, though I think it's a little overkill and some switch statements would be cleaner)

We probably would need to do some copying from Rekor v1, though maybe there isn't actually too much to copy, I think we just need each entry defined as a struct with json tags to handle unmarshalling the canonicalized body.

Where this lives, here or in sigstore/sigstore?

I think this library would be quite useful across Go clients.

pkg/verify/tlog.go
```diff
@@ -138,6 +141,78 @@ func VerifyTlogEntry(entity SignedEntity, trustedMaterial root.TrustedMaterial,
...
+			return fmt.Errorf("loading note verifier: %w", err)
+		}
+		err = verify.VerifyLogEntry(entry.TransparencyLogEntry(), noteVerifier)
+		if err != nil {
+			return fmt.Errorf("verifying log entry: %w", err)
+		}
+		if !bytes.Equal(entry.Signature(), entitySignature) {
+			return fmt.Errorf("transparency log signature does not match")
+		}
+		protoPubKey := entry.PublicKey()
```
Hayden-IO: IIRC this is confusingly named (to mirror rekor v1), this is a key _or_ certificate. It'd be nice to rename this or change this to return a verifier, and conditionally compare either a certificate or key to what's in the bundle.

examples/sigstore-go-signing/main.go
```diff
@@ -214,22 +216,46 @@ func main() {
...
+				Retries:   1,
+				Version:   version,
+				Origin:    origin,
+				PublicKey: publicKey,
 			}
 			opts.TransparencyLogs = append(opts.TransparencyLogs, sign.NewRekor(rekorOpts))
 		}
 	}
 
+	opts.UseRekorV2 = rekorVersion == uint32(2)
```
Hayden-IO: We also need to enforce that if Rekor v2 is used to sign, there is at least one timestamp authority.

pkg/bundle/bundle.go
cmurphy: The original one uses the rekor v1 generated models to parse and marshal the entry object. It's completely incompatible with the rekor v2 protobuf types.

pkg/verify/tlog.go
cmurphy: This cannot be changed, it's a public method.

pkg/tlog/entry.go
cmurphy: To the extent possible I've changed this to use the TLE instead of the rekor v1 OpenAPI models so that the divergence is minimized. I do not think a separate library would be appropriate. Sigstore-go is the only thing that would be using it and cosign eventually should be consuming sigstore-go.

cmurphy:
Refactored to reduce the amount of v1/v2 divergence to a point I'm happier with. Still need to address time, DSSE, and the verify example.

examples/sigstore-go-signing/main.go
cmurphy: sigstore-go verification policies are composable, it's up to the caller to decide whether to check for a timestamp or not. The bundle signer policies already enforce that there is a timestamp somewhere, and if the entry comes from rekor v2 there will be no integrated timestamp, so it will fail unless there is a TSA timestamp.

pkg/verify/tlog.go
cmurphy: Maybe I'm missing it but I don't think we have access to the artifact here. We just have the bundle.

cmurphy:
e2e tests waiting on https://github.com/sigstore/scaffolding/pull/1606 and https://github.com/sigstore/sigstore-go/pull/489

pkg/sign/signer.go
```diff
@@ -135,7 +135,12 @@ func Bundle(content Content, keypair Keypair, opts BundleOptions) (*protobundle.
 			}
 		}
 
-		verifierOptions = append(verifierOptions, verify.WithTransparencyLog(len(opts.TransparencyLogs)), verify.WithIntegratedTimestamps(len(opts.TransparencyLogs)))
+		verifierOptions = append(verifierOptions, verify.WithTransparencyLog(len(opts.TransparencyLogs)))
+		// Rekor v2 requires a timestamp authority, it will not provide integrated timestamps.
+		// Verification will fail if a timestamp authority is not provided for Rekor v2.
+		if len(opts.TimestampAuthorities) == 0 {
+			verifierOptions = append(verifierOptions, verify.WithIntegratedTimestamps(len(opts.TransparencyLogs)))
```
Hayden-IO, line 142: Is it possible to use the Rekor API version to return an error here rather than fail verification later? This is functional so if not, that's fine, but this will surface an error that won't make the issue obvious to the caller.

pkg/sign/transparency.go
```diff
@@ -60,23 +74,112 @@ type RekorOptions struct {
...
+	if b.VerificationMaterial.TlogEntries == nil {
+		b.VerificationMaterial.TlogEntries = []*protorekor.TransparencyLogEntry{}
+	}
+
+	b.VerificationMaterial.TlogEntries = append(b.VerificationMaterial.TlogEntries, tlogEntry)
+
+	return nil
+}
+
+func (r *Rekor) getRekorV2TLE(ctx context.Context, pubKeyPEM []byte, b *protobundle.Bundle) (*protorekor.TransparencyLogEntry, error) {
```
Hayden-IO: nit, I'd rather use something like `keyOrCertPEM` in all the places that we pass either a key or certificate. The current name is because Rekor v1 tacked on certificates later on into the key field. I'd say the variable name change would help clarify that it's not just a key.

Can also refactor this in another PR.

pkg/tlog/entry.go
```diff
@@ -88,14 +99,110 @@ func NewEntry(body []byte, integratedTime int64, logIndex int64, logID []byte, s
...
 	if inclusionProof != nil {
-		entry.logEntryAnon.Verification = &models.LogEntryAnonVerification{
-			InclusionProof: inclusionProof,
+		hashes := make([][]byte, len(inclusionProof.Hashes))
+		for i, s := range inclusionProof.Hashes {
+			hashes[i] = []byte(s)
+		}
+		entry.tle.InclusionProof = &v1.InclusionProof{
+			LogIndex: logIndex,
+			RootHash: []byte(*inclusionProof.RootHash),
```
Hayden-IO: The root hash and inclusion proof hashes are hex-encoded ([parsing example](https://github.com/sigstore/rekor/blob/1ef8b666115e6f93f80c74ce2b4cae67a529d1fc/pkg/tle/tle.go#L39-L51)), unless these are decoded elsewhere earlier in the stack, so we'll need to hex-decode for the TLE struct.

pkg/tlog/entry.go
```diff
@@ -88,14 +99,110 @@ func NewEntry(body []byte, integratedTime int64, logIndex int64, logID []byte, s
...
+		CanonicalizedBody: tle.CanonicalizedBody,
+	}
+	if tle.InclusionProof != nil {
+		inclusionProof := v1.InclusionProof{
+			LogIndex:   tle.InclusionProof.LogIndex,
+			TreeSize:   tle.InclusionProof.TreeSize,
+			Hashes:     tle.InclusionProof.Hashes,
+			Checkpoint: tle.InclusionProof.Checkpoint,
+		}
+		rootHash := make([]byte, hex.EncodedLen(len(tle.InclusionProof.RootHash)))
```
Hayden-IO: Same here, we don't need to hex-encode.

pkg/tlog/entry.go
```diff
@@ -148,79 +255,148 @@ func ParseEntry(protoEntry *v1.TransparencyLogEntry) (entry *Entry, err error) {
...
+			}
 		}
-	default:
-		return fmt.Errorf("unsupported entry type: %T", e)
 	}
-
 	return nil
 }
 
+func validateHashedRekordV002Entry(hr *rekortilespb.HashedRekordLogEntryV002) error {
```
Hayden-IO, line 288: For the future, we should make https://github.com/sigstore/rekor-tiles/blob/28f3c79d9855192fde1e896c27c4b7b225c7436f/pkg/types/hashedrekord/hashedrekord.go#L68 public and reuse this

pkg/tlog/entry.go
```diff
@@ -230,48 +406,89 @@ func (entry *Entry) PublicKey() any {
...
 }
 
 func (entry *Entry) LogIndex() int64 {
-	return *entry.logEntryAnon.LogIndex
+	return entry.tle.GetLogIndex()
 }
 
 func (entry *Entry) Body() any {
-	return entry.logEntryAnon.Body
+	return base64.StdEncoding.EncodeToString(entry.tle.CanonicalizedBody)
```
Hayden-IO, line 403: We should revisit this method. It returning `any` makes it a bit hard to use in practice. Should this return a byte array that clients have to parse? Since it's always JSON, should this method unmarshal the body into a struct with `version`, `kind`, and `spec`?

Looks like `Body()` isn't called outside of tests, so I'd be fine with a breaking API change for this method.

pkg/tlog/entry.go
```diff
@@ -230,48 +406,89 @@ func (entry *Entry) PublicKey() any {
...
+		IntegratedTime: &entry.tle.IntegratedTime,
+		LogID:          swag.String(string(entry.tle.GetLogId().KeyId)),
+		LogIndex:       swag.Int64(entry.tle.GetInclusionProof().GetLogIndex()),
+		Body:           base64.StdEncoding.EncodeToString(entry.tle.GetCanonicalizedBody()),
+		Verification: &models.LogEntryAnonVerification{
+			InclusionProof: &models.InclusionProof{
+				Checkpoint: swag.String(entry.tle.GetInclusionProof().GetCheckpoint().GetEnvelope()),
+				Hashes:     hashes,
+				LogIndex:   swag.Int64(entry.tle.GetInclusionProof().GetLogIndex()),
+				RootHash:   swag.String(string(entry.tle.GetInclusionProof().GetRootHash())),
```
Hayden-IO: This will need to be hex-encoded...sigh, these hex encodings are _fun_. Glad to not be working around OpenAPI v2 limitations now (iirc, there was something with byte arrays being unsupported?).

pkg/tlog/entry.go
```diff
@@ -230,48 +406,89 @@ func (entry *Entry) PublicKey() any {
...
-	err = rekorVerify.VerifyCheckpointSignature(&entry.logEntryAnon, verifier)
+	err = rekorVerify.VerifyCheckpointSignature(&logEntry, verifier)
 	if err != nil {
 		return err
 	}
 
 	return nil
 }
 
+func VerifyCheckpointAndInclusion(entry *Entry, verifier signature.Verifier, origin string) error {
```
Hayden-IO, line 456: Can you add a function comment for how this differs?

pkg/verify/tlog.go
```diff
@@ -153,3 +171,15 @@ func getVerifier(publicKey crypto.PublicKey, hashFunc crypto.Hash) (*signature.V
 func VerifyArtifactTransparencyLog(entity SignedEntity, trustedMaterial root.TrustedMaterial, logThreshold int, trustIntegratedTime bool) ([]root.Timestamp, error) { //nolint:revive
 	return VerifyTlogEntry(entity, trustedMaterial, logThreshold, trustIntegratedTime)
 }
+
+var treeIDSuffixRegex = regexp.MustCompile(".* - [0-9]+$")
```
Hayden-IO, line 175: I like this as a differentiator between v1 and v2, good idea!

pkg/verify/tlog.go
```diff
@@ -153,3 +171,15 @@ func getVerifier(publicKey crypto.PublicKey, hashFunc crypto.Hash) (*signature.V
...
+
+func hasSTH(entry *tlog.Entry) bool {
```
Hayden-IO: Can we name this to be about rekor v1? In either v1 or v2, there should be a checkpoint/signed tree head.

Hayden-IO (commented):
Great work on this!

pkg/tlog/entry.go
cmurphy: I was facing problems in the unit tests where different tests expected different TLE objects to have the root hash hex encoded or not. This configuration was the only way I could get it to work. I will try to track down exactly why this was happening, but I think the fact that the unit and e2e tests are passing indicate that it does have to be hex encoded.

pkg/tlog/entry.go
cmurphy: We can't use that here because that validates a `HashedRekordRequestV002`, here we have a `HashedRekordLogEntryV002`.

pkg/tlog/entry.go
cmurphy: > Looks like Body() isn't called outside of tests, so I'd be fine with a breaking API change for this method.

I disagree with this. The definition of an exported method is that it could be called by anything outside the code base. I was very careful not to break any APIs in any of these changes. I'd be fine with adding a new `BodyJSON` method and marking this one as deprecated (in a separate PR), but not with changing this one.

pkg/sign/signer.go
cmurphy: I tried but I could not find a good way to do that, there is no way to pass through the rekor API version without changing the method signature or a type definition. It could be deduced by looking at the KindVersion the tlog entry and checking if it is a hashedrekord v002 or dsse v002 type, but I don't think that's a foolproof indicator (what if we add more types to Rekor v2? what if we add those v002 types to Rekor v1? the type of entry is not technically related to whether a RFC3161 timestamp is required). I actually like the way this worked out because it's clear what the reality is, that if you're using Rekor v1 you need an integrated timestamp and if you're using Rekor v2 you need an external timestamp, and it will be clear in the verification result whether you did it correctly.

pkg/tlog/entry.go
Hayden-IO: That might be a bug in the unit tests parsing the fields of the TLE. I know we also had a bug in rekor-tiles, but that's been fixed (https://github.com/sigstore/rekor-tiles/pull/295). Hex encodings should only be needed when working with the OpenAPI generated objects.

pkg/tlog/entry.go
Hayden-IO: Ah yea. So the opposite, we should move this and create a entry verifier library in rekor-tiles. Linking to https://github.com/sigstore/rekor-tiles/issues/234 to figure this out later.

pkg/tlog/entry.go
Hayden-IO: Given the newness of the library, I don't feel as strongly around breaking changes, but I get what you're saying. We have a number of functions we've already marked as deprecated that we'll clean up in a later version, so let's do the same here, deprecate `Body()` and create a new function. Up to you for what this should return - I would find it most useful to return an unmarshaled struct with kind, version, and spec, or the marshaled JSON (and `entry.logEntryAnon.Body` can either be a JSON byte array, or a base64-encoded string of the JSON body ([code](https://github.com/sigstore/rekor/blob/1ef8b666115e6f93f80c74ce2b4cae67a529d1fc/pkg/tle/tle.go#L56-L66)), so we need to handle that)

cmurphy: Can we do that in a separate PR? It's not really related to rekor v2 IMO. Then we can debate there what the right thing to return is.

kommendorkapten (dismissed):

kommendorkapten:
Good work!

pkg/tlog/entry.go
```diff
@@ -88,14 +99,93 @@ func NewEntry(body []byte, integratedTime int64, logIndex int64, logID []byte, s
 	}
 
 	if inclusionProof != nil {
-		entry.logEntryAnon.Verification = &models.LogEntryAnonVerification{
-			InclusionProof: inclusionProof,
+		hashes := make([][]byte, len(inclusionProof.Hashes))
+		for i, s := range inclusionProof.Hashes {
+			hashes[i] = []byte(s)
```
Hayden-IO: Do we need to `hex.DecodeString` here as well? https://github.com/sigstore/rekor/blob/c0f3b8c639cb52bb2fa760465ad7d49e2ff19570/pkg/api/entries.go#L96

Hayden-IO (commented):
Looks really great!

pkg/tlog/entry.go
Hayden-IO: Can we also see if we have test coverage for verifying proofs from both Rekor v1 and v2?

cmurphy: > Can we also see if we have test coverage for verifying proofs from both Rekor v1 and v2?

This is mostly covered by the e2e tests which exercise both v1 and v2. They didn't catch this bug because this is now a deprecated function. I added a new unit test which tests this path and catches this bug.

Hayden-IO (dismissed):
Great work! Just needs a go mod tidy

Hayden-IO (approved):
🎉

PR #482, opened 2025-05-17
Error Wrapping in New function

lukehinds:
The error handling in `updater.New(&tmpCfg)` is currently returned direct, which is not in uniform with other error handling in the TUF client.

This change wraps some meaningful context care of `fmt.Errorf`


kommendorkapten (approved):
Thank you!

PR #484, opened 2025-05-19
Avoid naked errors from other modules

kommendorkapten:
#### Summary
Avoid returning naked errors from external packages

#### Release Note

NONE

### DIFF
--- a/pkg/root/trusted_root.go
+++ b/pkg/root/trusted_root.go
@@ -415,7 +415,7 @@ func FetchTrustedRoot() (*TrustedRoot, error) {
 func FetchTrustedRootWithOptions(opts *tuf.Options) (*TrustedRoot, error) {
 	client, err := tuf.New(opts)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to create TUF client %w", err)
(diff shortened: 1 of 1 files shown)


pkg/root/trusted_root.go
```diff
@@ -452,12 +474,12 @@ func NewLiveTrustedRoot(opts *tuf.Options) (*LiveTrustedRoot, error) {
 func NewLiveTrustedRootFromTarget(opts *tuf.Options, target string) (*LiveTrustedRoot, error) {
 	client, err := tuf.New(opts)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to create TUF client %w", err)
```
steiza, line 477: Minor: maybe we should also do this on line 418 where we're in a similar situation?

steiza (dismissed):
One minor comment; otherwise LGTM!

pkg/root/trusted_root.go
kommendorkapten: Oh, great catch. I'll update.

steiza (approved):
Thanks!

PR #485, opened 2025-05-21
Added a parameter to the tuf options for live refresh.

kommendorkapten:
#### Summary

The live trusted root was always refreshed every 24 hour, with no option to configure this. This commit introduces a new configuration parameter to set it, and adds logging so that it is clear when the root is refreshed.

#### Release Note

* Added configuration option for refresh period for the live trusted root.

#### Documentation

NONE


steiza (commented):
I understand why we did it, but it's a little weird to have `RefreshPeriod` be in `pkg/tuf/options.go`, but that it only gets used in `pkg/root/trusted_root.go`.

Should the refresh behavior live in `pkg/tuf` instead? Or is this really a `LiveTrustedRoot` option (this seems like it would be a pain to add)? Or are we constrained by the existing API, and this is the best option given the circumstances?

kommendorkapten:
Yeah, that's a good comment Zach. I started to work on a third new live trusted root function to configure this. Now that I have thought of it for a while, I should remove the parameter in the options and add a third method. A bit more painful, but it's the right thing to do.

steiza (approved):
LGTM!

PR #489, opened 2025-06-16
Add end to end tests

cmurphy:
Add integration tests using the new lightweight setup-sigstore-env action from sigstore/scaffolding. This CI action starts instances of both Rekor v1 and Rekor v2, so the tests are in a position to start testing the new Rekor v2 pathways immediately.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,42 @@
+# Copyright 2025 The Sigstore Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


Hayden-IO (approved):
Looks great!

My only comment is around the scaffolding setup - it's not yet well-optimized and will spin up infrastructure from scratch with each run. Having to wait ~10 minutes to get feedback after each commit isn't ideal. For now, leaving this test as non-blocking can signal it's not yet optimized, and we can encourage developers to run tests locally where Docker should handle caching.

kommendorkapten (approved):
Looks great!

PR #490, opened 2025-06-25
fail SigstoreTimestampingAuthority) Verify with nil Root

dmitris:
#### Summary
add a check to SigstoreTimestampingAuthority) Verify and return an error if tsa.Root is present but nil to avoid passing it along in the tsaverification.VerifyOpts.

See issue https://github.com/sigstore/cosign/issues/4261 and
a fix the "downstream" failure https://github.com/sigstore/timestamp-authority/pull/1099.

#### Verification
- cosign built with this change no longer panics
- the unit test for the change in this PR when added to the `main` code causes a [panic](https://gist.github.com/dmitris/84467d676e7412bd01d383ba94e8d022) failure.

#### Release Note
* Bug fixes and fixes of previous known issues
- properly handled  `SigstoreTimestampingAuthority.Root` being `nil` in `SigstoreTimestampingAuthority.Verify`

#### Documentation
N/A


codysoyland (approved):
Thanks!

PR #494, opened 2025-06-30
Add support for operator in SigningConfig

Hayden-IO:
The v0.2 signing config includes an Operator field so that clients will select Services from distinct operators. This PR adds support for the Operator field, where multi-service selection will select distinct instances. Later, we'll add support for the Operator field for instances declared in the trusted root, to verify that instances from the same operator do not all count towards meeting a threshold.

Since the v0.2 signing config is only distributed via the staging TUF repo, I've left the example as-is for now, since any integrators will need to declare their signing config manually for production.

### DIFF
--- a/pkg/root/signing_config.go
+++ b/pkg/root/signing_config.go
@@ -39,6 +39,7 @@ type Service struct {
 	MajorAPIVersion     uint32
 	ValidityPeriodStart time.Time
 	ValidityPeriodEnd   time.Time
+	Operator            string
 }
 
 type ServiceConfiguration struct {
@@ -64,6 +65,7 @@ func NewService(s *prototrustroot.Service) Service {
 		MajorAPIVersion:     s.GetMajorApiVersion(),
 		ValidityPeriodStart: start,
 		ValidityPeriodEnd:   end,
+		Operator:            s.GetOperator(),
 	}
 }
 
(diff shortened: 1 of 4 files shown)


pkg/root/signing_config.go
```diff
@@ -76,13 +78,22 @@ func SelectService(services []Service, supportedAPIVersions []uint32, currentTim
 		return "", fmt.Errorf("no supported API versions")
 	}
 
+	// Order supported versions from highest to lowest
 	sortedVersions := make([]uint32, len(supportedAPIVersions))
 	copy(sortedVersions, supportedAPIVersions)
 	slices.Sort(sortedVersions)
 	slices.Reverse(sortedVersions)
 
+	// Order services from newest to oldest
```
Hayden-IO, line 87: Previously, this would select the first service in the list, rather than the newest. This assumed that signing configs were pre-sorted. In practice (or at least for PGI), they should be, but I've added this explicitly to prevent unexpected selections. These sorts shouldn't be inefficient since the number of services should be limited.

pkg/root/signing_config.go
```diff
@@ -410,9 +437,9 @@ func FetchSigningConfigWithOptions(opts *tuf.Options) (*SigningConfig, error) {
 	return GetSigningConfig(client)
 }
 
-// FetchSigningConfig fetches the public-good Sigstore signing configuration target from TUF.
+// GetSigningConfig fetches the public-good Sigstore signing configuration target from TUF.
 func GetSigningConfig(c *tuf.Client) (*SigningConfig, error) {
-	jsonBytes, err := c.GetTarget("signing_config.json")
+	jsonBytes, err := c.GetTarget("signing_config.v0.2.json")
```
cmurphy: Could this be a breaking change? The comment says it's for PGI but it looks like this could be used with any TUF mirror?

Hayden-IO: While it could be used for any TUF mirror, this shouldn't be a breaking change because we intend for this function to only be used with PGI. It should be invisible to users whether the signing config (or trusted root) is coming from a later revision, which is what this function is designed for.

I reverted this change though because it will break if `GetSigningConfig` is used against the production TUF repo. Tracking updating this in https://github.com/sigstore/sigstore-go/issues/495.

cmurphy (approved):

kommendorkapten (approved):

PR #497, opened 2025-07-01
Allow public keys to sign hashedrekord

cmurphy:
There was a restriction when signing Rekor v1 hashedrekord entries that required them to use a certificate rather than a public key. This isn't a restriction imposed by Rekor, and it is going to need to have this kind of flexibility if and when this gets used for signing in cosign. This change eliminates the restriction and then updates the examples to show how this could be used.

### DIFF
--- a/examples/sigstore-go-signing/main.go
+++ b/examples/sigstore-go-signing/main.go
@@ -15,6 +15,9 @@
 package main
 
 import (
+	"crypto"
+	"crypto/x509"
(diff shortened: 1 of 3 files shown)


cmurphy:
Hoping to be able to use this in rekor-tiles tests so that we don't have to stand up a Fulcio instance.

Hayden-IO (approved):
Good catch! LGTM

PR #498, opened 2025-07-09
Add MarshalJSON to SigningConfig, fix marshaling bug

Hayden-IO:
This adds a MarshalJSON method to SigningConfig so that clients don't have to construct the proto structure themselves. This also fixes a bug in ToServiceProtobuf where we didn't handle the validity window where end was optional. Test cases have been updated as well to specify validity windows and operators, and I've verified that these tests fail without the bug fix.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "go.buildTags": "e2e"
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "go.buildTags": "e2e"
-}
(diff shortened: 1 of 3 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
kommendorkapten: Do we wanna keep this in the repo? I would prefer not to, but instead update the `.gitignore` to automatically ignore "personal configs".

Hayden-IO: My thought was that most developers aren’t configuring per-repo settings, and this would be useful to those using vscode. We’ve done this in other repos that use Go tags as well.

cmurphy: I also prefer keeping personal configs out of the shared repo.

Hayden-IO: No strong preference here - I've deleted the file.

cmurphy (approved):

kommendorkapten (approved):
Thanks!

PR #499, opened 2025-07-10
Select highest API version for SigningConfig services always

Hayden-IO:
Found a bug in the selection logic for SigningConfig services where a lower version service would be selected if it were newer than another service with a higher API version. Fix this issue looping through sorted versions and services until a set of services are found. Also added a test case to catch this.


codysoyland (approved):
lgtm!

PR #503, opened 2025-07-28
Refactor SelectService to return Service rather than URL, add supported API versions

Hayden-IO:
#### Summary

Each signing service implementation sets a global variable for the list
of URLs supported. This avoids clients that use the signing APIs needing
to know which set of service API versions are supported when selecting
services.

This also refactors ServiceService(s) to return the Service struct
itself rather than just a URL, since a signer will need to know the API
version when initializing its client, like for Rekor.

Also removed unnecessary nolints after un-deprecating TrustedRoot's
LogId.

#### Release Note

* Changed return value for root.SelectService and root.SelectServices to return Service struct rather than URL

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@
 /examples/oci-image-verification/oci-image-verification
 /examples/sigstore-go-signing/sigstore-go-signing
 /examples/sigstore-go-verification/sigstore-go-verification
+/.vscode
--- a/examples/sigstore-go-signing/main.go
+++ b/examples/sigstore-go-signing/main.go
@@ -145,12 +145,12 @@ func main() {
 	}
 
 	if *idToken != "" {
-		fulcioURL, err := root.SelectService(signingConfig.FulcioCertificateAuthorityURLs(), []uint32{1}, time.Now())
+		fulcioService, err := root.SelectService(signingConfig.FulcioCertificateAuthorityURLs(), sign.FulcioAPIVersions, time.Now())
 		if err != nil {
 			log.Fatal(err)
 		}
 		fulcioOpts := &sign.FulcioOptions{
-			BaseURL: fulcioURL,
+			BaseURL: fulcioService.URL,
 			Timeout: time.Duration(30 * time.Second),
 			Retries: 1,
 		}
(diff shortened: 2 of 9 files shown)


Hayden-IO:
FYI @spencerschrock You had mentioned awhile ago that the signing example had a lot of setup. This is my attempt at greatly simplifying this, should be more straightforward now.

Hayden-IO:
For more context into this PR, I'm working on a PR to use the sigstore-go sign API in Cosign and wanted to avoid a bunch of boilerplate setup.

steiza (commented):
I agree that things have gotten too complicated, but let's talk through some options here.

`sigstore-go-signing` [used to be much simpler](https://github.com/sigstore/sigstore-go/blob/74cb6b5ab7f797705fa75cd68aece0e79233614b/examples/sigstore-go-signing/main.go). Stepping back, it takes in things like `idToken`, and it needs to put that in `sign.CertificateProviderOptions` before it calls `sign.Bundle()`.

If there isn't an easy way to get the data we need out of `SigningConfig`, can we think about that API more before adding a `NewBundleOptions` into `pkg/sign`?

Definitely open to further discussion here! But something doesn't seem right that the steps needed are:
- Take the `id-token` supplied to `sigstore-go-signing`
- Load / construct a `signingConfig`
- Put the `id-token` into `initOpts`
- Call `NewBundleOptions` which itself calls `root.SelectServices` to create `BundleOptions.CertificateProviderOptions`
- Pass the resulting `BundleOptions` to `sign.Bundle`

... when we could just take the `id-token` and put it into `BundleOptions.CertificateProviderOptions` directly, like we used to?

steiza:
My example using `id-token` isn't fair, because a lot of this complexity is around service URLs, in particular with the transition from Rekor v1 to Rekor v2.

I guess philosophically I'm not sure if `pkg/sign` should support the signing config directly, or if that belongs elsewhere (should the signing config support `sign.BundleOptions` instead? should it live in `cosign`?)

If we do decide it should live in `pkg/sign`, maybe we should call it something like `SigningConfigToBundleOptions` (instead of `NewBundleOptions` - for some reason that name really gets to me!)

Hayden-IO:
> should the signing config support sign.BundleOptions instead? should it live in cosign?

I would like this feature to live somewhere in this codebase to avoid libraries/services taking an unnecessary dependency on the heavyweight Cosign. Scorecard is an example of such - they need to know the set of URLs to sign its output but don't need anything else.

There is still the issue of figuring out where the "batteries included" stuff, namely the OpenID Connect dance and the KMS interactions, should live. For now, `sigstore/sigstore` seems fine. _Maybe_ this feature would move there eventually, but given this is so tightly coupled with signing, I'd prefer it remain in `sigstore-go`.

`SigningConfigToBundleOptions` wouldn't be an accurate name if we want to compose all bundle options, for both the TrustedRoot and SigningConfig. I could omit the bit about composing the TrustedRoot, but I do like one method for constructing the options. I'm open to this if you'd prefer, though there will still be some boilerplate for clients.

Hayden-IO:
I’m thinking a bit more about the ID token example. Either a client provides the token directly, or needs to pull the URL from the Signing config to do the OIDC dance. The latter wouldn’t make sense for Sigstore-go. This would mean that clients may need to parse the signing config still.

lemme think a bit more on this.

Hayden-IO:
@steiza After implementing this in Cosign, I didn't find a need for the method to build the signing options from the signing config.

I removed the last commit and reworded this PR to just focus on exposing the set of supported API versions for each service.

steiza (approved):
LGTM - thanks!

PR #505, opened 2025-07-29
Make conformance compatible with rekor v2

loosebazooka:
Bundles with entries from rekor v2 will not have inclusion promises or integrated time

### DIFF
--- a/test/conformance/main.go
+++ b/test/conformance/main.go
@@ -243,8 +243,12 @@ func main() {
 		}
 
 		// Check bundle and trusted root for Tlog information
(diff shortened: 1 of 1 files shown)


Hayden-IO (approved):
Thanks! To confirm, this was run locally against the latest tests in the conformance suite?

loosebazooka:
Yeah, including the tests in https://github.com/sigstore/sigstore-conformance/pull/230

However I did notice that the verifier in its current state is probably not compatible with a bundle with mixed multiple tlog entries (v1 and v2)

Hayden-IO:
Will continue to follow up offline about the mixed v1-v2 case, which isn't critical to support.

PR #507, opened 2025-08-01
Remove noisy log message

Hayden-IO:
When initializing trusted material in Cosign, on every signing and verification that uses the new bundle format, this message would print. Removing it since it doesn't provide much value.


steiza (approved):
LGTM!

PR #510, opened 2025-08-08
Allow no timestamps to be provided when verifying a key

### LINKED ISSUE #501: Relax signing constraint requiring timestamping for Rekor v2 when a key is used
**Description**

For signing, sigstore-go [currently requires](https://github.com/sigstore/sigstore-go/blob/dcab992b2d5f5c5719258292916e562f9344d0bc/pkg/sign/signer.go#L139-L143) that when a transparency log is specified, a timestamp authority must also be specified, since Rekor v2 requires that a timestamp authority is used to fetch a signed timestamp. Timestamping is only required when Fulcio is used. Since sigstore-go supports providing key material to sign rather than use an ephemeral key+Fulcio cert, we should relax this constraint - if no CA is provided, then it can be assumed that a managed key will be used to sign an artifact and no timestamp authority is required. Relaxing the constraint will not be a breaking change, since we'll just be allowing more configurations.

### LINKED ISSUE #502: Relax verification constraint requiring timestamp when key material is provided
**Description**

During verification, [sigstore-go currently requires](https://github.com/sigstore/sigstore-go/blob/dcab992b2d5f5c5719258292916e562f9344d0bc/pkg/verify/signed_entity.go#L202-L209) a timestamp policy flag. Timestamping is only required when verifying a Fulcio certificate. If verifying with a provided key, then we shouldn't require a timestamping flag.

Hayden-IO:
This adds back an option that allows no timestamps to be provided for verification. This will only work when verifying with a key, and will throw an error if set when trying to verify a certificate.

This is also used when verifying while signing. If a log is used without a timestamp authority, we either verify a certificate with the integrated timestamp or no timestamp for a key. If no log or timestamp is provided, we use current time for a certificate or no timestamp for a key.

Fixes #501
Fixes #502

### DIFF
--- a/pkg/sign/signer.go
+++ b/pkg/sign/signer.go
@@ -108,21 +108,20 @@ func Bundle(content Content, keypair Keypair, opts BundleOptions) (*protobundle.
 		verifierPEM = []byte(pubKeyStr)
 	}
 
-	for _, timestampAuthority := range opts.TimestampAuthorities {
-		timestampBytes, err := timestampAuthority.GetTimestamp(opts.Context, signature)
-		if err != nil {
-			return nil, err
+	if len(opts.TimestampAuthorities) > 0 {
+		for _, timestampAuthority := range opts.TimestampAuthorities {
+			timestampBytes, err := timestampAuthority.GetTimestamp(opts.Context, signature)
+			if err != nil {
+				return nil, err
+			}
+			signedTimestamp := &protocommon.RFC3161SignedTimestamp{
+				SignedTimestamp: timestampBytes,
+			}
+			if bundle.VerificationMaterial.TimestampVerificationData == nil {
+				bundle.VerificationMaterial.TimestampVerificationData = &protobundle.TimestampVerificationData{}
+			}
+			bundle.VerificationMaterial.TimestampVerificationData.Rfc3161Timestamps = append(bundle.VerificationMaterial.TimestampVerificationData.Rfc3161Timestamps, signedTimestamp)
 		}
-
-		signedTimestamp := &protocommon.RFC3161SignedTimestamp{
-			SignedTimestamp: timestampBytes,
-		}
-
-		if bundle.VerificationMaterial.TimestampVerificationData == nil {
-			bundle.VerificationMaterial.TimestampVerificationData = &protobundle.TimestampVerificationData{}
-		}
-
-		bundle.VerificationMaterial.TimestampVerificationData.Rfc3161Timestamps = append(bundle.VerificationMaterial.TimestampVerificationData.Rfc3161Timestamps, signedTimestamp)
 
 		verifierOptions = append(verifierOptions, verify.WithSignedTimestamps(len(opts.TimestampAuthorities)))
 	}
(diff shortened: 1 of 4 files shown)


pkg/verify/signed_entity.go
```diff
@@ -199,10 +202,22 @@ func WithCurrentTime() VerifierOption {
...
+}
+
 func (c *VerifierConfig) Validate() error {
-	if !c.requireObserverTimestamps && !c.requireSignedTimestamps && !c.requireIntegratedTimestamps && !c.useCurrentTime {
+	if !c.requireObserverTimestamps && !c.requireSignedTimestamps && !c.requireIntegratedTimestamps && !c.useCurrentTime && !c.allowNoTimestamp {
 		return errors.New("when initializing a new Verifier, you must specify at least one of " +
-			"WithObserverTimestamps(), WithSignedTimestamps(), or WithIntegratedTimestamps()")
+			"WithObserverTimestamps(), WithSignedTimestamps(), or WithIntegratedTimestamps(). You may also specify " +
+			"WithCurrentTime() or WithNoObserverTimestamps()")
 	}
```
steiza, line 224: Should we also add a check here to make sure that if you have `c.allowNoTimestamp` you don't have any of the other options specified?

Otherwise what should the behavior be if you do something like `c.requireObserverTimestamps` and `c.allowNoTimestamp`?

steiza (commented):
One minor question, otherwise this looks good!

This is only reinforcing my feeling that trying to infer verification policies (like we do in `pkg/sign/signer.go`) is a mess and that we as an ecosystem need to come up with something better here. But to be clear, it's the best option we have today!

pkg/verify/signed_entity.go
Hayden-IO: Good idea, added this!

pkg/sign/signer.go
```diff
@@ -108,22 +108,21 @@ func Bundle(content Content, keypair Keypair, opts BundleOptions) (*protobundle.
...
-			return nil, err
-		}
-
-		signedTimestamp := &protocommon.RFC3161SignedTimestamp{
-			SignedTimestamp: timestampBytes,
-		}
-
-		if bundle.VerificationMaterial.TimestampVerificationData == nil {
-			bundle.VerificationMaterial.TimestampVerificationData = &protobundle.TimestampVerificationData{}
+	if len(opts.TimestampAuthorities) > 0 {
```
kommendorkapten, line 111: Nit: this if statement does not change the existing code as the for loop iterates over the slice for which the if statement checks the length. I feel like this is slightly more confusing than just the for loop as I would expect some piece of code that's not covered by the for loop not iterating so the if statement is redundant.

kommendorkapten (approved):
Logic is sound, just one comment on style. Thanks!

steiza (approved):
I agree with @kommendorkapten's nit, but otherwise LGTM!

pkg/sign/signer.go
Hayden-IO: This change is because of line 126, otherwise `verify.WithSignedTimestamps(len(opts.TSA))` is appended for each iteration of the loop. Alternatively I can wrap this line in a conditional, but that doesn't seem much cleaner.

PR #520, opened 2025-09-03
Support other key algorithms for Rekor v2

Hayden-IO:
We hardcoded ECDSA-P256-SHA256 as the only supported algorithm. This uses the algorithm registry to load the correct signing algorithm to specify its type and digest in the request to Rekor v2. This also fixes an incompatibility with Ed25519 and hashedrekord with Rekor v2, which requires Ed25519ph where the digest is provided during verification.

To test this, I've added support for other signing algorithms in EphemeralKeypair, which will also make the struct useable with Cosign when a signing algorithm is provided.

### DIFF
--- a/pkg/sign/certificate.go
+++ b/pkg/sign/certificate.go
@@ -28,6 +28,7 @@ import (
 	"strings"
 	"time"
 
+	protocommon "github.com/sigstore/protobuf-specs/gen/pb-go/common/v1"
 	"github.com/sigstore/sigstore-go/pkg/util"
 	"github.com/sigstore/sigstore/pkg/oauthflow"
 )
@@ -129,6 +130,11 @@ func (f *Fulcio) GetCertificate(ctx context.Context, keypair Keypair, opts *Cert
 	subject, err := oauthflow.SubjectFromUnverifiedToken(jwtString)
 	if err != nil {
 		return nil, err
+	}
+
+	// Fulcio doesn't support verifying Ed25519ph signatures currently.
+	if keypair.GetSigningAlgorithm() == protocommon.PublicKeyDetails_PKIX_ED25519_PH {
+		return nil, fmt.Errorf("ed25519ph unsupported by Fulcio")
 	}
 
 	// Sign JWT subject for proof of possession
--- a/pkg/sign/keys.go
+++ b/pkg/sign/keys.go
@@ -18,50 +18,89 @@ import (
 	"context"
 	"crypto"
 	"crypto/ecdsa"
-	"crypto/elliptic"
+	"crypto/ed25519"
 	"crypto/rand"
+	"crypto/rsa"
 	"crypto/sha256"
 	_ "crypto/sha512" // if user chooses SHA2-384 or SHA2-512 for hash
 	"crypto/x509"
 	"encoding/base64"
-	"errors"
+	"fmt"
 
 	protocommon "github.com/sigstore/protobuf-specs/gen/pb-go/common/v1"
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
+	"github.com/sigstore/sigstore/pkg/signature"
 )
 
 type Keypair interface {
 	GetHashAlgorithm() protocommon.HashAlgorithm
+	GetSigningAlgorithm() protocommon.PublicKeyDetails
 	GetHint() []byte
 	GetKeyAlgorithm() string
+	GetPublicKey() crypto.PublicKey
 	GetPublicKeyPem() (string, error)
 	SignData(ctx context.Context, data []byte) ([]byte, []byte, error)
 }
 
 type EphemeralKeypairOptions struct {
-	// Optional hint of for signing key
+	// Optional fingerprint for public key
 	Hint []byte
-	// TODO: support additional key algorithms
+	// Optional algorithm for generating signing key
+	Algorithm protocommon.PublicKeyDetails
 }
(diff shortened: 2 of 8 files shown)


Hayden-IO:
In the first iteration of this change, I had created a test struct implementing support for other signing algorithms. It seemed useful to just add support to EphemeralKeypair, but if there's any concerns, I can change this.

pkg/sign/keys.go
```diff
@@ -70,63 +108,71 @@ func NewEphemeralKeypair(opts *EphemeralKeypairOptions) (*EphemeralKeypair, erro
...
-
+// SignData returns the signature and the data to sign, which is a digest except when
+// signing with Ed25519.
 func (e *EphemeralKeypair) SignData(_ context.Context, data []byte) ([]byte, []byte, error) {
-	hashFunc, err := getHashFunc(e.hashAlgorithm)
-	if err != nil {
-		return nil, nil, err
+	hf := e.algDetails.GetHashType()
+	var dataToSign []byte
+	if hf == crypto.Hash(0) {
```
kommendorkapten: An alternative is to create a hash function for Ed25519 here that just returns a copy of the provided data during hashing (would let us get rid of the if statement). 

The downside with that is that we would need to create a copy which can hurt if the payload is big. Maybe add a comment here that we chose this over a hash function that just returns the data for performance reasons? Or are there any other reason you think about structuring the code like this?

pkg/sign/keys_test.go
```diff
@@ -15,43 +15,141 @@
...
+		}
 
-	defaultEphemeralKeypair, err := NewEphemeralKeypair(nil)
-	assert.Nil(t, err)
-	hint = defaultEphemeralKeypair.GetHint()
-	assert.NotEqual(t, hint, []byte(""))
+		for _, alg := range unsupportedAlgorithms {
+			opts := &EphemeralKeypairOptions{Algorithm: alg}
+			_, err := NewEphemeralKeypair(opts)
+			assert.NotNil(t, err)
```
kommendorkapten: Nit but should use `assert.Error` here.

kommendorkapten:
```suggestion
			assert.Error(t, err)
```

kommendorkapten:
Thanks Hayden! The code looks good, some minor comments only.

steiza (commented):
The approach looks fine to me!

pkg/sign/keys.go
Hayden-IO: Interestingly, the `dataToSign` return value is ignored in most use cases. It's only used when providing the digest for hashedrekord. I'd prefer to add a comment and not create a special hash function, since crypto/ed25519 uses Hash(0) as a signal for whether to sign with pure Ed25519 or a variant.

pkg/sign/certificate.go
```diff
@@ -132,7 +136,10 @@ func (f *Fulcio) GetCertificate(ctx context.Context, keypair Keypair, opts *Cert
 	}
 
 	// Sign JWT subject for proof of possession
-	subjectSignature, _, err := keypair.SignData(ctx, []byte(subject))
+	// Fulcio doesn't support verifying Ed25519ph signatures at the moment,
+	// so if we're signing with the prehash variant, we need to adapt the
+	// signer to sign with Ed25519 instead.
+	subjectSignature, _, err := keypair.SignData(context.WithValue(ctx, adaptEd25519Ctx{}, true), []byte(subject))
```
Hayden-IO: We did something similar in Cosign - https://github.com/sigstore/cosign/blob/eed2a117b8e707d702f43e113bc3488a4e3af457/cmd/cosign/cli/sign/sign.go#L633

kommendorkapten: Just so I am understanding this. If a client uses Ed25519ph when requesting a cert from fulcio, we silently fallback to Ed25519 but **with the same key**?

Hayden-IO: Correct, with the same key. We only fallback to Ed25519 to sign the proof of possession, not the artifact signature (Rekor does properly verify ed25519ph).

The reason Fulcio doesn't support Ed25519ph is because crypto/x509 doesn't support the prehash variant. There's no public OID to represent the key algorithm. When I work on the next major release of Fulcio, I'll see what I can to hack this in.

steiza: I appreciate the work being done here to try to work around a lack of support for ed25519ph in Fulcio, but I wonder if instead we should return an error for now instead.

What you have here works, but if / when Fulcio supports ed25519ph in the future, figuring out how to unwind this and migrate people off is going to be a bit tricky. Is there a particular need for this use-case right now?

Hayden-IO: The use-case is a bit niche, to support hashedrekord entries with ed25519 signatures and with a Fulcio certificate (note that Rekor already supports verifying both ed25519 and -ph signatures).

Note that if you are signing an entry with a key and _not_ requesting a Fulcio certificate, then this isn't needed. This would only be if you use Cosign with `--issue-certificate` or if you specify the signature algorithm (not yet supported, https://github.com/sigstore/cosign/pull/3497).

I'm OK with an error for now.

Hayden-IO: I've dropped support for ed25519ph with Fulcio. I also added e2e testing when signing with a self-managed key.

steiza (approved):
LGTM; nice work!

PR #525, opened 2025-09-15
Set user agent for TUF and Rekor v2 clients

Hayden-IO:
This adds version-specific user agents for TUF and Rekor v2 calls, matching what we have for Fulcio, the TSA, and Rekor v1.

Ref: https://github.com/sigstore/cosign/issues/4406


kommendorkapten (approved):

jku (approved):
lgtm.

I'll mention that `util.ConstructUserAgent()` typically produces user-agents like `"sigstore-go/v1.1.2"` -- here the "v" is quite unusual. It's not an issue for me, just pointing it out

steiza (approved):

PR #527, opened 2025-09-25
Add support for signing config for conformance test suite

Hayden-IO:
The signing config along with a trusted root file may be passed to sign-bundle to test Rekor v2, which is not present in both the staging and production TUF metadata. This adds support for the signing config when it's provided.


kommendorkapten (approved):
Looks good, thank you!

PR #528, opened 2025-09-26
Add note regarding API compatibility when using signing config

Hayden-IO:
As discussed in the Go meeting today, we want developers to be aware of the consequences of using a signing config and always selecting the highest API version, which may lead to verifiers unable to verify.


kommendorkapten (approved):
I wasn't attending the go-meeting, but the comment makes sense. Clarifies the potential issues 👍

steiza (approved):

PR #536, opened 2025-11-07
Bump conformance to latest

Hayden-IO:
To populate https://sigstore.github.io/sigstore-conformance/

Need to run a cron job to have this run from main, since that's where the results will be picked up from.

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -22,6 +22,8 @@ on:
   pull_request:
     branches:
       - main
+  schedule:
+    - cron: '0 11 * * 1,4' # US work hours
 
 permissions:
   contents: read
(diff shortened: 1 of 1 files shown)


Hayden-IO:
FYI @jku

.github/workflows/conformance.yml
```diff
@@ -22,6 +22,8 @@ on:
   pull_request:
     branches:
       - main
+  schedule:
+    - cron: '0 11 * * 1,4' # US work hours
```
kommendorkapten: This would be 3 AM PST right? As the cron time (11) would be taken the server time which is UTC?

Hayden-IO: Good catch, thanks. Bumped it 8 hours.

kommendorkapten (approved):
Thanks

PR #537, opened 2025-11-09
Bump production Sigstore TUF root to latest

Hayden-IO:
See https://github.com/sigstore/root-signing/blob/main/metadata/root.json to compare and verify


kommendorkapten (approved):
Thanks!

PR #557, opened 2026-01-15
Add preferred method for verifying log entry proofs

Hayden-IO:
A recent vuln report in Cosign was due to a misunderstanding that VerifySET did not perform complete validation of the Rekor entry, in particular a comparison between the bundle's artifact digest, public key and signture, and the entry's contents. This is a minor update to the docstrings for the three log entry verification functions to suggest using the more comprehensive function when possible.


kommendorkapten (approved):
Thanks!

Do we have any understanding of how used the other functions are? If they can be used in an unsafe mode I'm thinking if we can remove them. At least declare them deprecated to start with. Thoughts?

Hayden-IO:
I see these as lower level functions but not necessarily something that needs to be deprecated. They're useful building blocks still, you just need to know what you're doing.

PR #558, opened 2026-01-15
docs(verify): annotate identity matchers and regexp semantics

1seal:
Thanks for the pointer.

This PR adds GoDoc annotations to SubjectAlternativeNameMatcher and IssuerMatcher (and their fields) around pkg/verify/certificate_identity.go#L26-L34 to make the default Go regexp semantics explicit:

- Regexp.MatchString is substring-based unless the pattern is anchored.
- Added guidance to use ^...$ if callers intend full-string matching.
- Clarified that the exact-value fields (SubjectAlternativeName / Issuer) are exact matches when set.

If you’d prefer this guidance to live elsewhere (for example, pkg/verify docs or a README section), I can move it accordingly.

### DIFF
--- a/pkg/verify/certificate_identity.go
+++ b/pkg/verify/certificate_identity.go
@@ -23,13 +23,28 @@ import (
 	"github.com/sigstore/sigstore-go/pkg/fulcio/certificate"
 )
 
+// SubjectAlternativeNameMatcher specifies how to match a certificate's Subject
+// Alternative Name (SAN).
 type SubjectAlternativeNameMatcher struct {
-	SubjectAlternativeName string        `json:"subjectAlternativeName"`
-	Regexp                 regexp.Regexp `json:"regexp,omitempty"`
+	// SubjectAlternativeName, if set, must match the certificate SAN exactly.
+	SubjectAlternativeName string `json:"subjectAlternativeName"`
+	// Regexp, if set, is evaluated against the certificate SAN using Go regexp
+	// semantics.
+	//
+	// Note: regexp matching is not anchored by default; use ^...$ if you intend
+	// to match the entire SAN value.
+	Regexp regexp.Regexp `json:"regexp,omitempty"`
 }
 
+// IssuerMatcher specifies how to match a certificate's issuer identity.
 type IssuerMatcher struct {
-	Issuer string        `json:"issuer"`
+	// Issuer, if set, must match the OpenID Connect issuer exactly.
+	Issuer string `json:"issuer"`
+	// Regexp, if set, is evaluated against the OpenID Connect issuer using Go
+	// regexp semantics.
+	//
+	// Note: regexp matching is not anchored by default; use ^...$ if you intend
+	// to match the entire issuer value.
 	Regexp regexp.Regexp `json:"regexp,omitempty"`
 }
 


pkg/verify/certificate_identity.go
```diff
@@ -23,13 +23,40 @@ import (
 	"github.com/sigstore/sigstore-go/pkg/fulcio/certificate"
 )
 
+// SubjectAlternativeNameMatcher specifies how to match a certificate's Subject
+// Alternative Name (SAN).
+//
+// If Regexp is set, it is evaluated using Go regexp semantics. Note that regexp
```
Hayden-IO: Let's omit this line here, as this is repetitive with the docstring for `Regexp`.

pkg/verify/certificate_identity.go
```diff
@@ -23,13 +23,40 @@ import (
...
-	Regexp                 regexp.Regexp `json:"regexp,omitempty"`
+	// SubjectAlternativeName, if set, must match the certificate SAN exactly.
+	SubjectAlternativeName string `json:"subjectAlternativeName"`
+	// Regexp, if set, is evaluated against the certificate SAN using Go regexp
+	// semantics.
+	//
+	// Note: regexp matching is not anchored by default; use ^...$ if you intend
+	// to match the entire SAN value.
+	//
+	// This field is encoded to JSON as a string via MarshalJSON.
```
Hayden-IO: This can be omitted.

pkg/verify/certificate_identity.go
```diff
@@ -23,13 +23,40 @@ import (
...
+	Regexp regexp.Regexp `json:"regexp,omitempty"`
 }
 
+// IssuerMatcher specifies how to match a certificate's issuer identity.
+//
+// If Regexp is set, it is evaluated using Go regexp semantics. Note that regexp
```
Hayden-IO: same here, this can be omitted

Hayden-IO (changes requested):
Can you also sign the DCO?

pkg/verify/certificate_identity.go
1seal: done, removed the redundant lines from the struct docstrings and dropped the MarshalJSON note. also reworked the commit to include a DCO sign-off. force-pushed the updated commit

pkg/verify/certificate_identity.go
```diff
@@ -23,13 +23,30 @@ import (
...
+	//
+	// Note: regexp matching is not anchored by default; use ^...$ if you intend
+	// to match the entire SAN value.
+	Regexp regexp.Regexp `json:"regexp,omitempty"`
 }
 
+// IssuerMatcher specifies how to match a certificate's issuer identity.
 type IssuerMatcher struct {
-	Issuer string        `json:"issuer"`
+	// Issuer, if set, must match the certificate issuer exactly.
```
Hayden-IO:
```suggestion
	// Issuer, if set, must match the OpenID Connect issuer exactly.
```

pkg/verify/certificate_identity.go
```diff
@@ -23,13 +23,30 @@ import (
...
+	Issuer string `json:"issuer"`
+	// Regexp, if set, is evaluated against the certificate issuer using Go
```
Hayden-IO:
```suggestion
	// Regexp, if set, is evaluated against the OpenID Connect issuer using Go
```

pkg/verify/certificate_identity.go
```diff
@@ -23,13 +23,30 @@ import (
...
+	// regexp semantics.
+	//
+	// Note: regexp matching is not anchored by default; use ^...$ if you intend
+	// to match the entire issuer value.
+	//
+	// This field is encoded to JSON as a string via MarshalJSON.
```
Hayden-IO: This can be deleted as well

Hayden-IO (changes requested):

pkg/verify/certificate_identity.go
1seal: done, updated Issuer/Regexp wording to “OpenID Connect issuer” and removed the MarshalJSON note as suggested. tests pass and I force-pushed the updated commit (DCO > sign-off included)

Hayden-IO (approved):
Thanks!

PR #562, opened 2026-01-17
fix(tlog): fail closed for rekor v2 parsing

1seal:
summary:
- tighten rekor v2 parsing to require `apiVersion: "0.0.2"` and a supported v2 `spec` oneof; otherwise fall back to rekor v1 parsing
- make `ValidateEntry()` fail closed when a rekor v2 entry has an unset/unsupported `spec` oneof
- add tests covering v1 fallback and v2 empty-spec rejection

### DIFF
--- a/pkg/tlog/entry.go
+++ b/pkg/tlog/entry.go
@@ -280,6 +280,8 @@ func ValidateEntry(entry *Entry) error {
 			if err != nil {
 				return err
 			}
+		default:
+			return fmt.Errorf("unsupported rekor v2 entry type: %T", e)
 		}
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


pkg/tlog/entry.go
```diff
@@ -525,9 +523,16 @@ func unmarshalRekorV1Entry(body []byte) (types.EntryImpl, error) {
 
 func unmarshalRekorV2Entry(body []byte) (*rekortilespb.Entry, error) {
 	logEntryBody := rekortilespb.Entry{}
-	err := protojson.Unmarshal(body, &logEntryBody)
-	if err != nil {
+	if err := protojson.Unmarshal(body, &logEntryBody); err != nil {
+		return nil, ErrInvalidRekorV2Entry
+	}
+	if logEntryBody.GetApiVersion() != "0.0.2" {
```
kommendorkapten: I think we should switch the logic around here, first get the entry type, then check the api version based on that as the entry type is the factor that decides which api versions that are allowed.

1seal:
agreed — i updated unmarshalRekorV2Entry to first determine the v2 spec type (oneof) and then validate apiVersion against the allowed version(s) for that type, since entry type is what drives the version constraint.

added a regression test for a valid v2 spec with a mismatched apiVersion (now rejected). pushed in d57cc40.

kommendorkapten:
Thank you! The change looks good, can you just ammend the commit with signoff?

1seal:
done — amended with dco signoff and force-pushed. new commit: 8bd066d.

kommendorkapten (dismissed):
Thank you, looks good.

Will wait for @Hayden-IO to give a look as well, as this is code he is more familiar with than I am.

Hayden-IO (dismissed):
Thanks, this LGTM! Can you rebase?

1seal:
thanks, updated. the cap is now enforced in (*bundle.Bundle).TlogEntries() (before make(...)/parsing), with the same too many verificationMaterial.tlogEntries: got=%d max=%d error format

1seal:
done — rebased onto latest origin/main and force-pushed. new commits: 7036375, a45b041.

Hayden-IO (approved):
Thanks!

PR #566, opened 2026-01-20
ensure we have verification material before attempting to extract public key

bobcallaway:
we weren't checking to see if `pem.Decode()` succeeded or whether we had extracted a verifier in case of parsing a rekor v2 bundle. this explicitly checks for this scenarios and adds a unit test case for it.

thanks to @1seal for catching this


Hayden-IO (approved):

1seal:
thanks for landing the fix. i pulled origin/main at 732d3cfa41e7 (merge of #566) and reran my harness unpatched. TestPEMNilDerefFixed passes and tlog.Entry.PublicKey() is now panic-free on malformed input and fails closed (returns nil) - i see [CALLSITE_HIT] + [NC_MARKER] only. happy to paste pr566.log + sha256 here if useful

PR #567, opened 2026-01-20
fix(bundle): cap tlogEntries during JSON parse

1seal:
summary:
- add a hard upper bound on `verificationMaterial.tlogEntries` in `(*bundle.Bundle).TlogEntries()` before allocating and parsing entries
- keep existing behavior for bundles within the cap (default max=100 as a reasonable upper bound); callers should still enforce overall bundle byte-size limits at ingestion
- add tests covering the cap behavior

### DIFF
--- a/pkg/bundle/bundle.go
+++ b/pkg/bundle/bundle.go
@@ -15,10 +15,8 @@
 package bundle
 
 import (
-	"bytes"
 	"crypto/x509"
 	"encoding/base64"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
@@ -48,7 +46,7 @@ var ErrDecodingB64 = fmt.Errorf("%w: decoding base64", ErrInvalidAttestation)
 
 const mediaTypeBase = "application/vnd.dev.sigstore.bundle"
 
-// defense-in-depth: bound attacker-controlled counts during parse.
+// defense-in-depth: bound the number of tlog entries processed from a bundle.
 // note: callers should still enforce overall bundle byte-size limits at ingestion.
 const maxAllowedTlogEntries = 100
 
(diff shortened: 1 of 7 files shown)


1seal:
note: this cap is enforced before protojson.Unmarshal; capping in Bundle.TlogEntries() is too late to affect parse-time allocation.

pkg/bundle/bundle.go
Hayden-IO, line 318: Sorry if I was unclear in the email, but this should be enforced here, not at unmarshaling time.

Hayden-IO (changes requested):

1seal:
update: per your request, the cap is now enforced in Bundle.TlogEntries() (before allocating/parsing entries) rather than during JSON unmarshaling.

pkg/bundle/bundle.go
```diff
@@ -301,6 +305,14 @@ func (b *Bundle) TlogEntries() ([]*tlog.Entry, error) {
 		return nil, nil
 	}
 
+	if len(b.VerificationMaterial.TlogEntries) > maxAllowedTlogEntries {
+		return nil, ErrValidationError(fmt.Errorf(
+			"too many verificationMaterial.tlogEntries: got=%d max=%d",
```
Hayden-IO:
```suggestion
			"too many transparency log entries in the bundle: got=%d max=%d",
```

pkg/bundle/bundle_test.go
```diff
@@ -31,6 +33,60 @@ import (
...
+		if i > 0 {
+			b.WriteByte(',')
+		}
+		b.WriteString(`{}`)
+	}
+	b.WriteString(`]}}`)
+	return b.Bytes()
+}
+
+func TestBundleUnmarshalJSON_TlogEntriesOverLimitFailsClosed(t *testing.T) {
```
Hayden-IO: Can you structure these tests as a table - https://go.dev/wiki/TableDrivenTests?

1seal:
done — updated the error string to "too many transparency log entries in the bundle: got=%d max=%d" and converted the cap tests to table-driven style. also updated the assertions to match the new error text.

Hayden-IO (dismissed):

kommendorkapten:
As the (proto) JSON parse is already happening for the bundle itself, this new check only enforces limits for parsing of the canonicalized body, right. So a big bundle could still create problems? That's fine as clients may first check the size of the bundle before accepting. 

However, we already have checks in place for the number of allowed tlog entries and signed timestamps:
https://github.com/sigstore/sigstore-go/blob/37d30e1b472bc742e2bbf17067f173bfc7f70ab5/pkg/verify/tlog.go#L32
https://github.com/sigstore/sigstore-go/blob/37d30e1b472bc742e2bbf17067f173bfc7f70ab5/pkg/verify/tsa.go#L24

Instead of setting a hard coded value in bundle.go TLogEntries, how about update the `SignedEntity` interface: https://github.com/sigstore/sigstore-go/blob/37d30e1b472bc742e2bbf17067f173bfc7f70ab5/pkg/verify/interface.go#L58

to accept the max parameter and return error if too many exists? This would be the same change for Timestamps too of course. This would be a breaking change though, as the api is public. I'm thinkin that we can adda new set of functions `LimitedTlogEntries(int max)` or something to not change the behavior of the existing ones. And possibly deprecate them?

What I don't fancy now is that we have two different mitigations in place, with different knobs for tuning. One to minimize overhead during parsing, and one to minimize overhead during verification. Combining them would be great.

1seal:
thanks, this is a great point and i agree with the concern.

if the cap is only enforced in bundle.go (e.g., in Bundle.TlogEntries), then protojson.Unmarshal has already materialized verificationMaterial.tlogEntries into an attacker-sized in-memory structure. in that shape, the check mainly limits the downstream per-entry parsing/canonicalization work, but it does not mitigate the parse-time allocation spike that motivated this change in the first place.

i agree that callers should enforce an overall bundle byte-size limit at ingestion, but for this specific issue the goal is to fail fast on attacker-controlled tlogEntries count *before* protojson.Unmarshal expands the array.

on the “two knobs” concern: +1, it would be nicer to have a single source of truth for these limits (tlog entries + timestamps) shared between parsing-time and verification-time mitigations.

to avoid a breaking change to SignedEntity, one possible path is:
- centralize the limits as shared exported constants (or a small shared package) and have both the parse-time guard and the verify-time checks reference the same values.
- optionally add new non-breaking APIs like LimitedTlogEntries(max int) / LimitedTimestamps(max int), keep existing methods as-is, and migrate verify to the limited variants over time.

happy to adjust the approach based on what you prefer (shared constants vs limited methods), but i’d recommend keeping the parse-time guard so large/invalid bundles don’t pay the allocation cost.

kommendorkapten:
> centralize the limits as shared exported constants (or a small shared package) and have both the parse-time guard and the verify-time checks reference the same values.

This is what I think would be the best option now to not cause any breaking changes.

1seal:
i centralized the caps into a shared exported package (pkg/limits) so both the bundle parse-time guard and verify-time checks reference the same values.

notable change: the bundle tlogEntries cap is now 32 (aligned with existing VerifyTlogEntry behavior). bundles with 33+ entries already cannot verify, so parsing now fails early instead of allowing 33–100 through.

also added the analogous shared constant for timestamps (MaxAllowedTimestamps = 32) and wired VerifySignedTimestamp to it.

note: the bundle guard runs after protojson.Unmarshal, so it does not cap allocations during JSON decoding; callers should still enforce bundle size limits at ingestion.

pkg/verify/tsa.go
```diff
@@ -18,11 +18,10 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/sigstore/sigstore-go/pkg/limits"
 	"github.com/sigstore/sigstore-go/pkg/root"
 )
 
-const maxAllowedTimestamps = 32
```
Hayden-IO, line 24: I'd keep this within the package rather than export a constant. It's not configurable either way and it keeps the exported API cleaner.

pkg/bundle/bundle.go
```diff
@@ -301,6 +302,17 @@ func (b *Bundle) TlogEntries() ([]*tlog.Entry, error) {
 		return nil, nil
 	}
 
+	// Note: this bound is enforced after protojson.Unmarshal, so it does not cap
+	// allocations during JSON decoding; callers should still enforce bundle size
```
Hayden-IO, line 306: Let's not be prescriptive here, we aren't suggesting any limits on bundle size. I'd remove the comment.

Hayden-IO (changes requested):

Hayden-IO:
Going to close this out, as verification already has a limit in place and providing an abnormally large bundle is possible in a number of ways.

PR #581, opened 2026-02-01
Support Verification in sigstore/cosign with X.509 Certificate Chain

yangkenneth:
#### Summary

- Within sigstore/cosign we had opened a [thread](https://github.com/sigstore/cosign/pull/4614#issuecomment-3708746997) with @Hayden-IO around supporting OCI Image validation within the v0.3 bundle using certificate chains. 

> So tl;dr - it is fine to include the certificate chain in a v0.3 bundle. If there's any constraints in sigstore-go around parsing v0.3 bundles with chains, we can relax those constraints. Implementing support in Cosign is the intended design - see https://github.com/sigstore/sigstore-go/issues/132 and https://github.com/sigstore/sigstore-go/issues/298 (Edit: also https://github.com/sigstore/sigstore-go/pull/253 for more discussion) for some context, where we discussed an interface that allows us to create verifiers for alternative (aka not Sigstore-spec compliant) signing/verification paths for private deployments.

- For the change in https://github.com/sigstore/cosign/pull/4614 if we attempt to create a v0.3 bundle which includes a certificate chain and verify the OCI Image we get the following error. By adding a configurable `--allow-certificate-chain` flag we load the Intermediate Certificate Authorities that are exist within the Sigstore bundle to establish the Trust Store.

```sh
cosign verify \
    --insecure-ignore-tlog \
    --insecure-ignore-sct \
    --check-claims=true \
    --certificate-identity example.com \
    --certificate-oidc-issuer-regexp ".*" \
    --trusted-root trusted-roots.json \
    --new-bundle-format \
    --use-signed-timestamps \
    --experimental-oci11 \
    {OCI IMAGE}
WARNING: Skipping tlog verification is an insecure practice that lacks transparency and auditability verification for the signature.
Error: no signatures found
error during command execution: no signatures found
```

```sh
cosign verify \
    --insecure-ignore-tlog \
    --insecure-ignore-sct \
    --check-claims=true \
    --certificate-identity example.com \
    --certificate-oidc-issuer-regexp ".*" \
    --trusted-root trusted-roots.json \
    --new-bundle-format \
    --use-signed-timestamps \
    --experimental-oci11 \
    --allow-certificate-chain \
    {OCI IMAGE}

The following checks were performed on each of these signatures:
  - The cosign claims were validated
  - Existence of the claims in the transparency log was verified offline
  - The code-signing certificate was verified using trusted certificate authority certificates
```

#### Release Note

- NONE

#### Documentation

- https://docs.sigstore.dev/cosign/signing/signing_with_containers

### DIFF
--- a/pkg/bundle/bundle.go
+++ b/pkg/bundle/bundle.go
@@ -108,14 +108,11 @@ func (b *Bundle) validate() error {
 		}
 	}
 
-	// if bundle version >= v0.3, require verification material to not be X.509 certificate chain (only single certificate is allowed)
-	if semver.Compare(bundleVersion, "v0.3") >= 0 {
-		certs := b.VerificationMaterial.GetX509CertificateChain()
-
-		if certs != nil {
-			return errors.New("verification material cannot be X.509 certificate chain (for bundle v0.3)")
-		}
-	}
+	// Note: v0.3 bundles support both single certificates and certificate chains.
+	// The original constraint requiring single certificates was relaxed to support
+	// BYO-PKI use cases where intermediate CAs are ephemeral and need to be
+	// included in the bundle rather than pre-configured in the trusted root.
+	// See: https://github.com/sigstore/cosign/pull/XXXX for context.
 
 	// if bundle version is >= v0.4, return error as this version is not supported
 	if semver.Compare(bundleVersion, "v0.4") >= 0 {
--- a/pkg/bundle/bundle.go
+++ b/pkg/bundle/bundle.go
@@ -52,15 +52,20 @@ func ErrValidationError(err error) error {
 
 type Bundle struct {
 	*protobundle.Bundle
-	hasInclusionPromise bool
-	hasInclusionProof   bool
+	hasInclusionPromise   bool
+	hasInclusionProof     bool
+	allowCertificateChain bool
 }
 
-func NewBundle(pbundle *protobundle.Bundle) (*Bundle, error) {
+func NewBundle(pbundle *protobundle.Bundle, opts ...BundleOption) (*Bundle, error) {
 	bundle := &Bundle{
 		Bundle:              pbundle,
 		hasInclusionPromise: false,
 		hasInclusionProof:   false,
+	}
+
+	for _, opt := range opts {
+		opt(bundle)
 	}
 
 	err := bundle.validate()
@@ -105,6 +110,17 @@ func (b *Bundle) validate() error {
 		// if bundle version >= v0.2, require inclusion proof
 		if len(entries) > 0 && !b.hasInclusionProof {
 			return errors.New("inclusion proof missing in bundle (required for bundle v0.2)")
+		}
+	}
+
+	// if bundle version >= v0.3, only allow certificate chain if option is set
+	if semver.Compare(bundleVersion, "v0.3") >= 0 {
+		if !b.allowCertificateChain {
+			certs := b.VerificationMaterial.GetX509CertificateChain()
+
+			if certs != nil {
+				return errors.New("verification material cannot be X.509 certificate chain (for bundle v0.3)")
+			}
 		}
 	}
 
@@ -195,9 +211,13 @@ func validateBundle(b *protobundle.Bundle) error {
 	return nil
 }
 
-func LoadJSONFromPath(path string) (*Bundle, error) {
+func LoadJSONFromPath(path string, opts ...BundleOption) (*Bundle, error) {
 	var bundle Bundle
 	bundle.Bundle = new(protobundle.Bundle)
+
+	for _, opt := range opts {
+		opt(&bundle)
+	}
 
 	contents, err := os.ReadFile(path)
 	if err != nil {
--- a/pkg/bundle/bundle.go
+++ b/pkg/bundle/bundle.go
@@ -108,12 +108,6 @@ func (b *Bundle) validate() error {
 		}
 	}
 
-	// Note: v0.3 bundles support both single certificates and certificate chains.
-	// The original constraint requiring single certificates was relaxed to support
-	// BYO-PKI use cases where intermediate CAs are ephemeral and need to be
-	// included in the bundle rather than pre-configured in the trusted root.
-	// See: https://github.com/sigstore/cosign/pull/XXXX for context.
-
 	// if bundle version is >= v0.4, return error as this version is not supported
 	if semver.Compare(bundleVersion, "v0.4") >= 0 {
 		return fmt.Errorf("%w: bundle version %s is not yet supported", ErrUnsupportedMediaType, bundleVersion)
--- a/pkg/bundle/bundle.go
+++ b/pkg/bundle/bundle.go
@@ -57,7 +57,7 @@ type Bundle struct {
 	allowCertificateChain bool
 }
 
-func NewBundle(pbundle *protobundle.Bundle, opts ...BundleOption) (*Bundle, error) {
+func NewBundle(pbundle *protobundle.Bundle, opts ...Option) (*Bundle, error) {
 	bundle := &Bundle{
 		Bundle:              pbundle,
 		hasInclusionPromise: false,
@@ -211,7 +211,7 @@ func validateBundle(b *protobundle.Bundle) error {
 	return nil
 }
 
-func LoadJSONFromPath(path string, opts ...BundleOption) (*Bundle, error) {
+func LoadJSONFromPath(path string, opts ...Option) (*Bundle, error) {
 	var bundle Bundle
 	bundle.Bundle = new(protobundle.Bundle)
 
--- a/pkg/bundle/bundle.go
+++ b/pkg/bundle/bundle.go
@@ -271,13 +271,15 @@ func (b *Bundle) VerificationContent() (verify.VerificationContent, error) {
 			return nil, ErrValidationError(err)
 		}
 		var intermediates []*x509.Certificate
-		for _, certProto := range certs {
-			intermediate, err := x509.ParseCertificate(certProto.RawBytes)
-			if err != nil {
-				return nil, ErrValidationError(err)
-			}
-			if intermediate.IsCA {
-				intermediates = append(intermediates, intermediate)
+		if b.allowCertificateChain {
+			for _, certificate := range certs[1:] {
+				intermediate, err := x509.ParseCertificate(certificate.RawBytes)
+				if err != nil {
+					return nil, ErrValidationError(err)
+				}
+				if intermediate.IsCA && !verify.IsSelfSigned(intermediate) {
+					intermediates = append(intermediates, intermediate)
+				}
 			}
 		}
 		cert := &Certificate{
--- a/pkg/bundle/bundle.go
+++ b/pkg/bundle/bundle.go
@@ -300,6 +300,29 @@ func (b *Bundle) VerificationContent() (verify.VerificationContent, error) {
 	}
 }
 
+func (b *Bundle) IntermediateContent() []*x509.Certificate {
+	if b.VerificationMaterial == nil {
+		return nil
+	}
+
+	chain := b.VerificationMaterial.GetX509CertificateChain()
+	if chain == nil {
+		return nil
+	}
+
+	var intermediates []*x509.Certificate
+	for _, certProto := range chain.Certificates {
+		cert, err := x509.ParseCertificate(certProto.RawBytes)
+		if err != nil {
+			continue
+		}
+		if cert.IsCA {
+			intermediates = append(intermediates, cert)
+		}
+	}
+	return intermediates
+}
+
 func (b *Bundle) HasInclusionPromise() bool {
 	return b.hasInclusionPromise
 }
--- a/pkg/bundle/bundle.go
+++ b/pkg/bundle/bundle.go
@@ -109,12 +109,6 @@ func (b *Bundle) validate() error {
 		}
 	}
 
-	// Note: v0.3 bundles support both single certificates and certificate chains.
-	// The original constraint requiring single certificates was relaxed to support
-	// BYO-PKI use cases where intermediate CAs are ephemeral and need to be
-	// included in the bundle rather than pre-configured in the trusted root.
-	// See: https://github.com/sigstore/cosign/pull/XXXX for context.
-
 	// if bundle version is >= v0.4, return error as this version is not supported
 	if semver.Compare(bundleVersion, "v0.4") >= 0 {
 		return fmt.Errorf("%w: bundle version %s is not yet supported", ErrUnsupportedMediaType, bundleVersion)
--- a/pkg/bundle/bundle.go
+++ b/pkg/bundle/bundle.go
@@ -270,8 +270,19 @@ func (b *Bundle) VerificationContent() (verify.VerificationContent, error) {
 		if err != nil {
 			return nil, ErrValidationError(err)
 		}
+		var intermediates []*x509.Certificate
+		for _, certProto := range certs {
+			intermediate, err := x509.ParseCertificate(certProto.RawBytes)
+			if err != nil {
+				return nil, ErrValidationError(err)
+			}
+			if intermediate.IsCA {
+				intermediates = append(intermediates, intermediate)
+			}
+		}
 		cert := &Certificate{
-			certificate: parsedCert,
+			certificate:   parsedCert,
+			intermediates: intermediates,
 		}
 		return cert, nil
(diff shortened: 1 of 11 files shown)


kommendorkapten:
@Hayden-IO Just removing the check for the certificate chain here seems a bit too permissive. This could potential have effects for clients using a PGI-like deployment (such as GitHub) which operates under the same semantics. 

However, I still don't think this would work as IIRC chain building ignores what's in the bundle, it only relies on the intermediates from the bundle: 
https://github.com/sigstore/sigstore-go/blob/3f2ee9eda9b2891647354114ff9cba1167bbcb52/pkg/root/certificate_authority.go#L47-L50

Hayden-IO:
+1 to @kommendorkapten, I agree this is too permissive. I'd be open to some sort of a policy flag that allows this check to be bypassed when set explicitly by the calling library, though off the top of my head I'm not sure the best way to cleanly do this.

yangkenneth:
> +1 to @kommendorkapten, I agree this is too permissive. I'd be open to some sort of a policy flag that allows this check to be bypassed when set explicitly by the calling library, though off the top of my head I'm not sure the best way to cleanly do this.

@Hayden-IO and @kommendorkapten pushed an update to support passing verification options through the bundle. If we want this to be opt-in, we’d need a corresponding change in cosign to add a flag through cosign for verification (e.g. `--allow-certificate-chain`). That way the default behavior stays as-is, and users can explicitly enable validation using certificate chains when using the new bundle format.

```sh
cosign verify \
    --insecure-ignore-tlog \
    --insecure-ignore-sct \
    --check-claims=true \
    --certificate-identity example.com \
    --certificate-oidc-issuer-regexp ".*" \
    --trusted-root trusted-roots.json \
    --new-bundle-format \
    --use-signed-timestamps \
    --experimental-oci11 \
    --allow-certificate-chain \
    {OCI IMAGE}
```

kommendorkapten:
Sorry for a late response but having an explicit parameter `--allow-certificate-chain` seems good to me, as that could be implemented in sigstore-go as a `WithBundleIntermediate` option to be passed to the verifier.

Hayden-IO:
LGTM to this approach as well. I've asked @yangkenneth to file an issue on sigstore/sigstore-conformance to start the discussion with other clients about whether or not we want this standardized across clients.

pkg/bundle/bundle.go
Hayden-IO, line 1: Can you add a test as well?

yangkenneth: @Hayden-IO added tests to validation the options as well as creating an issue within https://github.com/sigstore/sigstore-conformance/issues/342. Let me know if you need anything else in the meantime 🙏

woodruffw (changes requested):
I'm not a maintainer of sigstore-go so my opinion isn't binding here, but IMO this would be a problematic addition: it effectively punches a hole in the v2/v3 bundle distinction for a use case that is best solved by using the v2 bundle format directly (since it isn't deprecated or obsolete, and is intended for this kind of X.509 topology). 

(xref https://github.com/sigstore/sigstore-conformance/issues/342#issuecomment-4060166938 for related)

yangkenneth:
> I'm not a maintainer of sigstore-go so my opinion isn't binding here, but IMO this would be a problematic addition: it effectively punches a hole in the v2/v3 bundle distinction for a use case that is best solved by using the v2 bundle format directly (since it isn't deprecated or obsolete, and is intended for this kind of X.509 topology).
> 
> (xref [sigstore/sigstore-conformance#342 (comment)](https://github.com/sigstore/sigstore-conformance/issues/342#issuecomment-4060166938) for related)

Appreciate the insight @woodruffw; the blocker that I originally ran into was because the current release for cosign does not allow bundle versions [under v0.3](https://github.com/sigstore/cosign/blob/v3.0.5/pkg/oci/remote/signatures.go#L88-L90).

This seems to leave us with a few potential paths forward:

1. Modify sigstore/cosign to accept v0.2 bundles; however this raises the question about long-term trajectory. If there are protobuf changes for the [Sigstore bundle](https://github.com/sigstore/protobuf-specs/blob/69e5b5d27559a51c7afbba17496dbb95bb9f3e35/protos/sigstore_bundle.proto) this becomes a breaking change where implementations using certificate chains would not be able to leverage the latest spec.
2. Allow v0.3 bundles to optionally contain certificate chains through an explicit opt-in flag. An comment would need to be made within the [protobuf spec](https://github.com/sigstore/protobuf-specs/blob/69e5b5d27559a51c7afbba17496dbb95bb9f3e35/protos/sigstore_bundle.proto#L88-L92) to allow certificate chains in v0.3.

Would appreciate hearing both your and @Hayden-IO's opinions here on what the best path forward is.

woodruffw:
Yeah, I think option 1 probably makes the most sense.

> Allow v0.3 bundles to optionally contain certificate chains through an explicit opt-in flag. An comment would need to be made within the protobuf spec to allow certificate chains in v0.3.

This would effectively be a significant semantic change, which IMO would mean it would need to be v4 instead of v3. The protobuf-specs themselves also don't proscribe things like opt-in flags, so I think this would be a bit of a mismatch.

Hayden-IO:
I'm going to move the conversation back over to conformance to make sure we can get some of the other clients to chime in as well, we'll hold off on making changes to sigstore-go until we have some consensus.

Hayden-IO:
@yangkenneth Would you be able to make the update to https://github.com/sigstore/cosign/pull/4614 that demonstrates how this will be used? It'd be helpful to review both this PR and Cosign at the same time. You can temporarily `replace` the go.mod package reference for sigstore-go to your fork so it compiles.

As discussed in the [client meeting](https://github.com/sigstore/sigstore-conformance/issues/342#issuecomment-4068019889), we'll allow this with the proposed flag.

Hayden-IO:
@yangkenneth just let me know when this PR and the Cosign PR are ready for review!

yangkenneth:
> @yangkenneth just let me know when this PR and the Cosign PR are ready for review!

@Hayden-IO ready for a review whenever you have capacity; have the associated changes in https://github.com/sigstore/cosign/pull/4614 which should work if you replace the sigstore-go package with this build. Let me know if you have any questions but thanks again for the help here!

pkg/bundle/bundle.go
```diff
@@ -259,8 +270,19 @@ func (b *Bundle) VerificationContent() (verify.VerificationContent, error) {
 		if err != nil {
 			return nil, ErrValidationError(err)
 		}
+		var intermediates []*x509.Certificate
+		for _, certProto := range certs {
```
kommendorkapten: Could we make the cert parsing optional depending on if the bundle allows certificate chain or not?

pkg/bundle/bundle_test.go
```diff
@@ -127,6 +127,7 @@ func TestMinVersion(t *testing.T) {
 		{"blank", "", "", false},
 		{"invalid", "garbage", "v0.1", false},
 	} {
+		tc := tc
```
kommendorkapten: since go1.22 it's safe to use the automatic loop variable as each variable is diferent, so this assignment is not needed right?

pkg/bundle/bundle.go
```diff
@@ -259,8 +270,19 @@ func (b *Bundle) VerificationContent() (verify.VerificationContent, error) {
...
+			intermediate, err := x509.ParseCertificate(certProto.RawBytes)
+			if err != nil {
+				return nil, ErrValidationError(err)
+			}
+			if intermediate.IsCA {
```
kommendorkapten: We should have a check here to not include root certificates (self signed) as that's prohibited by the spec. Possibly even reject the bundle. What do you think @Hayden-IO ?

https://github.com/sigstore/protobuf-specs/blob/4a31a816c74309e66a4c037c7e20f2500a588f8a/protos/sigstore_bundle.proto#L71

pkg/sign/signer.go
```diff
@@ -74,7 +78,33 @@ func Bundle(content Content, keypair Keypair, opts BundleOptions) (*protobundle.
 
 	// Add verification information to bundle
 	var verifierPEM []byte
-	if opts.CertificateProvider != nil {
+	switch {
+	case opts.CertificateChainProvider != nil:
+		certificateChain, err := opts.CertificateChainProvider.GetCertificateChain(opts.Context, keypair, opts.CertificateProviderOptions)
```
kommendorkapten, line 85: Are we sure here that the root cert is not included?

pkg/verify/certificate.go
```diff
@@ -23,8 +23,19 @@ import (
...
 func VerifyLeafCertificate(observerTimestamp time.Time, leafCert *x509.Certificate, trustedMaterial root.TrustedMaterial) ([][]*x509.Certificate, error) { // nolint: revive
+	return verifyLeafCertificate(observerTimestamp, leafCert, trustedMaterial, nil)
+}
+
+func verifyLeafCertificate(observerTimestamp time.Time, leafCert *x509.Certificate, trustedMaterial root.TrustedMaterial, intermediates []*x509.Certificate) ([][]*x509.Certificate, error) {
 	for _, ca := range trustedMaterial.FulcioCertificateAuthorities() {
-		chains, err := ca.Verify(leafCert, observerTimestamp)
+		caToVerify := ca
+		if fca, ok := ca.(*root.FulcioCertificateAuthority); ok && len(intermediates) > 0 {
+			withIntermediates := *fca
```
kommendorkapten, line 32: nit but this assignment feels unnecessary?

pkg/bundle/bundle.go
Hayden-IO: +1 to rejecting the bundle if it includes self-signed certificates.

pkg/sign/signer.go
yangkenneth: The provider returns the full certificate chain, but based on the [protobuf-spec](https://github.com/sigstore/protobuf-specs/blob/4a31a816c74309e66a4c037c7e20f2500a588f8a/protos/sigstore_bundle.proto#L71) and the points you and @Hayden-IO raised where we should not include any self-signed CA within the bundle I've added a [check](https://github.com/sigstore/sigstore-go/pull/581/changes#diff-dd3b54cb8b706d541132455467c1743cb3fc5574673e35a0f87e36a4098ca89eR96-R107) to support this pattern.

pkg/bundle/bundle.go
yangkenneth: Made some changes where during the generation of the bundle we [fail](https://github.com/sigstore/sigstore-go/pull/581/changes#diff-dd3b54cb8b706d541132455467c1743cb3fc5574673e35a0f87e36a4098ca89eR101-R103) if a self-signed certificate exists but during verification we [ignore](https://github.com/sigstore/sigstore-go/pull/581/changes#diff-65c44fe56b78421271d45f008ac025a643305397d5e100029bb64156bf0840d1R280-R282) based on the [protobuf-spec](https://github.com/sigstore/protobuf-specs/blob/4a31a816c74309e66a4c037c7e20f2500a588f8a/protos/sigstore_bundle.proto#L75-L79) to maintain backwards compatibility.

yangkenneth:
@Hayden-IO @kommendorkapten made some updates based on the feedback, seeing if you al have another change to review so we can get this pushed through. Thanks!

kommendorkapten (dismissed):
Looks good, thanks! Waiting for @Hayden-IO to take a look as well.

yangkenneth:
> Looks good, thanks! Waiting for @Hayden-IO to take a look as well.

Hey @Hayden-IO 👋 just bumping this if you have some time to review, thanks!

Hayden-IO:
yep, it's on my plate, sorry for the delay! will look before eow

pkg/bundle/bundle.go
```diff
@@ -259,8 +270,21 @@ func (b *Bundle) VerificationContent() (verify.VerificationContent, error) {
...
 			return nil, ErrValidationError(err)
 		}
+		var intermediates []*x509.Certificate
+		if b.allowCertificateChain {
+			for _, certificate := range certs[1:] {
+				intermediate, err := x509.ParseCertificate(certificate.RawBytes)
+				if err != nil {
+					return nil, ErrValidationError(err)
+				}
+				if intermediate.IsCA && !verify.IsSelfSigned(intermediate) {
```
Hayden-IO: Should this check throw an error rather than silently skip the certificate? Otherwise, an invalid chain might be constructed.

pkg/bundle/options.go
```diff
@@ -0,0 +1,27 @@
+// Copyright 2023 The Sigstore Authors.
```
Hayden-IO:
```suggestion
// Copyright 2026 The Sigstore Authors.
```

pkg/verify/certificate.go
```diff
@@ -32,3 +43,10 @@ func VerifyLeafCertificate(observerTimestamp time.Time, leafCert *x509.Certifica
 
 	return nil, errors.New("leaf certificate verification failed")
 }
+
+func IsSelfSigned(certificate *x509.Certificate) bool {
```
Hayden-IO: Can this function be under [internal](https://github.com/sigstore/sigstore-go/tree/main/internal/) in a new package? Want to avoid exporting any additional functions.

Hayden-IO (commented):
Appreciate the patience in getting this reviewed! Just a few small comments, otherwise LGTM. Can you also rebase?

yangkenneth:
> Appreciate the patience in getting this reviewed! Just a few small comments, otherwise LGTM. Can you also rebase?

Thanks for the review @Hayden-IO! Addressed your comments (https://github.com/sigstore/sigstore-go/pull/581/changes/08626bac75578cd767e592cce6809330db679e53) and tested the changes as well.

pkg/bundle/bundle.go
```diff
@@ -259,8 +271,27 @@ func (b *Bundle) VerificationContent() (verify.VerificationContent, error) {
...
+		var intermediates []*x509.Certificate
+		if b.allowCertificateChain {
+			for _, cert := range certs[1:] {
+				intermediate, err := x509.ParseCertificate(cert.RawBytes)
+				if err != nil {
+					return nil, ErrValidationError(err)
+				}
+				// protobuf-specs does not allow signers to include self-signed certificates,
+				// but allows verifiers to tolerate non-compliant bundles for backwards compatibility.
+				if certificate.IsSelfSigned(intermediate) {
```
Hayden-IO, line 283: Why should this not return an error as well?

yangkenneth: If I recall correctly v0.2 and older bundles can contain self-signed certificates within the protobuf bundle. Only within v0.3 did the requirements tighten and self-signed certificates cannot be included within the bundle during the signing process. 

If we have an enforcement where `allowCertificateChain` can only be used for v0.3+ then throwing an error here should be fine, otherwise I added this for backwards compatibility in case older versions can use this path as well. Let me know if you want me to add an explicit error here since the expectation should be that we use this verification path for v0.3+ bundle versions.

https://github.com/sigstore/protobuf-specs/blob/4a31a816c74309e66a4c037c7e20f2500a588f8a/protos/sigstore_bundle.proto#L75-L79

Hayden-IO: Self-signed would be fine but only if it were at the end of the chain, which is what the comment was referring to iirc. One of the related issues is that the trust root chain really shouldn't be a chain, it should be a bundle of certificates with chain building being the responsibility of the verification library, not the client's responsibility, at which point we could just skip over superfluous certs.

Anywho, let's constrain this just for v0.3+, since that is the current bundle revision and what the current library will only produce. We can revisit this if it comes up that someone wants to use an older version, though I doubt it will.

yangkenneth: Sounds good! Added the enforcement and validation check https://github.com/sigstore/sigstore-go/pull/581/commits/c5fd166ae3c1eede1dc46c9a2800388cae37a248 👍

Hayden-IO (approved):
Thanks for all your work on this! Will get the Cosign change in much faster.

Hayden-IO:
@woodruffw Did you have anything you wanted to add, since you had left a review?

kommendorkapten (approved):
Thank you!

yangkenneth:
@Hayden-IO @kommendorkapten thank you both for the reviews and getting this through! Would it be possible to cut out a new release so I can reference this version within the follow-up `sigstore/cosign` changes?

Hayden-IO:
yep, will cut v1.2.2 momentarily. feel free to use the latest commit hash for now in the pr

Hayden-IO:
https://github.com/sigstore/sigstore-go/releases/tag/v1.2.2

PR #587, opened 2026-02-18
chore: remove large test dependencies by replacing ctfe usage

Hayden-IO:
Replaces the usage of `github.com/google/certificate-transparency-go/trillian/ctfe` in `pkg/verify/sct_test.g` with a local implementation of `GetCTLogID`.

The `ctfe` package was only being used to calculate a Log ID (SHA256 hash of the public key) but was pulling in significant heavy dependencies, including `database/sql` drivers. This results in a cleaner dependency graph.

Gemini-generated, prompted with:
"Can you figure out the largest dependencies in the dep graph and figure out how to cut them?"

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,7 +25,6 @@ require (
 )
 
 require (
-	filippo.io/edwards25519 v1.1.0 // indirect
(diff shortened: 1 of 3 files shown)


kommendorkapten (approved):
Love it ✂️

PR #590, opened 2026-02-18
Set minimum threshold for WithIntegratedTimestamps

Hayden-IO:
Even if this was set to 0, unless WithNoObserverTimestamps is also set, timestamp verification will fail because no timestamps will be found. This change provides earlier feedback on a misconfiguration.

Thanks to @1seal for reporting this improvement.


steiza (approved):
Hopefully folks who were using Rekor v1 integrated timestamps are getting ready for Rekor v2 by making use of a timestamp authority, but regardless this change is consistent with `WithSignedTimestamps()`, `WithObserverTimestamps()`, etc.

Issue #561, opened 2026-01-16
Conformance test failures


cmurphy:
**Description**

New conformance tests were added that don't pass in sigstore-go and we're having to XFAIL:

- `test_verify*managed-key-happy-path]`
- `test_verify*managed-key-and-trusted-root]`
- `test_verify\[PATH-message-digest-mismatch_fail]`

The non-conformance should be fixed and the XFAILs removed.

PR #600, opened 2026-03-16
verify message digest matches artifact hash

piceri:
Working on https://github.com/sigstore/sigstore-go/issues/561

#### Summary
This change is to address a [new, failing conformance test](https://github.com/sigstore/sigstore-conformance/pull/312).
- Change verifyMessageSignature -> verifyMessageSignatureContent
- Fail if message digest is empty or message digest algorithm is invalid
- Check artifact hash against message digest after verifying signature
- Add testing for supported hash algorithms and mismatched digests
- Remove xfail from conformance test workflow

#### Release Note
- Verification checks the message digest against the artifact has in addition to verifying the message signature. 

#### Documentation
N/A

### DIFF
--- a/pkg/verify/signature.go
+++ b/pkg/verify/signature.go
@@ -380,12 +380,13 @@ func isDigestInSlice(digest []byte, digestSlice [][]byte) bool {
 }
 
 func verifyMessageSignatureContent(verifier signature.Verifier, msg MessageSignatureContent, artifact io.Reader) error {
+	// Message digest is an unauthenticated hint and MUST NOT be used to verify the signature
 	if msg.Digest() == nil {
 		return errors.New("message is missing artifact digest")
 	}
 	hashFunc, err := algStringToHashFunc(msg.DigestAlgorithm())
 	if err != nil {
-		return fmt.Errorf("could not verify message digest: %w", err)
+		return fmt.Errorf("could not verify message digest algorithm: %w", err)
 	}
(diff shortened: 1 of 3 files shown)


pkg/verify/signature.go
```diff
@@ -379,12 +379,26 @@ func isDigestInSlice(digest []byte, digestSlice [][]byte) bool {
 	return false
 }
 
-func verifyMessageSignature(verifier signature.Verifier, msg MessageSignatureContent, artifact io.Reader) error {
-	err := verifier.VerifySignature(bytes.NewReader(msg.Signature()), artifact)
+func verifyMessageSignatureContent(verifier signature.Verifier, msg MessageSignatureContent, artifact io.Reader) error {
+	if msg.Digest() == nil {
```
steiza, line 384: Minor: should we add a comment here explaining that the digest is an unauthenticated hint, but we are checking it here as a sanity check?

steiza (commented):
I know this is a draft, but generally LGTM! One minor comment.

kommendorkapten (approved):

steiza (approved):

Hayden-IO:
Just a very late drive-by comment - Long-term, I'd like a new bundle revision that removes the message digest. If this is just a sanity check, then it's unnecessary and so it should be removed.

This isn't the only place where we have duplicated values - there's a good amount around tlog entries as well.

PR #602, opened 2026-03-20
feat(tuf): Add GetTopLevelTargets to Client API

k4leung4:
## Summary
- Expose `GetTopLevelTargets()` on the TUF `Client`, delegating to the underlying go-tuf v2 `updater.GetTopLevelTargets()`
- Allows callers to enumerate available targets and inspect their metadata (e.g. custom fields) without downloading them

## Motivation
This is needed to support the migration of [sigstore/policy-controller](https://github.com/sigstore/policy-controller) from go-tuf v1 to v2 (see https://github.com/sigstore/policy-controller/pull/1945).

The policy-controller's TrustRoot reconciler needs to:
1. Check if a `trusted_root.json` target exists in the TUF repository
2. Fall back to scanning custom metadata on individual targets to discover Fulcio, Rekor, and CTLog entries

Currently policy-controller uses go-tuf v2's `updater.Updater` directly. With this addition, it can use the sigstore-go TUF client instead, which provides a cleaner abstraction with built-in caching, user-agent handling, and cache validity management.

## Test plan
- [x] `go test ./pkg/tuf/...` passes
- [x] New test: `TestGetTopLevelTargets` — verifies multiple targets returned with correct metadata

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/tuf/client.go
+++ b/pkg/tuf/client.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/theupdateframework/go-tuf/v2/metadata"
 	"github.com/theupdateframework/go-tuf/v2/metadata/config"
 	"github.com/theupdateframework/go-tuf/v2/metadata/fetcher"
 	"github.com/theupdateframework/go-tuf/v2/metadata/updater"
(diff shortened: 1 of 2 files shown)


kommendorkapten:
Thanks for you contribution. 

However I would advice against merging this. Listing the targets a TUF repository provided by looking at the top level targets is antithetical for how TUF operates. The top level targets _may_ be empty, and the targets like `trusted_root.json` can be provided via a delegate, so in that scenario this patch would lead to the incorrect behaviour. 

Proposed is have the client call `GetTargetInfo()` which will correctly find the target by looking at delegations if necessary. Based on the response of that the client can decide if it should fallback to look at the individual targets.

k4leung4:
> Thanks for you contribution.
> 
> However I would advice against merging this. Listing the targets a TUF repository provided by looking at the top level targets is antithetical for how TUF operates. The top level targets _may_ be empty, and the targets like `trusted_root.json` can be provided via a delegate, so in that scenario this patch would lead to the incorrect behaviour.
> 
> Proposed is have the client call `GetTargetInfo()` which will correctly find the target by looking at delegations if necessary. Based on the response of that the client can decide if it should fallback to look at the individual targets.

ack, thank you

PR #611, opened 2026-04-10
Verify log entry digest matches artifact/envelope

### LINKED ISSUE #177: Compare the Rekor entry body to the artifact
**Description**

Tracking bug for https://github.com/sigstore/sigstore-go/blob/main/pkg/verify/tlog.go#L174

This is not absolutely necessary because we do already compare against the [signature](https://github.com/sigstore/sigstore-go/blob/main/pkg/verify/tlog.go#L165-L167) which should effectively bind the entry to the artifact.

Hayden-IO:
A log entry is composed of an artifact hash, signature, and verifier (public key or certificate). When verifying an inclusion proof, we currently pass the canonicalized body to the inclusion proof verifier, where it uses the digest of the canonicalized body as the leaf hash. We must then verify the canonicalized body matches the provided verification material.

We can either recompute the canonicalized body using the artifact hash, signature, and verifier, or we can compare each component individually. That's what's done here - currently, we are comparing the verifier and signature. Comparing the artifact hash is not absolutely necessary because the signature should bind the entry to the artifact, but it shouldn't be assumed that signatures are unique due to malleability.

This PR adds the comparison of the artifact hash. For hashedrekord entries, we compare the digest from the bundle - note that this is marked as an unauthenticated hint, but we now compare it to the provided artifact as part of SignedEntity verification. For dsse/intoto entries we verify only the payload hash, because the envelope hash is over uncanonicalized JSON (and the envelope hash was dropped in Rekor v2).

Arguably, the proper fix is to reconstruct the leaf hash from the artifact, signature and verifier, ignoring the canonicalized body entirely. However, we currently pass the canonicalized body to the inclusion proof verifier, so I didn't want to make a larger change to that library as well at the moment. It would also likely mean a refactoring in sigstore-go with method changes for exported functions.

Fixes #177

### DIFF
--- a/pkg/testing/ca/ca.go
+++ b/pkg/testing/ca/ca.go
@@ -43,6 +43,7 @@ import (
 	"github.com/sigstore/rekor/pkg/generated/models"
 	"github.com/sigstore/rekor/pkg/pki"
 	"github.com/sigstore/rekor/pkg/types"
+	dsseEntry "github.com/sigstore/rekor/pkg/types/dsse"
 	"github.com/sigstore/rekor/pkg/types/hashedrekord"
 	"github.com/sigstore/rekor/pkg/types/intoto"
 	"github.com/sigstore/rekor/pkg/types/rekord"
@@ -326,8 +327,7 @@ func (ca *VirtualSigstore) GenerateTlogEntry(leafCert *x509.Certificate, envelop
 	if err != nil {
 		return nil, err
 	}
-
-	rekorBody, err := generateRekorEntry(intoto.KIND, intoto.New().DefaultVersion(), envelopeBytes, leafCertPem, sig, ca.signingAlgorithmDetails)
+	rekorBody, err := generateRekorEntry(dsseEntry.KIND, dsseEntry.New().DefaultVersion(), envelopeBytes, leafCertPem, sig, ca.signingAlgorithmDetails)
 	if err != nil {
 		return nil, err
 	}
@@ -461,7 +461,7 @@ func createEntry(ctx context.Context, kind, apiVersion string, blobBytes, certBy
 		PKIFormat:      string(pki.X509),
 	}
 	switch kind {
-	case rekord.KIND, intoto.KIND:
+	case rekord.KIND, intoto.KIND, dsseEntry.KIND:
 		props.ArtifactBytes = blobBytes
 		props.SignatureBytes = sigBytes
 	case hashedrekord.KIND:
--- a/pkg/tlog/entry.go
+++ b/pkg/tlog/entry.go
@@ -423,6 +423,72 @@ func (entry *Entry) TransparencyLogEntry() *v1.TransparencyLogEntry {
 	return entry.tle
 }
 
+func (entry *Entry) GetHashedRekordDigest() (digest []byte, algorithm string, ok bool) {
+	if entry == nil {
+		return nil, "", false
+	}
+	if entry.rekorV1Entry != nil {
+		if e, ok := entry.rekorV1Entry.(*hashedrekord_v001.V001Entry); ok {
+			if e.HashedRekordObj.Data != nil && e.HashedRekordObj.Data.Hash != nil {
+				hashBytes, err := hex.DecodeString(*e.HashedRekordObj.Data.Hash.Value)
+				if err != nil {
+					return nil, "", false
+				}
+				return hashBytes, *e.HashedRekordObj.Data.Hash.Algorithm, true
+			}
+		}
+	}
+	if entry.rekorV2Entry != nil {
+		spec := entry.rekorV2Entry.GetSpec()
+		if spec != nil {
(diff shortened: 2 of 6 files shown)


Hayden-IO:
I wanted to get this in for a few reasons:

* When I remove DSSE v2, I'm going to need to update verification. I think the diff will be cleaner to add this first and then refactor to support DSSE envelopes + hashedrekord.
* I'd like to start moving away from using the canonicalized body, and again I think the diff will be cleaner and we'll be less likely to make a mistake starting from this and refactoring.
* This has been bothering me for two years :)

pkg/testing/ca/ca.go
```diff
@@ -326,8 +327,7 @@ func (ca *VirtualSigstore) GenerateTlogEntry(leafCert *x509.Certificate, envelop
 	if err != nil {
 		return nil, err
 	}
-
-	rekorBody, err := generateRekorEntry(intoto.KIND, intoto.New().DefaultVersion(), envelopeBytes, leafCertPem, sig, ca.signingAlgorithmDetails)
+	rekorBody, err := generateRekorEntry(dsseEntry.KIND, dsseEntry.New().DefaultVersion(), envelopeBytes, leafCertPem, sig, ca.signingAlgorithmDetails)
```
Hayden-IO, line 330: Updating to reflect that dsses and not intotos are used now.

pkg/verify/tlog.go
```diff
@@ -136,7 +138,53 @@ func VerifyTlogEntry(entity SignedEntity, trustedMaterial root.TrustedMaterial,
 			return nil, errors.New("transparency log certificate does not match")
 		}
 
-		// TODO: if you have access to artifact, check that it matches body subject
+		// Ensure that the digest/payload in the bundle matches the tlog entry
+		switch {
+		case sigContent.MessageSignatureContent() != nil:
+			// This message digest must be compared to the provided artifact
```
Hayden-IO, line 144: Just flagging this line, particularly to @steiza and @piceri. Given the change Eric just made, I think this is safe to do, but it is relying on what we called an unauthenticated hint.

I'd call this a step in the right direction, with the long-term fix being what I suggested in the commit message.

kommendorkapten: For now I think this is the right call.

pkg/tlog/entry.go
```diff
@@ -423,6 +423,60 @@ func (entry *Entry) TransparencyLogEntry() *v1.TransparencyLogEntry {
...
+				hashBytes, err := hex.DecodeString(*e.HashedRekordObj.Data.Hash.Value)
+				if err != nil {
+					return nil, "", false
+				}
+				return hashBytes, *e.HashedRekordObj.Data.Hash.Algorithm, true
+			}
+		}
+	}
+	if entry.rekorV2Entry != nil {
+		if e, ok := entry.rekorV2Entry.GetSpec().GetSpec().(*rekortilespb.Spec_HashedRekordV002); ok {
```
kommendorkapten: Can `rekorV2Entry.GetSpec()` return `nil`? If so this can crash the verifier.

pkg/tlog/entry.go
```diff
@@ -423,6 +423,60 @@ func (entry *Entry) TransparencyLogEntry() *v1.TransparencyLogEntry {
...
+				return e.HashedRekordV002.GetData().GetDigest(), e.HashedRekordV002.GetData().GetAlgorithm().String(), true
+			}
+		}
+	}
+	return nil, "", false
+}
+
+func (entry *Entry) GetDssePayloadHash() (digest []byte, ok bool) {
+	if entry.rekorV2Entry != nil {
+		if e, ok := entry.rekorV2Entry.GetSpec().GetSpec().(*rekortilespb.Spec_DsseV002); ok {
```
kommendorkapten: Same here, nested `GetSpec()` that may return nil?

pkg/tlog/entry.go
kommendorkapten: I see that `GetSpec` works on a nil receiver, so this should be fine. However, I would suggest we add a test to verify this, e.g. like this:

```go
func TestUnmarshalRekorV2Entry(t *testing.T) {
	var body = []byte("{}")

	re, err := unmarshalRekorV2Entry(body)

	assert.Error(t, err)
	assert.Nil(t, re)
}
```

as the unmarshalRekorV2Entry also calls `GetSpec().GetSpec()`, verifying we don't crash on empty input can be good.

kommendorkapten (dismissed):
Thanks, looks good. Only comment is that we could add one more test to verify we don't crash on empty/partial/malformed input.

pkg/tlog/entry.go
Hayden-IO: Added and added some additional guards against malformed entries.

kommendorkapten (approved):
Thanks!

PR #630, opened 2026-05-23
bundle: cap raw TlogEntries length before per-entry parse

tonghuaroot:
Follow-up to #567 — moves the existing `maxAllowedTlogEntries = 32` cap earlier in the parse path so the bound is enforced on the raw decoded slice length before each entry is individually parsed.

Today the constant is consulted in `pkg/verify/tlog.go:VerifyTlogEntry`, which means `(*Bundle).TlogEntries()` (and the `validate()` call that runs during `Bundle.UnmarshalJSON`) iterate and call `tlog.ParseTransparencyLogEntry` for every element of `VerificationMaterial.TlogEntries` before the count check fires. Moving the check to the start of `TlogEntries()` keeps the cap intact and keeps the parse step's work proportional to the configured limit rather than to the bundle's declared entry count.

Also lifts the constant (and the analogous `maxAllowedTimestamps` cap) into a new `pkg/limits` package so `pkg/bundle/bundle.go` and `pkg/verify/{tlog,tsa}.go` reference the same source of truth. Values are unchanged (32 in both cases).

## Summary
- Add `pkg/limits` with exported `MaxAllowedTlogEntries` and `MaxAllowedTimestamps` (both 32, matching today's behavior).
- Add an early count check at the top of `(*Bundle).TlogEntries()` so the cap fires before the per-entry parse loop.
- Update `pkg/verify/tlog.go` and `pkg/verify/tsa.go` to reference the shared constants.
- Add a regression test in `pkg/bundle/bundle_test.go` asserting the early return surfaces the count-based error (entries are intentionally bare so per-entry parse would otherwise error first).

## Test plan
- [x] `go test ./pkg/bundle/... ./pkg/verify/... ./pkg/limits/...` — all existing tests pass; new `TestTlogEntriesCountCap` and `TestTlogEntriesCountAtCap` pass.
- [x] `go vet ./pkg/bundle/... ./pkg/verify/... ./pkg/limits/...` clean.
- [x] No public API change — the `pkg/verify` constants were unexported; the new shared constants are additive.

Notes for reviewers:
- This is intentionally a smaller, more focused change than #567: same direction on centralizing the constants that @kommendorkapten endorsed in that thread, without changing the cap value or adding new methods to `SignedEntity`.
- The bound here runs after `protojson.Unmarshal`, so it does not cap allocations during JSON decoding itself; callers should still enforce a bundle byte-size limit at ingestion (unchanged from current guidance).


internal/limits/limits.go
```diff
@@ -0,0 +1,29 @@
...
+// Package limits centralizes the per-bundle upper bounds that are shared
+// between the bundle parse path and the verify path. Keeping a single
+// source of truth avoids drift between the two layers and lets the parse
+// step short-circuit on counts that the verify step would later reject.
+package limits
+
+// MaxAllowedTlogEntries is the upper bound on the number of transparency
+// log entries a bundle may carry. The value matches the historical cap
+// enforced inside VerifyTlogEntry.
+const MaxAllowedTlogEntries = 32
```
Hayden-IO, line 21: As I mentioned in the other PR, these shouldn't be publicly exported values.

tonghuaroot: Good call — moved the package from `pkg/limits/` to `internal/limits/` in 463da5a. The constants remain uppercase since cross-package access requires it, but the `internal/` location keeps them off the public Go API surface so external importers cannot depend on them. The 4 import paths in `pkg/bundle/{bundle,bundle_test}.go` and `pkg/verify/{tlog,tsa}.go` were updated to point at the new path; `TestTlogEntriesCountCap` and `TestTlogEntriesCountAtCap` pass.

internal/limits/limits.go
```diff
@@ -0,0 +1,29 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package limits centralizes the per-bundle upper bounds that are shared
+// between the bundle parse path and the verify path. Keeping a single
+// source of truth avoids drift between the two layers and lets the parse
+// step short-circuit on counts that the verify step would later reject.
+package limits
```
Hayden-IO, line 17: This package comment doesn't need to describe this being a single source of truth.

internal/limits/limits.go
```diff
@@ -0,0 +1,29 @@
...
+
+// MaxAllowedTlogEntries is the upper bound on the number of transparency
+// log entries a bundle may carry. The value matches the historical cap
+// enforced inside VerifyTlogEntry.
```
Hayden-IO: This doesn't need to mention a "historical cap", same as below, as that isn't a meaningful description.

pkg/bundle/bundle.go
```diff
@@ -301,6 +302,16 @@ func (b *Bundle) TlogEntries() ([]*tlog.Entry, error) {
 		return nil, nil
 	}
 
+	// Bound the number of transparency log entries we will materialize
+	// before iterating and parsing each one. The same cap is later
+	// re-checked in pkg/verify, so enforcing it here keeps the parse
+	// step's work proportional to the configured limit instead of to
+	// the bundle's declared entry count.
```
Hayden-IO: Code is self-describing, not sure this needs a comment

pkg/bundle/bundle_test.go
```diff
@@ -1104,3 +1105,57 @@ func Test_BundleValidation(t *testing.T) {
 		})
 	}
 }
+
+func TestTlogEntriesCountCap(t *testing.T) {
+	// Build a slice of entries that exceeds the shared cap. The individual
```
Hayden-IO: The test code is self-describing, this comment isn't adding anything.

pkg/bundle/bundle_test.go
```diff
@@ -1104,3 +1105,57 @@ func Test_BundleValidation(t *testing.T) {
...
+	got, err := b.TlogEntries()
+	require.Error(t, err)
+	require.Nil(t, got)
+	require.Contains(t, err.Error(), "too many tlog entries")
+	require.Contains(t, err.Error(), fmt.Sprintf("%d", overLimit))
+	require.Contains(t, err.Error(), fmt.Sprintf("%d", limits.MaxAllowedTlogEntries))
+}
+
+func TestTlogEntriesCountAtCap(t *testing.T) {
+	// At exactly the cap the count check must not fire. The slice here
```
Hayden-IO: same here

Hayden-IO (changes requested):
Left some comments on the code comments. These comments are fairly typical for LLM output in that they are either not needed or overly descriptive, so please review that they are meaningful and in line with Go best practices.

tonghuaroot:
Thanks @Hayden-IO. Pushed [34db922](https://github.com/sigstore/sigstore-go/pull/630/commits/34db922) trimming the redundant comments per the five threads:

- `internal/limits/limits.go:19` — collapsed the package doc to one sentence; dropped the "single source of truth" framing.
- `internal/limits/limits.go:23` (and `:29`) — dropped the "matches the historical cap enforced inside Verify*" sentence on both constants.
- `pkg/bundle/bundle.go:309` — removed the five-line block above the count check; the conditional is self-describing.
- `pkg/bundle/bundle_test.go:1110` (TestTlogEntriesCountCap) and `:1139` (TestTlogEntriesCountAtCap) — removed the comment blocks at the top of each test.

`go vet` clean and the two regression tests still pass on the trimmed source.

Hayden-IO (approved):

kommendorkapten (approved):
Thanks!

PR #633, opened 2026-05-27
Prevent multi-log threshold bypasses via single compromised log

Hayden-IO:
A verifier configured with WithTransparencyLog(N>1) or WithSignedCertificateTimestamps(N>1) expected defense-in-depth against the compromise of a single log instance. However, threshold counting treated verified entries per-entry rather than per-log.

As a result, a single compromised transparency log could forge multiple entries with different indices, and a single compromised CT log could verify multiple times (either across multiple certificate chains or via multiple embedded SCTs), fully satisfying the multi-log threshold requirements and defeating the multi-log policy.

This change ensures that a threshold of N can only be satisfied by N independent logs by enforcing that each distinct transparency log and CT log identity counts as at most one witness towards threshold and observer timestamp policies.

### DIFF
--- a/pkg/verify/sct.go
+++ b/pkg/verify/sct.go
@@ -50,12 +50,16 @@ func VerifySignedCertificateTimestamp(chains [][]*x509.Certificate, threshold in
 		return err
 	}
 
-	verified := 0
+	verifiedLogs := make(map[string]bool)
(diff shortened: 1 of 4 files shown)


steiza (approved):
LGTM!

PR #627, opened 2026-05-14
Encode Rekor v2 DSSE envelopes as hashedrekord

codysoyland:
## Summary

Implements the encoding proposed in sigstore/architecture-docs#63 (rekor-v2-spec §6.1.4): on Rekor v2, DSSE envelopes are uploaded as a `hashedrekord` whose digest covers `Hash(PAE(payloadType, payload))`, where `Hash` is derived from the signing algorithm (ECDSA P-256 → SHA-256, P-384 → SHA-384, etc.; ed25519 must use ed25519ph).

On the verification side, the canonicalized Rekor entry body is now reconstructed from the bundle contents rather than relying on the body in the bundle.

Rekor v1 behavior is unchanged.

PR #635, opened 2026-05-28
Verify Rekor v2 inclusion using reconstructed leaf hash

codysoyland:
## Summary
- For Rekor v2 hashedrekord entries, reconstruct the entry hash from bundle
  contents (signature, certificate/key, digest) and verify inclusion against
  it, rather than relying on the CanonicalizedBody in the bundle.
- Non-hashedrekord entries (e.g. DSSE) fall back to the existing
  `VerifyLogEntry` path.
- Bumps `rekor-tiles/v2` to pick up `VerifyLogEntryWithHash` and
  `hashedrekord.ToEntryHash`.

Split from #627 per review feedback.


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/sigstore/sigstore-go
 
-go 1.25.0
+go 1.25.1
```
Hayden-IO: If we can hold off merging for one day, I'm queueing up PRs to drop the dep that forced this upgrade.

pkg/verify/tlog.go
```diff
@@ -205,6 +227,81 @@ func VerifyTlogEntry(entity SignedEntity, trustedMaterial root.TrustedMaterial,
 	return verifiedTimestamps, nil
 }
 
+// reconstructV2EntryHash rebuilds a Rekor v2 hashedrekord entry hash from
+// bundle contents, so inclusion can be verified without trusting the
```
Hayden-IO: nit, I'd say without _relying_ on the canonicalized body, rather than _trusting_. We still did verify the canonicalized body, it was just a more complicated step.

pkg/verify/tlog.go
```diff
@@ -205,6 +227,81 @@ func VerifyTlogEntry(entity SignedEntity, trustedMaterial root.TrustedMaterial,
...
+	)
+	if leafCert := verificationContent.Certificate(); leafCert != nil {
+		pubKey = leafCert.PublicKey
+		verifierProto = &rekortilespb.Verifier{
+			Verifier: &rekortilespb.Verifier_X509Certificate{
+				X509Certificate: &protocommon.X509Certificate{RawBytes: leafCert.Raw},
+			},
+		}
+	} else if pkp := verificationContent.PublicKey(); pkp != nil {
+		v, err := trustedMaterial.PublicKeyVerifier(pkp.Hint())
```
Hayden-IO, line 246: To confirm, `verificationContent.PublicKey()` doesn't specify the public key since this is coming from the bundle, correct? Since we only have the hint, we have to look up the key in the trusted material?

Hayden-IO (commented):
thanks for splitting this up!

go.mod
Hayden-IO: https://github.com/sigstore/rekor-tiles/pull/807

pkg/verify/tlog.go
codysoyland: I agree that's better language 👍🏻

go.mod
codysoyland: I see that's merged. I can bump the pin but do you want to cut a release first?

pkg/verify/tlog.go
codysoyland: That's right, the actual public key is in the TrustedMaterial, not in the bundle.

go.mod
Hayden-IO: You can bump to the commit hash, will cut a release after a few more changes.

Hayden-IO (approved):

PR #645, opened 2026-06-15
Reject certificate identity with no SAN or issuer criteria

### LINKED ISSUE #644: WithCertificateIdentity accepts an empty identity that matches any certificate
**Description**

`WithCertificateIdentity` accepts a `CertificateIdentity` with no SAN and no issuer criteria, and that identity then matches every certificate. `NewCertificateIdentity` rejects the same empty identity, so the two paths disagree.

The cause is that `SubjectAlternativeNameMatcher.Verify` and `IssuerMatcher.Verify` return nil (a match) when both the exact value and the regexp are empty. So `CertificateIdentity{}` matches any SAN and any issuer.

Reproduction, in package `verify`:

```go
func TestEmptyIdentityMatchesAnything(t *testing.T) {
    arbitrary := certificate.Summary{
        CertificateIssuer:      "https://evil.example.com",
        SubjectAlternativeName: "https://github.com/attacker/anything@refs/heads/main",
    }

    _, err := NewCertificateIdentity(SubjectAlternativeNameMatcher{}, IssuerMatcher{}, certificate.Extensions{})
    require.Error(t, err) // constructor rejects an empty identity

    pc := &PolicyConfig{}
    require.NoError(t, WithCertificateIdentity(CertificateIdentity{})(pc)) // but this accepts it

    _, err = CertificateIdentities{CertificateIdentity{}}.Verify(arbitrary)
    require.NoError(t, err) // and it matches an arbitrary certificate
}
```

Correct behavior: an identity with no SAN and no issuer criteria should be rejected, the same way `NewCertificateIdentity` rejects it, instead of matching every certificate. Skipping identity verification is otherwise only reachable through the explicit `WithoutIdentitiesUnsafe`.

This needs a caller to pass an empty `CertificateIdentity{}`, so it is a misconfiguration rather than attacker-controlled input, but it fails open, which is why it seems worth closing. Fix is small: validate in `WithCertificateIdentity` as `NewCertificateIdentity` does, or fail closed when both matcher fields are empty.

**Version**

v1.2.1 (commit `4594ab4`).

kanywst:
#### Summary

WithCertificateIdentity accepted a certificate identity with no SAN or issuer criteria, which then matches any certificate (#644). NewCertificateIdentity already rejects it;
this moves that check into a validate method so both paths reject an empty identity. SAN-only and issuer-only identities are unaffected.

Reviewers can run go test ./pkg/verify/ -run TestWithCertificateIdentityRejectsEmptyIdentity; it fails before this change and passes after.

Closes #644

#### Release Note

Fixed WithCertificateIdentity accepting a certificate identity with no SAN or issuer criteria, which matched any certificate.

#### Documentation

NONE

### DIFF
--- a/pkg/verify/certificate_identity.go
+++ b/pkg/verify/certificate_identity.go
@@ -118,6 +118,9 @@ func (s *SubjectAlternativeNameMatcher) MarshalJSON() ([]byte, error) {
 // Verify checks if the actualCert matches the SANMatcher's Value and
 // Regexp – if those values have been provided.
 func (s SubjectAlternativeNameMatcher) Verify(actualCert certificate.Summary) error {
+	if s.SubjectAlternativeName == "" && s.Regexp.String() == "" {
+		return errors.New("when verifying a certificate identity, there must be subject alternative name criteria")
+	}
 	if s.SubjectAlternativeName != "" &&
 		actualCert.SubjectAlternativeName != s.SubjectAlternativeName {
 		return &ErrValueMismatch{"SAN", string(s.SubjectAlternativeName), string(actualCert.SubjectAlternativeName)}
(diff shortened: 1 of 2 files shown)


steiza (commented):
Nice find!

Since the `Verify()` functions are public, I think we should be checking to ensure the criteria are not empty there.

steiza (approved):
Can I approve changes I pushed? I haven't edited a contributed pull request before.

kommendorkapten (approved):
Thanks!

Hayden-IO (approved):

PR #646, opened 2026-06-15
Update workflows to use Actions network firewall

steiza:
#### Summary

So that we log the outbound network traffic that happens during builds!

We're doing a private preview of the Actions network firewall that logs all outbound network activity and uploads it as an Actions workflow artifact to the job. See for example https://github.com/steiza/sigstore-go/actions/runs/27418947268 (scroll down to "Artifacts").

I added github.com/sigstore to the feature flag so that it can use the private preview. If there are any issues, simply revert the `runs-on:` back to `ubuntu-latest`.

#### Release Note

NONE

#### Documentation

N/A

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,7 +27,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-latest, ubuntu-latest, windows-latest]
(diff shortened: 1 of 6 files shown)


Hayden-IO (approved):
Very cool!

PR #647, opened 2026-06-16
Handle 409s from Rekor by fetching the entry

Hayden-IO:
When a CI pipeline fails midway, a client may try to upload the same
signed entry again. This will result in Rekor returning a 409.
Additionally, occasionally Rekor v1 fails with a 409 - the root cause
isn't known, it's a flake in Trillian.

This change gracefully handles a 409 Conflict by retrieving the
transparency log entry. For Rekor v1, this is simple, just use the UUID
in the response to lookup the entry by UUID.

Rekor v2 is not handled because we aren't able to fully construct the
TLE without the log's public key, which is the TLE's LogID. A comment
has been added with details on how someone could implement it in the
future.

### DIFF
--- a/pkg/sign/transparency.go
+++ b/pkg/sign/transparency.go
@@ -23,6 +23,7 @@ import (
 	"encoding/pem"
 	"errors"
 	"fmt"
+	"path"
 	"time"
 
 	ssldsse "github.com/secure-systems-lab/go-securesystemslib/dsse"
@@ -58,6 +59,7 @@ const (
 
 type RekorClient interface {
 	CreateLogEntry(params *entries.CreateLogEntryParams, opts ...entries.ClientOption) (*entries.CreateLogEntryCreated, error)
+	GetLogEntryByUUID(params *entries.GetLogEntryByUUIDParams, opts ...entries.ClientOption) (*entries.GetLogEntryByUUIDOK, error)
 }
 
 type RekorV2Client interface {
@@ -217,6 +219,19 @@ func (r *Rekor) getRekorV2TLE(ctx context.Context, keyOrCertPEM []byte, b *proto
 	}
 	tle, err := r.options.ClientV2.Add(ctx, req)
 	if err != nil {
+		// Note: A 409 Conflict from Rekor v2 is currently not handled. The simplest
+		// solution is to sign again to generate a different signature.
+		// If someone wants to gracefully handle a 409 Conflict (e.g. from CI retries),
+		// they would need to:
+		// 1. Extract the log index from the error string (e.g. "an equivalent entry already exists... with index <idx>")
+		//    or from the x-log-index response header.
+		// 2. Fetch the checkpoint from the log's base URL and parse its size and root hash.
+		// 3. Use the tessera HTTP fetcher and ProofBuilder to fetch the inclusion proof for that index.
+		// 4. Fetch the entry bundle for the index, decode it, and extract the matching entry.
+		// 5. Construct the TransparencyLogEntry manually.
+		// However, doing so requires the log's public key (to compute the LogId field), which is
+		// only available from the trusted root, making it impossible to correctly populate the bundle here
+		// without access to the Verifier or TrustedRoot. Thus, we return the error.
 		return nil, fmt.Errorf("adding rekor v2 entry: %w", err)
 	}
 	return tle, nil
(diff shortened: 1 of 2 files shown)


pkg/sign/transparency.go
```diff
@@ -217,11 +229,108 @@ func (r *Rekor) getRekorV2TLE(ctx context.Context, keyOrCertPEM []byte, b *proto
...
+		return nil, fmt.Errorf("entry index %d out of bounds in bundle of size %d", entryIndex, len(bundle.Entries))
+	}
+
+	e := &rekortilespb.Entry{}
+	if err := protojson.Unmarshal(bundle.Entries[entryIndex], e); err != nil {
+		return nil, fmt.Errorf("unmarshaling entry: %w", err)
+	}
+
+	tle := &protorekor.TransparencyLogEntry{
+		// TODO: We need to populate the log ID, but we can't because
```
Hayden-IO: @steiza @cmurphy @kommendorkapten Wanted to get some input on options here.

With Rekor v1, when we fetch an entry from one of the log's read APIs, the log returns a complete TLE structure. For Rekor v2, the client must reconstruct the TLE, since there is no endpoint. This is _almost_ doable:

* The index is returned in both a custom header and the error message when responding with a 409
* The inclusion proof is constructed by fetching the relevant tiles computed with the index
* The canonicalized body and kind/version come from the entry bundle computed with the index
* The checkpoint comes from the log

Unfortunately, the log ID cannot be recomputed with anything from a Rekor v2 log. We've defined the log ID as the encoded public key hash, which comes from the trusted root (or in Rekor v1's case, it's either returned by the log or can be fetched from /api/v1/log). We aren't guaranteed to have the trusted root here.

The way we represent a log according to C2SP is a combination of the log's origin (e.g. `log2025-1.rekor.sigstore.dev`) and the first 4 bytes of its key ID as per c2sp.org/signed-note. Those values are included in the checkpoint.

So this will all be resolved going forward if we standardize on c2sp.org/tlog-proof and get rid of these redundant fields in the TLE. But what do we now?

Options:

* Rekor v2 uploads simply don't handle 409s
* Change the key lookup logic in pkg/verify/tlog to use the origin/key ID if there is no log ID present in the bundle. We also have to relax some of the constraints when parsing the bundle that mandate a log ID. I'd add a conformance test to verify if other clients can handle no log ID as well - some might, IIRC some just try to verify with all log keys.
* Inject the trusted root into the context higher up the call stack. If a trusted root is available, then fetch the entry and populate the log ID when the response is a 409.

Thoughts? Other options?

kommendorkapten: Ignoring 409 will break certain flows, as the same entry can't be uploaded again, but the expected failure rate is low. How bad would it be if we resign then try again from a client side? For ECDSA this would work as the signatures are non-determinisitc, and getting a new signature would so create a new TLE, and so not hit by the deduplication. How bad of an idea is that? 

If we think that is a terrible idea (to leave stray entries in the log) I would vote on injecting the trusted root, as verifying the received proof is before constructing the bundle is a good thing anyways.

Hayden-IO: I don't think that's a significant ask. Unfortunately, I don't have an example of a pipeline where this is coming up, I assume it's from bespoke pipelines. Curious if anyone has hit this with artifact attentions or a trusted publishing workflow. Also, this hasn't come up for v2 yet since there is no real world usage.

My biggest concern is this ends up being a bigger effort because we produce bundles that aren't compatible across clients. I think it's most important to resolve in cosign, 

I can fix this just for V1, since we know these errors are a server side flake occasionally. I can just leave a comment about we could go about resolving this for v2 if it becomes an issue.

kommendorkapten: IIRC this has come up when the npm registry submits the publish attestation, due to the async nature of how they are processed/ack'ed it became an issue. But this is only due to internal failure, the processing pipeline could also be updated to "retry from the start" on failure. 

As Rekor V2 will be a breaking change, I'm leaning into injecting the trust root for client side reconstruction, or a resign (with a small risk of stale entries).

Hayden-IO: Just coming back to this - given the amount of code needed to rebuild the TLE, I'd lean towards just getting this change in as-is for Rekor v1 only if that's alright, since that flake can happen due to no fault of the signer.

kommendorkapten (approved):

PR #658, opened 2026-07-16
fix(verify): prevent nil-pointer panic in TLogEntries

cmenjivar:
#### Summary

This PR hardens bundle verification against malformed or empty transparency-log data by adding defensive nil checks to the bundle and verification paths. When a bundle or its verification material is nil or incomplete, verification now returns explicit validation errors instead of panicking with a nil-pointer dereference. This prevents downstream crashes in consumers such as TFLint when processing certain GitHub Artifact Attestations.

Closes TFLint (issue #2591).

#### Release Note

Fixed a panic during bundle verification for malformed or empty transparency log data in some GitHub Artifact Attestations.

#### Documentation

None. No user-facing documentation changes are required for this bug fix.

#### Verified locally with:

go test ./pkg/bundle ./pkg/verify

### DIFF
--- a/pkg/bundle/bundle.go
+++ b/pkg/bundle/bundle.go
@@ -330,12 +330,12 @@ func (b *Bundle) HasInclusionProof() bool {
 }
 
 func (b *Bundle) TlogEntries() ([]*tlog.Entry, error) {
-	if b == nil {
-		return nil, errors.New("Cannot retrieve tlog entries from a nil bundle")
+	if b.VerificationMaterial == nil {
+		return nil, ErrMissingVerificationMaterial
 	}
 
-	if b.VerificationMaterial == nil {
-		return nil, errors.New("bundle verification material is uninitialized or nil")
+	if len(b.VerificationMaterial.TlogEntries) == 0 {
+		return nil, nil
 	}
 
 	if n := len(b.VerificationMaterial.TlogEntries); n > limits.MaxAllowedTlogEntries {
--- a/pkg/bundle/bundle.go
+++ b/pkg/bundle/bundle.go
@@ -97,6 +97,11 @@ func (b *Bundle) validate() error {
 		return fmt.Errorf("%w: bundle version %s is not supported", ErrUnsupportedMediaType, bundleVersion)
 	}
 
+	err = validateBundle(b.Bundle)
+	if err != nil {
+		return fmt.Errorf("invalid bundle: %w", err)
+	}
+
 	// fetch tlog entries, as next check needs to check them for inclusion proof/promise
 	entries, err := b.TlogEntries()
 	if err != nil {
@@ -133,10 +138,6 @@ func (b *Bundle) validate() error {
 		return fmt.Errorf("%w: bundle version %s is not yet supported", ErrUnsupportedMediaType, bundleVersion)
 	}
 
-	err = validateBundle(b.Bundle)
-	if err != nil {
-		return fmt.Errorf("invalid bundle: %w", err)
-	}
 	return nil
 }
 
@@ -329,8 +330,12 @@ func (b *Bundle) HasInclusionProof() bool {
 }
 
 func (b *Bundle) TlogEntries() ([]*tlog.Entry, error) {
+	if b == nil {
+		return nil, errors.New("Cannot retrieve tlog entries from a nil bundle")
+	}
+
 	if b.VerificationMaterial == nil {
-		return nil, nil
+		return nil, errors.New("bundle verification material is uninitialized or nil")
 	}
 
 	if n := len(b.VerificationMaterial.TlogEntries); n > limits.MaxAllowedTlogEntries {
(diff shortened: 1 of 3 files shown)


pkg/bundle/bundle.go
```diff
@@ -329,8 +330,12 @@ func (b *Bundle) HasInclusionProof() bool {
 }
 
 func (b *Bundle) TlogEntries() ([]*tlog.Entry, error) {
+	if b == nil {
+		return nil, errors.New("Cannot retrieve tlog entries from a nil bundle")
+	}
+
 	if b.VerificationMaterial == nil {
-		return nil, nil
+		return nil, errors.New("bundle verification material is uninitialized or nil")
```
Hayden-IO: This is an acceptable state though, shouldn't a client calling this handle an empty array?

pkg/bundle/bundle.go
```diff
@@ -133,10 +138,6 @@ func (b *Bundle) validate() error {
 		return fmt.Errorf("%w: bundle version %s is not yet supported", ErrUnsupportedMediaType, bundleVersion)
 	}
 
-	err = validateBundle(b.Bundle)
```
Hayden-IO, line 136: Why is this being moved?

pkg/bundle/bundle.go
```diff
@@ -329,8 +330,12 @@ func (b *Bundle) HasInclusionProof() bool {
 }
 
 func (b *Bundle) TlogEntries() ([]*tlog.Entry, error) {
+	if b == nil {
```
Hayden-IO: This is an unnecessary check, if b were nil, b.TlogEntries would have panicked already.

Hayden-IO (changes requested):
You mentioned an issue number, but I'm not sure what issue this is referring to?

cmenjivar:
> You mentioned an issue number, but I'm not sure what issue this is referring to?

https://github.com/terraform-linters/tflint/issues/2591

Hayden-IO:
Thanks, taking a look

pkg/bundle/bundle.go
cmenjivar: Because the earlier placement was effectively doing the same validation too early, before the method had finished the rest of its bundle checks. Moving it up makes the validation happen immediately after the bundle object is inspected, so malformed bundles are rejected before the method continues into the TLog and inclusion-proof logic.

In other words, the change makes the validation flow more direct:
- check bundle structure first
- then proceed with the rest of the validation logic

That also keeps the behavior consistent with the existing validateBundle path and avoids letting later steps run on an invalid bundle.

pkg/bundle/bundle.go
cmenjivar: I changed this to preserve the original empty-result behavior for an empty TLog entry list: `TlogEntries()` now returns nil, nil for that case. I also kept the defensive nil check for a nil bundle and switched the missing-verification-material path to use the existing package validation error, so the behavior stays consistent with the rest of the bundle validation flow.

pkg/bundle/bundle.go
cmenjivar: Removed the redundant nil receiver check; the method now keeps the existing behavior for missing verification material without an extra guard.

Hayden-IO:
@steiza Could you take a look at the linked issue? I think the fix here to move the validation logic up is fine, but I'm curious if there's something else going on, either with us needing to recommend different verification options to integrators or if there was some issue with GHA provenance generation that led to missing log entries.

pkg/bundle/bundle.go
Hayden-IO: Thanks, this SGTM.

pkg/verify/tlog.go
```diff
@@ -48,7 +48,10 @@ import (
 func VerifyTlogEntry(entity SignedEntity, trustedMaterial root.TrustedMaterial, logThreshold int, trustIntegratedTime bool) ([]root.Timestamp, error) { //nolint:revive
 	entries, err := entity.TlogEntries()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to retrieve transparency log entries: %w", err)
+	}
+	if len(entries) == 0 {
```
Hayden-IO: `VerifyTlogEntry` will only be called if the client has configured the verifier to expect a logged signature. An empty entries list shouldn't cause a panic, it should skip the loops and fail to meet the threshold as expected. I don't think we need to have this check.

cmenjivar: removed `if len(entries) == 0` check from `VerifyTlogEntry`

steiza:
> @steiza Could you take a look at the linked issue? I think the fix here to move the validation logic up is fine, but I'm curious if there's something else going on, either with us needing to recommend different verification options to integrators or if there was some issue with GHA provenance generation that led to missing log entries.

Yeah, so what happened here is that GitHub made some [breaking changes to the API with version `2026-03-10`](https://docs.github.com/en/rest/about-the-rest-api/breaking-changes?apiVersion=2026-03-10) which included changes to the artifact attestation APIs. Those breaking changes were mistakenly backported to API version `2022-11-28`, and that mistaken backport has now been reverted.

Hayden-IO (approved):

cmenjivar:
Thanks for the review and approval! It looks like I don't have write permissions to merge into this branch due to branch protection rules. Could you (or another maintainer) please merge this when you have a moment?

PR #666, opened 2026-07-30
verify: stop materializing in-toto predicates the verifier never reads

mattmoor:
#### Summary

While profiling a production service that verifies DSSE bundles whose in-toto predicates run to megabytes (vulnerability reports; SBOM and build-provenance predicates have the same shape), we found that statement parsing dominated the verifier's memory profile, for two compounding reasons:

  1. Envelope.Statement() parses the entire payload with protojson.Unmarshal, which materializes the predicate as a structpb tree: one heap object per JSON node. For large predicates this allocates roughly 10x the payload size and hundreds of thousands of objects per parse, all garbage moments later.

  2. The verification path parses the statement up to three times per bundle (once per artifact-policy check, once to populate the result), multiplying that cost.

Verification itself only ever consumes the statement type, the subjects, and the predicate type. Introduce summarizeStatement, which decodes exactly those fields with encoding/json and leaves Predicate nil, and use it in the artifact-policy verification paths. Building the VerificationResult still parses the full statement by default, so no API behavior changes; a new WithoutStatementPredicate option lets callers that read the predicate from the verified payload directly (or not at all) opt out of materializing it there too.

The included benchmark, on a statement with a ~1.4MiB predicate:

    BenchmarkStatementParse/full       16.0ms/op  14.0MB/op  410152 allocs/op
    BenchmarkStatementParse/summary     4.2ms/op   1.4MB/op      38 allocs/op

i.e. a ~10x reduction in bytes and a four-orders-of-magnitude reduction in object count per parse. In the profiled service, verifying a batch of ~740 such bundles, statement parsing accounted for roughly half of all allocations on the read path; with the verification-path summary plus the opt-out, peak heap during the batch halved.

#### Release Note

Dramatically reduce the amount of wasted memory allocation in the verification path.

#### Documentation

N/A

### DIFF
--- a/pkg/verify/signature.go
+++ b/pkg/verify/signature.go
@@ -254,7 +254,9 @@ func verifyEnvelopeWithArtifacts(verifier signature.Verifier, envelope EnvelopeC
 	if err := verifyEnvelope(verifier, envelope); err != nil {
 		return err
 	}
-	statement, err := envelope.Statement()
+	// Verification only consumes the statement's subjects; summarizing
+	// avoids materializing the (potentially very large) predicate.
+	statement, err := summarizeStatement(envelope)
 	if err != nil {
 		return fmt.Errorf("could not verify artifact: unable to extract statement from envelope: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


mattmoor:
Not sure what the best-practice is here anymore, so tagging some friends 😄 

cc @loosebazooka @Hayden-IO @codysoyland

mattmoor:
The e2e test failure looks like a GOPROXY flake? 🤔 
```
golang.org/x/net@v0.56.0: read "https://proxy.golang.org/golang.org/x/net/@v/v0.56.0.zip": unexpected EOF
```

kommendorkapten (approved):
Looks good, thanks!

PR #671, opened 2026-08-21
retry transient TUF downloads by default

crazy-max:
needs https://github.com/sigstore/sigstore/pull/2396

#### Summary

Configure the default TUF fetcher with a retryable HTTP client so temporary server failures are retried before trusted root loading fails. This preserves non-retry behavior for permanent 404 responses and keeps custom fetchers unchanged.

This builds on https://github.com/sigstore/sigstore/pull/2396, which added a shared `github.com/sigstore/sigstore/pkg/httpretry` package with a retrying `http.RoundTripper`, default retry policy, jittered backoff, `Retry-After` handling, and request replay safeguards. Using that shared package keeps this change small and avoids adding sigstore-go-specific retry logic or another third-party retry dependency.

This fixes a reliability gap in the default Sigstore TUF path. Downstream clients such as cosign rely on sigstore-go to load the live trusted root from TUF before verifying bundles in the new bundle format. Older cosign behavior could later surface the generic error `trusted root is required when using new bundle format` when trusted root loading had failed earlier, which made the failure look like missing verification configuration rather than a TUF download problem. That diagnostic was fixed in https://github.com/sigstore/cosign/pull/4878 by returning the wrapped TUF error for new bundle verification. This PR is complementary to that cosign fix because it reduces the chance of reaching that trusted-root loading failure in the first place.

We observed this class of failure in CI while verifying several OCI 1.1 Sigstore bundles with `cosign verify --experimental-oci11 --new-bundle-format`. Earlier images in the same job verified successfully, while a later image failed while loading trusted material. Returning the real TUF error is the right diagnostic behavior, but the default TUF fetch path should also be resilient enough to retry ordinary transient HTTP failures before reporting that trusted root loading failed.

The downstream workaround space is awkward. Consumers can materialize trust root data once and pass it explicitly with `--trusted-root`, or they can add retry wrappers around their verification steps. Those approaches are useful for workflows that intentionally want to manage a trusted root explicitly, but they are still compensating for a lower-level reliability issue. The default TUF fetch path should tolerate ordinary transient failures itself, instead of pushing every consumer toward local retry wrappers, checked-in trust material, or integration-specific special casing.

This behavior appears to have existed before and was lost during the TUF client rewrite. Commit https://github.com/sigstore/sigstore-go/commit/5a823fc2e6fbddbefacce83fe2e34cd0040736c9 moved the older client to `filejsonstore` and explicitly added HTTP retries by setting `HTTPRemoteOptions.Retries` to `tufclient.DefaultHTTPRetries`. https://github.com/sigstore/sigstore-go/pull/41 later implemented the broader TUF client spec work from https://github.com/sigstore/sigstore-go/issues/38, switched sigstore-go to go-tuf v2, and replaced the old `HTTPRemoteOptions` path with `metadata/updater` plus a pluggable `fetcher.Fetcher`. That migration introduced the current default `fetcher.NewDefaultFetcher()` setup, but no equivalent retrying HTTP client was carried forward. This PR restores that resilience in the new go-tuf v2 fetcher model.

The scope is intentionally narrow. The change only affects the default fetcher constructed by `tuf.DefaultOptions()` and by the fallback path in `tuf.New()` when callers do not provide a custom fetcher. It does not change TUF validation, trust policy, cache semantics, bundle verification policy, public APIs, or any caller-provided fetcher. It keeps the Sigstore user agent behavior and installs the shared Sigstore retrying HTTP client on the go-tuf default fetcher. Permanent failures such as HTTP 404 are still returned without retrying, which avoids hiding repository or target configuration mistakes.

This is covered by focused tests in `pkg/tuf`. One test uses an `httptest.Server` that returns HTTP 503 twice before succeeding, and verifies that the default fetcher retries and returns the downloaded body. A second test returns HTTP 404 and verifies that the default fetcher does not retry permanent not-found responses.

#### Release Note

The default Sigstore TUF fetcher retried transient HTTP download failures by using the shared Sigstore HTTP retry client. Custom fetchers continued to control their own HTTP behavior.

#### Documentation

NONE. This change does not add a public API or new user-facing configuration, and callers using the default TUF options receive the more reliable fetch behavior automatically.

cc @Hayden-IO


go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/go-openapi/strfmt v0.27.0
 	github.com/go-openapi/swag/conv v0.27.3
 	github.com/google/certificate-transparency-go v1.3.3
+	github.com/hashicorp/go-retryablehttp v0.7.8
```
Hayden-IO: On Rekor, someone asked for us to not use hashicorp repos as their license is incompatible with CNCF guidelines. I implemented my own no-dep retry client, though I bet there's an existing one as well. I could move the retry client to the Sigstore utilities repository?

crazy-max: That makes sense. I originally used `go-retryablehttp` because it was already present in the module graph as a transitive dep, so the intent was to reuse a mature retry implementation without introducing a wholly new dependency to the project.

I agree that making it a direct dependency is probably not the right tradeoff if the HashiCorp licensing concern is relevant for CNCF projects.

Also agree that a small shared Sigstore utility for retryable HTTP clients would be useful, so TUF/Rekor/other clients can share one consistent retry policy instead of each package carrying its own version.

crazy-max: Opened https://github.com/sigstore/sigstore/pull/2396

crazy-max:
made some temp changes to test https://github.com/sigstore/sigstore/pull/2396

go.mod
```diff
@@ -89,9 +87,11 @@ require (
...
-	google.golang.org/genproto/googleapis/api v0.0.0-20260526163538-3dc84a4a5aaa // indirect
-	google.golang.org/genproto/googleapis/rpc v0.0.0-20260523011958-0a33c5d7ca68 // indirect
+	google.golang.org/genproto/googleapis/api v0.0.0-20260727163830-6c54dddc4772 // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20260720155508-bb71a54f79dc // indirect
 	google.golang.org/grpc v1.82.1 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	k8s.io/klog/v2 v2.140.0 // indirect
 )
+
+replace github.com/sigstore/sigstore => github.com/crazy-max/sigstore v0.0.0-20260825142716-1972768a7c46
```
steiza: Now that https://github.com/sigstore/sigstore/pull/2396 landed, I think we can back this out?

steiza (commented):
Looks reasonable to me! I think after you back out your testing change, this is ready for final review.

codysoyland (dismissed):

crazy-max:
I updated this to use the shared `github.com/sigstore/sigstore/pkg/httpretry` package from https://github.com/sigstore/sigstore/pull/2396 instead of carrying sigstore-go-specific retry logic. The change is now limited to installing the shared retrying HTTP client on the default TUF fetcher. Custom fetchers remain unchanged, and permanent 404 responses still do not retry.

kommendorkapten (approved):

kommendorkapten (changes requested):
Can you re-run go mod tidy please, there seems to be some issues with it.

crazy-max:
> Can you re-run go mod tidy please, there seems to be some issues with it.

Oh indeed, need to update `./examples/oci-image-verification`

crazy-max:
@kommendorkapten should be good now

kommendorkapten (approved):
Thank you!

PR #676, opened 2026-09-04
sign: retry Fulcio and TSA requests with httpretry

codysoyland:
The Fulcio and timestamp authority clients each carried a hand-rolled retry loop: retry only on 5xx/429 with a fixed 2^n-second backoff (no cap, no jitter, no Retry-After), no retry on connection errors, a request built without the caller's context, and on exhaustion the response body was closed inside the loop and then read again, so the returned error lost the server's message. #671 brought in sigstore/sigstore's `httpretry` package for the TUF fetcher; this PR moves Fulcio and TSA onto the same package so sigstore-go has one retry policy rather than three.

Behaviour changes for reviewers: connection errors are now retried; backoff is jittered exponential (1s floor, 30s cap) and honours Retry-After; 501 is no longer retried; `Timeout` now bounds the whole call including backoff rather than each attempt, matching how Rekor v1's go-openapi timeout already behaves; `Retries: 0` still means a single attempt. Rekor v1 is unchanged (rekor's client retries internally). Rekor v2 will follow in a separate PR once rekor-tiles exposes a transport option.

Key changes:
- **`newRetryTransport` helper** (`pkg/sign/retry.go`): wraps an optional base transport in `httpretry.Transport` with `MaxRetries` taken from the `Retries` option (clamped for gosec G115); shared by Fulcio and TSA, and by Rekor v2 later.
- **Retry loops removed**: `GetCertificate` and `GetTimestamp` issue one `http.NewRequestWithContext` + `client.Do`; the `bytes.Buffer`/`bytes.Reader` bodies set `GetBody`, so the POSTs are replayable and headers such as `Authorization` are re-sent via `Request.Clone`.
- **Option docs**: `Timeout`, `Retries`, and `Transport` comments describe the new semantics; an injected `Transport` is now wrapped, so callers whose transport already retries should set `Retries: 0`.
- **Tests**: existing mock transports keep working; backoff is zeroed per instance via `disableBackoff` so the retry tests no longer sleep (~6s saved); new cases cover transport-error retry, `Retries: 0` making a single attempt, and the exhaustion error including the response body.

Assisted by Claude Fable 5.1

### DIFF
--- a/pkg/sign/certificate.go
+++ b/pkg/sign/certificate.go
@@ -23,7 +23,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"math"
 	"net/http"
 	"strings"
 	"time"
@@ -54,11 +53,11 @@ type Fulcio struct {
 type FulcioOptions struct {
 	// URL of Fulcio instance
 	BaseURL string
-	// Optional timeout for network requests (default 30s; use negative value for no timeout)
+	// Optional timeout for the whole request including retries (default 30s; use negative value for no timeout)
 	Timeout time.Duration
-	// Optional number of times to retry on HTTP 5XX
+	// Optional number of times to retry on transient failures (connection errors, HTTP 429, HTTP 5xx other than 501); zero disables retries
 	Retries uint
-	// Optional Transport (for dependency injection)
+	// Optional base Transport (for dependency injection); it is wrapped by the retrying transport, so set Retries to 0 if it already retries
 	Transport http.RoundTripper
 }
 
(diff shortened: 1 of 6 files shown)


kommendorkapten (approved):
Looks good!

steiza (changes requested):
We somewhat recently added a retry client to github.com/sigstore/sigstore (https://github.com/sigstore/sigstore/pull/2396) - if we're changing functionality here, maybe we can use that instead?

codysoyland:
> We somewhat recently added a retry client to github.com/sigstore/sigstore ([sigstore/sigstore#2396](https://github.com/sigstore/sigstore/pull/2396)) - if we're changing functionality here, maybe we can use that instead?

That's actually the client this PR uses (`newRetryTransport` just wraps it to inject the transport and max retries so the existing options still work).

steiza (approved):
Oops! Somehow I missed the implementation in `pkg/sign/retry.go` - thanks!

PR #679, opened 2026-09-18
Read the proto spellings of _type and predicateType in the statement summary

arpitjain099:
`summarizeStatement` reads `_type` and `predicateType` off the DSSE payload with `encoding/json`. protojson, which `Envelope.Statement` uses, accepts the proto field names too, so a statement spelled `"type"` and `"predicate_type"` parses either way but comes out of the summary with both fields empty.

That matters because the summary is not a side path: `WithoutStatementPredicate` routes the verification result through it, and `VerifySignatureWithArtifactDigests` always does. Subjects are unaffected (same spelling in both), so digest matching is fine; it is the statement type and predicate type that go missing.

The fix reads both spellings and rejects a statement that sets both, which is what protojson does with a repeated field. `BenchmarkStatementParse/summary` is unchanged at 26 allocs/op, so the reason the summary exists is intact.

One difference is left: protojson rejects duplicate members and `encoding/json` takes the last. Catching that needs a token scan of the whole payload, which costs more than the summary saves.


kommendorkapten:
👋  When do you run in to the problem you describe? Is that when a malformed in-toto statement is being processed? The [in-toto statement schema](https://github.com/in-toto/attestation/blob/main/spec/v1/statement.md) marks `_type` and `predicateType` as required.

